PR #67, opened 2019-07-10
Exploring moving back to Beam

expede:
Some indications that they may have solved the 15 minute build issue. I have a slight preference for Beam over Selda, but it's really not a big deal. Mostly the kind of thing that I'll play with in my spare time.


expede:
Way way way too much work to switch at this stage. I actually do prefer the Beam style, and would probably pick it if starting a new project. Kind of sad that it was so slow to compile right when we started this project 🤷‍♀️ Not enough of an advantage to blow many hours on.

Issue #73, opened 2019-07-18
Improve curl example


bmann:
# Problem

I got to constructing what I think is a valid curl request, but got a 401.

# Solution

Had to kind of puzzle my way through this, using the user/pass that I have, especially the Basic Auth Header. Looks like curl -u user:pass will do this for us.

Will update and PR to README if I get something working.

bmann:
Yay! OK, also need to wrap u/p in single quotes, but that worked, will update the readme.

PR #74, opened 2019-07-18
Updated curl examples

bmann:
Should just work with heroku credentials. curl -u is easier than separately encoding user/pass, and --data-binary is required for binary files.

**Summary**

* [ ] Improve curl examples #73


README.md
```diff
@@ -163,12 +162,14 @@ curl -i \
...
+Using the `--data-binary` and a /path/to/file can upload binary file types.
+
 ```shell
 curl -i \
     -X POST \
-    -H "Authorization: Basic abcdef==" \
+    -u 'USER:PASS' \
     -H "Content-Type: application/octet-stream" # This line
-    -d '{"hi":1}' \
+    --data-binary @/path/to/file \
```
expede: May also want to show a plaintext example, not just files. Thoughts?

bmann: Yes, we should show plaintext as well as files — eg. a “Hello FISSION” string, a Markdown file, and JSON.

As I recall we can do interesting things with JSON + IPFS?

Maybe I’ll also do a HEROKU.md file that very specifically calls out how to work with what is essentially your own endpoint. Gets people used to Heroku even if they don’t access stuff from there. We only care about having a way to get people credentials.

PR #75, opened 2019-07-19
Split public and private docs

expede:


expede:
Not deployed yet, but cc @bmann

Issue #81, opened 2019-08-07
`DELETE` does not remove CID from `user_cids` table


dholms:
# Summary

## Problem
DELETE request to server does not remove the CID from the associated user in the db

## Solution
Remove CID from user table when deleting a resource

# Detail
**To Reproduce**
1. Post a resource for a given user
2. `GET http://localhost:1337/ipfs/cids` should include CID for the resource
3. `DELETE http://localhost:1337/ipfs/{CID}` returns `202`
4. `GET http://localhost:1337/ipfs/cids` still includes CID for the resource

**Expected behavior**
After `DELETE`, `GET http://localhost:1337/ipfs/cids` should not include the CID

PR #82, opened 2019-08-15
Fix double quotes in DB issue

expede:
# Problem

1. [X] #81 `DELETE` does not remove CID
2. [X] Inconsistent double quotes around CID shown to user

# Analysis

## `DELETE` does not remove CID

* An extra layer of double quotes were present in the output from IPFS's `stdout`
  * e.g. Via log: `["\"QmQWoEjbKZhj1qpPSKUv7LRepRWTvZfvCoKBJFU9CQGrWa\""]`
    * End user sees `"QmQWoEjbKZhj1qpPSKUv7LRepRWTvZfvCoKBJFU9CQGrWa"`
* This was also putting extra double quotes _in the DB_
* When we looked up the CID, we would request it as a regular string
* No matches, because the DB is full of CIDs wrapped in an extra layer of quotes

## Inconsistent double quotes around CID shown to user

The above double-quote issue was noticed earlier in development, and assumed to be a JSON encoding issue, and handled in the `ToJSON` typeclass instance.

# Solution

1 & 2. Strip the extra quotes when reading from IPFS's `stdout`

### DIFF
--- a/src/Fission/IPFS/SparseTree.hs
+++ b/src/Fission/IPFS/SparseTree.hs
@@ -16,11 +16,8 @@ import Fission.IPFS.Path.Types
 import Fission.IPFS.SparseTree.Types
 
 linearize :: SparseTree -> Either Error.Linearization Path
-linearize = fmap (Path . wrap "\"") . go
+linearize = fmap Path . go
   where
-  wrap :: Text -> Text -> Text
-  wrap outside inside = outside <> inside <> outside
-
   go :: SparseTree -> Either Error.Linearization Text
   go = \case
     Stub      (Name name)    -> Right $ UTF8.textShow name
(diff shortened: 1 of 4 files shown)


src/Fission/IPFS/SparseTree.hs
```diff
@@ -16,11 +16,8 @@ import Fission.IPFS.Path.Types
 import Fission.IPFS.SparseTree.Types
 
 linearize :: SparseTree -> Either Error.Linearization Path
-linearize = fmap (Path . wrap "\"") . go
+linearize = fmap Path . go
   where
-  wrap :: Text -> Text -> Text
-  wrap outside inside = outside <> inside <> outside
-
```
expede: This always felt weird

PR #83, opened 2019-08-15
Remove `/ipfs` from provision host

expede:
# Problem

Heroku provision includes the `/ipfs` path. This makes it difficult to access other parts of the API.

# Solution

No longer append `/ipfs` in provision

### DIFF
--- a/src/Fission/Web/Heroku.hs
+++ b/src/Fission/Web/Heroku.hs
@@ -48,7 +48,7 @@ create Request {_uuid, _region} =
 
     let
       userConfig = Heroku.UserConfig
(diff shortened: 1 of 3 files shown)


expede:
cc @bmann @dholms

src/Fission/Web/Heroku.hs
```diff
@@ -48,7 +48,7 @@ create Request {_uuid, _region} =
 
     let
       userConfig = Heroku.UserConfig
-        { Heroku._interplanetaryFissionUrl      = url <> "/ipfs"
```
expede: That's the line

PR #90, opened 2019-08-28
Pin to cluster

expede:
* [x] Cleaner interface
* [x] Can talk to cluster

Actually kind of interesting solution. Initially started switching everything to the web interface. Later realized that the local node is bootstrapped to the cluster, so we can write files to disk, and pin after the fact. This will be faster (since the Web API is not currently streaming it often has to write temp files anyways), and we leverage IPFS. Also if we upload a file that has a copy elsewhere, the cluster can grab it from multiple sources, rather than have it only come from the Web API.


src/Fission/Storage/IPFS.hs
```diff
@@ -45,90 +50,105 @@ get cid@(IPFS.CID hash) = IPFS.Proc.run ["cat"] (UTF8.textToLazyBS hash) >>= \ca
...
       -> m (Either IPFS.Error.Add CID)
-unpin cid@(CID hash) = IPFS.Proc.runErr' ["pin", "rm"] (UTF8.textToLazyBS hash) >>= \case
-  (ExitSuccess, _) -> do
-    logDebug $ "Unpinned CID " <> display hash
-    return $ Right cid
+unpin cid@(CID hash) = IPFS.Client.run (IPFS.Client.unpin hash False) >>= \case
+  Right Pin.Response { cids } ->
+    case cids of
+      [cid'] -> do
+        logDebug $ "Pinned CID " <> display hash
```
expede: @expede FIX ME

PR #105, opened 2019-09-16
fixing FromEnv instances & prod default for host

dholms:
# Problem
Envy not pulling values from the environment

# Solution
Re-add `FromEnv` instances for env variable types

Fixes https://github.com/fission-suite/web-api/issues/103


fission-web/Main.hs
```diff
@@ -36,7 +36,7 @@ main = do
   Just manifest <- decodeFileStrict "./addon-manifest.json"
 
   _processCtx  <- mkDefaultProcessContext
-  _host        <- decode .!~ Web.Host "localhost:1337"
+  _host        <- decode .!~ Web.Host "https://runfission.com"
   _ipfsPath    <- decode .!~ IPFS.BinPath "/usr/local/bin/ipfs"
   _ipfsTimeout <- decode .!~ IPFS.Timeout 150
   _dbPath      <- decode .!~ DB.Path "web-api.sqlite"
```
expede: It's outside of the diff, so can't comment on the line, but line 54 looks broken to me 🤔 We should at least add an explicit default while we have the hood up.

```hs
isTLS <- getFlag "TLS" .!~ True
```

library/Fission/IPFS/Types.hs
```diff
@@ -48,4 +49,6 @@ newtype Timeout = Timeout { getSeconds :: Natural }
                     , Show
                     , Generic
                     )
-  deriving anyclass FromEnv
+
+instance FromEnv Timeout where
+  fromEnv _ = Timeout <$> env "IPFS_TIMEOUT"
```
expede: I'm not getting this value showing up locally via the REPL 🤔 Will dig a bit

expede: Good news! It was a `fish` issue, not a PR issue 👍

fission-web/Main.hs
dholms: Broken how? The flag works fine as is, but does default to False. I'll fix that

expede: > Broken how? 
This shouldn't typecheck: `if (isTLS :: Maybe Bool) then ... else ...`. Adding the `.!~` fixes it

library/Fission/Environment.hs
```diff
@@ -55,5 +55,5 @@ withEnv key fallback transform = pure (maybe fallback transform) <*> lookupEnv k
 --
 -- >>> getFlag ""
 -- False
```
expede: Nonzero chance that this doctest will now fail

expede:
```shell
 ~/D/W/F/web-api  ╍╍ envy  stack test fission-web-api:fission-doctest

# ...

/Users/expede/Documents/Work/FISSION/web-api/.doctest-tmp/library/Fission/Environment.hs:57: failure in expression `getFlag ""'
expected: False
 but got: Nothing

Examples: 19  Tried: 19  Errors: 0  Failures: 1
```

dholms: yuup

PR #106, opened 2019-09-16
add note about needing to install liblzma

dholms:
# Problem
Quick start instructions don't work on linux because `liblzma-dev` must be installed separately

# Solution
Add a note in Readme QuickStart instructions to `sudo apt install liblzma-dev` on linux

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ brew service start ipfs
 # Enable sample Heroku config
 mv addon-manifest.json.example addon-manifest.json
 
-# If using linux, install liblzma-dev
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,9 @@ brew service start ipfs
 # Enable sample Heroku config
 mv addon-manifest.json.example addon-manifest.json
 
+# If using linux, install liblzma-dev
```
expede: Nitpicky typo:Linux has a capital "L"

dholms: Oops good looks

PR #107, opened 2019-09-18
added POST /ipfs/dag route for dag puts

dholms:
## Summary
Added POST /ipfs/dag rotue for dag puts

This allows for more complex IPLD structures to be added via the web-api. This includes directory structures to allow for uploading websites.

### Possible issues
- I reused Add Errors for these routes since the error handling is very similar and I saw they were used in pin/unpin routes as well.
- The route currently accepts plaintext and octet-streams while `ipfs dag put` expects input to be JSON formatted. This is mostly just a matter of where the error is handled. Should the route only accept JSON? I did it this way because:
  - it more closely mimics the ipfs api
  - I was having a hard time getting Servant to accept unstructured JSON

Resolves https://github.com/fission-suite/web-api/issues/102


expede:
> accept unstructured JSON

Yeah, we generally don't want unstructured JSON. We can 100% write a data type that expresses what we expect to/from the IPFS DAG. Might be a good exercise to go through. It's totally possible to do unstructured JSON, but is considered an anti-pattern.

library/Fission/Storage/IPFS.hs
```diff
@@ -4,6 +4,7 @@ module Fission.Storage.IPFS
   , get
   , pin
   , unpin
+  , dagput
```
expede: This is fine, but it can't hurt to move this to `Fission.IPFS.DAG.put`. We'll likely accrue more DAG functions as time goes on, and we lose nothing by being super clear with naming.

library/Fission/Storage/IPFS.hs
```diff
@@ -154,3 +155,19 @@ logLeft errStr = do
...
+        => Has IPFS.BinPath  cfg
+        => Has IPFS.Timeout  cfg
+        => Lazy.ByteString
+        -> m (Either IPFS.Error.Add IPFS.CID)
+dagput raw = IPFS.Proc.run ["dag", "put", "-f", "dag-pb"] raw >>= \case
+  (ExitSuccess, result, _) ->
+    case CL.lines result of
+      [cid] ->  pin . mkCID . UTF8.stripN 1 $ UTF8.textShow cid
+      bad   -> return . Left . UnexpectedOutput $ UTF8.textShow bad
+  (ExitFailure _, _, err) -> return . Left . UnknownAddErr $ UTF8.textShow err
```
expede: 💅 Minor style points

1. I generally try to drop to the next line if any of the branches do
2. `return` is fine, but if we're not in `do` notation it can be a good signal to use `pure`. They absolutely the same function, but `return` often implies that you should be looking for a `do`.

library/Fission/Web/IPFS/Dag.hs
```diff
@@ -0,0 +1,43 @@
+module Fission.Web.IPFS.Dag
```
expede: "DAG" is an initialism, and should be upcased

library/Fission/Web/IPFS.hs
```diff
@@ -38,10 +39,12 @@ type AuthedAPI = Auth :> UnauthedAPI
 type UnauthedAPI = "cids" :> CID.API
               :<|> Upload.API
               :<|> Pin.API
+              :<|> ("dag" :> Dag.API)
 
 type SimpleAPI = "cids" :> CID.API
             :<|> Upload.Simple.API
             :<|> Pin.API
+            :<|> ("dag" :> Dag.API)
```
expede: I really should rearrange the order in these types to make them easier to maintain 🤔 In either case, I'm pretty sure that we don't need the parentheses [UPDATE: removed locally and they're not needed)

expede: While we're in here:

```haskell
type UnAuthedAPI = CommonAPI :<|> Upload.API
type SimpleAPI   = CommonAPI :<|> Upload.Simple.API

type CommonAPI = "cids" :> CID.API
            :<|> "dag"  :> Dag.API
            :<|> Pin.API
```

This will break order in a few places, but is cleaner. Thoughts?

library/Fission/Web/IPFS/Dag.hs
```diff
@@ -0,0 +1,43 @@
...
+    => HasLogFunc        cfg
+    => User
+    -> RIOServer         cfg API
+put User { _userID } (Serialized rawData) = Storage.IPFS.dagput rawData >>= \case
+  Right newCID -> do
+    void $ User.CID.createX _userID [newCID]
+    return newCID
+
+  Left err ->
+    Web.Err.throw err
```
expede: 👍

expede:
🔢 We should also bump the version number in `package.yaml`

library/Fission/Storage/IPFS/DAG/Put.hs
```diff
@@ -0,0 +1,35 @@
...
+        => Has IPFS.BinPath  cfg
+        => Has IPFS.Timeout  cfg
+        => Lazy.ByteString
+        -> m (Either IPFS.Error.Add IPFS.CID)
+put raw = IPFS.Proc.run ["dag", "put", "-f", "dag-pb"] raw >>= \case
+  (ExitSuccess, result, _) ->
+    case CL.lines result of
+      [cid] ->  pin . mkCID . UTF8.stripN 1 $ UTF8.textShow cid
+      bad   -> pure . Left . UnexpectedOutput $ UTF8.textShow bad
+  (ExitFailure _, _, err) -> pure . Left . UnknownAddErr $ UTF8.textShow err
```
expede: ❓Hmm, may have overshot the mark a bit, I should have been clearer 😅 Just stick `put` in `Fission.IPFS.DAG`, unless we expect to have a bunch of `put` related functionality?

library/Fission/Web/IPFS.hs
expede: Got labelled as outdated, but isn't :octocat:

Issue #70, opened 2019-07-13
Attempt moving Config to SuperRecord


expede:
# Problem

Classic duplicate field stuff. More flexibility would be nice / `Data.Has` feels quite manual.

# Solution

Attempt moving to `SuperRecord`

## Nota Bene

Mostly a note-to-self for weekend hackery

expede:
https://docs.google.com/spreadsheets/d/14MJEjiMVulTVzSU4Bg4cCYZVfkbgANCRlrOiRneNRv8/edit#gid=0

expede:
Ran through the application and swapped out `Data.Has` for `SuperRecord` (#114). Ran some perf tests locally with `h2load`. The numbers were... not what I was expecting! (in a good way 😅)

<img width="1330" alt="Screen Shot 2019-09-22 at 15 23 48" src="https://user-images.githubusercontent.com/1052016/65395256-4e74ab80-dd4d-11e9-9772-f82cf623b3c9.png">

PR #114, opened 2019-09-22
SuperRecord experiment

expede:
#70


expede:
The type errors get pretty hairy, and I'm pretty sure that we can get the relative benefits with `Data.Has` and `DuplicateRecordFields`. Good experiment; but closing #70

PR #119, opened 2019-10-01
default ipfs/cids to return JSON instead of PlainText

dholms:
# Problem
/ipfs/cids route returns a plaintext string. Which forces the frontend to parse it into an array of CIDs
# Solution
default /ipfs/cids to return JSON, in this case an array of strings (CIDs)


dholms:
Fixes https://github.com/fission-suite/ipfs-photo-gallery/issues/7

PR #122, opened 2019-10-03
add registration route

dholms:
# Problem

Users currently have to provision a Heroku instance to get a username and password.

# Solution

Added a /register route that provisions a user and returns their credentials

### Issues

`Fission.User.Mutate` can be cleaned up (some WET code), but before I do that, I was thinking it might be better to change the db schema slightly. Namely:
- remove `herokuAddOnId` col from Users table
- add `userID` col to heroku_add_ons table
- OR create a relationship table between the two

Reason being that herokuAddOn will no longer be on every user, and if we keep linking services in this way, we could see the user table blow up in size. Thoughts?

Closes https://github.com/fission-suite/web-api/issues/120


library/Fission/User/Mutate.hs
```diff
@@ -22,13 +26,28 @@ import qualified Fission.User.Table as Table
 
 -- | Create a new, timestamped entry
 create :: MonadRIO    cfg m
+       => MonadSelda      m
+       => HasLogFunc cfg
+       => SecretDigest
+       -> m (ID User)
+create sekret = do
```
bechurch: 💅 we should probably find a new name for `secret` during conflicts. perhaps be more verbose with something like "userSecret" or "secretKey"

library/Fission/User/Mutate.hs
```diff
@@ -22,13 +26,28 @@ import qualified Fission.User.Table as Table
...
+  now <- liftIO getCurrentTime
+
+  uID <- insertWithPK Table.users
+    [User def Regular True Nothing sekret <@ now]
```
bechurch: Knowledge share: could you explain to me what `def` does here?

library/Fission/Web/Register.hs
```diff
@@ -0,0 +1,47 @@
...
+  ( API
+  , register
+  ) where
+
+import           RIO
+
+import           Database.Selda as Selda
+
+import           Servant
+import Data.Has
```
bechurch: 💅 Needs some indentation by the looks of it!

library/Fission/Web/Register.hs
```diff
@@ -0,0 +1,47 @@
...
+  logInfo $ mconcat
+    [ "Provisioned user: "
+    , displayShow userID
+    ]
+  
+  return User.Config
+    { _interplanetaryFissionUrl      = url
+    , _interplanetaryFissionUsername = User.hashID userID
+    , _interplanetaryFissionPassword = Secret secret
+    }
```
bechurch: Question: Do we care here about new line at the EoF's?

library/Fission/Web/Swagger/Docs.hs
```diff
@@ -53,3 +54,8 @@ ping :: Swagger -> Swagger
 ping = applyTagsFor ops  ["Ping" & description ?~ "Check for liveness"]
   where
     ops = subOperations (Proxy :: Proxy Web.PingRoute) (Proxy :: Proxy Web.API)
+
+register :: Swagger -> Swagger
```
bechurch: Nice! 😍

library/Fission/User/Mutate.hs
dholms: That's actually residual from Brooke's code. I think "userSecret" is more clear as well. Curious to hear her thoughts

library/Fission/Web/Register.hs
dholms: unclear, but that does seem to be our standard. Good catch!

library/Fission/User/Mutate.hs
dholms: Not completely sure on this one tbh, but here's my understanding:
`def` is short for [default](https://hackage.haskell.org/package/selda-0.5.0.0/docs/Database-Selda.html#v:def)

Check the documentation for [Database.Selda.insert](https://hackage.haskell.org/package/selda-0.5.0.0/docs/Database-Selda.html#v:insert)
Here, `def` is passed in in place of `userID`. We don't want to pass in an actual userID (since it's auto-incrementing). So `def` tells Selda to default `userID`. In this case to `autoPrimary` (see `Fission.User.Table` for table definition)

library/Fission/User/Mutate.hs
expede: Yeah, sure, let's do it 👍 In this case it absolutely makes sense. For a lot of functions, verbose names harm reuse (which is something that you get a LOT more of in Haskell), but this function is pretty concrete ✅

library/Fission/User/Mutate.hs
expede: @dholms is correct 👍

library/Fission/Web/Heroku.hs
```diff
@@ -65,7 +65,7 @@ provision :: HasLogFunc      cfg
 provision Request {_uuid, _region} = do
   Web.Host url <- Config.get
   secret       <- liftIO $ Random.text 200
-  userID       <- User.create _uuid _region secret
+  userID       <- User.createWithHeroku _uuid _region secret
```
expede: Oooh that reads well. Great 👍👍👍👍👍👍

library/Fission/Platform/User.hs
```diff
@@ -1,5 +1,5 @@
-module Fission.Platform.Heroku.UserConfig
-  ( UserConfig (..)
+module Fission.Platform.User
```
expede: Hmm, thoughts on getting this out of `Platform` (platforms are like: Heroku, GitHub Apps, &c)? Maybe `Fission.User.Provision.Types`. You could still do this afterwards:

```haskell
import qualified Fission.User.Provision.Types as User
```

Thoughts?

library/Fission/Platform/User.hs
```diff
@@ -24,10 +24,10 @@ data UserConfig = UserConfig
              , Generic
              )
 
-makeLenses ''UserConfig
-$(deriveJSON lens_SCREAMING_SNAKE_CASE ''UserConfig)
+makeLenses ''Config
+$(deriveJSON lens_SCREAMING_SNAKE_CASE ''Config)
```
expede: FYI, I'm going to trash any `deriveJSON` that uses options. I really dislike the entire `Fission.Internal.JSON` module. Not for this PR, but happening eventually.

library/Fission/Platform/User.hs
```diff
@@ -15,7 +15,7 @@ import Data.Swagger as Swagger
 import Fission.Internal.JSON
 import Fission.Security
 
-data UserConfig = UserConfig
+data Config = Config
```
expede: 🤔 Perhaps `Provision` as in "provision response". You're right that we can use this to provision via the CLI (for example), not just Heroku. `Config` tends to  mean the `cfg` in `MonadRIO cfg m`. Thoughts?

library/Fission/User/Mutate.hs
expede: 👩‍🏫

> "userSecret" or "secretKey"

So, again, I fully agree that `userSecret` makes sense here. Zero argument.

We have more information than just the name: `secret' :: SecretDigest`. Not for this case, because digests make a lot of sense in many places, but if we want to be super clear, it is idiomatic to wrap in a `newtype` wrapper and have `s :: SecretUserKey` or similar.

library/Fission/User/Mutate.hs
bechurch: Awesome thanks @dholms !!

library/Fission/Platform/User.hs
dholms: Yeah makes a lot of sense :+1:

expede:
Quick read through looks great 👍 Will deep dive in a moment

library/Fission/Web.hs
```diff
@@ -94,3 +95,4 @@ server host' = Web.Swagger.server host'
           :<|> const Heroku.server
           :<|> const Auth.verify
           :<|> pure Ping.pong
+          :<|> Register.register
```
expede: 🤔 Maybe `Register.create`, `User.register`, `User.create`, or `Register.newUser`?

I feel like `User.create` is the clearest, but would love to hear your opinions 💭

library/Fission/Web/Swagger/Docs.hs
```diff
@@ -53,3 +54,8 @@ ping :: Swagger -> Swagger
...
+register = applyTagsFor ops  ["Register" & description ?~ "Register and provision a user"]
```
expede: Just took a look at this in Swagger, and it makes me think that we should rename this whole thing to `User` routes. Other than it's "standard" REST, an example of why: let's say we later allow users to add an email address. If they want to update that, we'll want them to make a `PUT` request to update the same resource. I don't think that it's conceptually a "registration" that they're updating; it's the user.

This would be a minor change to this PR: change `/register` to `/user`, update the module names, and all good.

Thoughts?

expede: Also, heads up: I'm going to scrap the `docs` vs `all_docs` distinction. Sure, consumers don't need to know what the Heroku endpoint looks like, but it really helps when you're running one of these yourself. I mention it here because the `/register` endpoint got added to `all_docs` but not `docs`. It's too easy to forget, and unnecessarily complicates the routes if we're not using it. I was going to DRY up some of that code, but this seems more sane.

expede:
> I noticed that the other day when I was going through the schema. I would be all for separating heroku from the Users table as much as possible. We are only going to become more Heroku agnostic overtime.

# Disclaimer

It's totally possible that I'm misreading the comment. I'm SUPER jet-lagged. Also, as always I want to emphasize that despite reams of reasoning, I do not intend to shut down discussion! Below is the reasoning for the current design, and I have been known to be wrong about things.

# TL;DR

Funny story — it was _initially_ broken out!

# Intuition

The intuition is one that I have every time that I see this type of relationship because I largely came up with `ActiveRecord` via Rails. It turns out that the AR pattern is actually pretty inefficient, and mirroring all of your application logic in the database fails separation of concerns (DB schema tightly coupled to application logic). Not that we're really doing much with the data _right now_, but speaking in generalities here.

# Appeal to Authority

The current design is per SQL best practices. Even the Ecto team recommends having multiple nullable columns rather than join tables (unless it's many-to-many), and especially recommends avoiding polymorphic tables. We won't have an unbounded number of relationships — it'll always be restricted to specific platforms that we integrate with, and they're all conceptually nullable one-to-ones.

# Design Space

Below are our schema design options. They're all isomorphic, but vary on how they describe the relationships semantically, and how they perform.

## Current Design

```
User
  * id
  * name
  * herokuFK
  * githubFK
  * otherthingFK
```

### Pros

* Efficient lookups
  * No join table
* Nullable one-to-one relationship enshrined directly
  * No special uniqueness indices required
* Super easy to tell how many integrations a user has

### Cons

* Wider table (marginally)
* Get back a couple fields that you may not care about (but don't have to select them)

## Integrations Table

This is really just breaking out the integrations into a separate table. If you stack the two tables next to each other, we have the same as in what we have now, just with some database indirection.

```
User
  * id
  * name

Integrations
  * userFK
  * herokuFK
  * githubFK
  * otherthingFK
```

### Pros

* Keeps integration info out of the user "object" if you're using active record

### Cons

* Denormalized (granted that's not _always_ a problem)
* More complex joins to get associations
* Adds an extra column to the DB (`userFK`)
* Needs an uniqueness index

## Polymorphic Table

```
User
  * id
  * name

Integrations
  * userFK
  * integrationFK
  * integrationEnum

IntegrationEnum = Heroku | GitHub | OtherThing
```

### Pros

* Keeps integration info out of the user "object" if you're using active record
* Save columns as the number of integrations grows (because of the enum)

### Cons

* Denormalized (not _always_ a problem, but no reason here)
* More complex joins to get associations
* Either need a stored procedure or a required multiple trips to figure out which concrete table this joins to
* Needs a more complex set of uniqueness indexes

# Heroku-agnostic

> We are only going to become more Heroku agnostic overtime.

Yes and no; we'll always allow users to sign up with Heroku. Already today the Heroku field is wrapped in a `Maybe` — the database mapper knows that this makes it a nullable field. Today we need to have a Heroku add-on to do anything, we won't as of this PR.

expede:
> Issues
>
> Fission.User.Mutate can be cleaned up (some WET code), but before I do that, I was thinking it might be better to change the db schema slightly. Namely:
>
>    remove herokuAddOnId col from Users table
>    add userID col to heroku_add_ons table
>    OR create a relationship table between the two
>
> Reason being that herokuAddOn will no longer be on every user, and if we keep linking services in this way, we could see the user table blow up in size. Thoughts?

See the above! You already don't _need_ a Heroku ID — the field is nullable

library/Fission/Web.hs
expede: In fact, if we make this change, then we should then also do `GET /user/verify` or similar for the current `/auth/verify` route 🤔

library/Fission/Web/Swagger/Docs.hs
dholms: yeah fan of this^ I was thinking after the fact of changing everything to `/auth` (ie `/auth/register`, `auth/verify`, etc) but I like `/user` better

dholms:
Agreed an integration tables would probably be overkill at this point, and hadn't thought through the fact that it denormalizes the database. My initial instinct was actually the opposite (was just thinking of it as a many-to-many table), but you're totally right

Point taken that putting the userFK in each integrations table will make it an increasingly complex query in order to return complete user information.

I'm good with leaving it as is for now. I'll refactor the WET code with the current schema

library/Fission/Web/User/Verify.hs
```diff
@@ -1,6 +1,6 @@
-module Fission.Web.Auth.Verify
+module Fission.Web.User.Verify
```
expede: Beautiful 👍

PR #123, opened 2019-10-04
Register cli

dholms:
# Problem

Users need to make API requests to provision an account which is not an intuitive flow.

# Solution

- Added a command to the CLI that hits the /register endpoint to provision a new user. 
- Save user credentials in `~/.fission.yaml`
- Check before making request to see if user already has credentials in `~/.fission.yaml`

Closes https://github.com/fission-suite/web-api/issues/121

_Note: this is a branch off of https://github.com/fission-suite/web-api/pull/122. Won't merge until that PR is merged_


library/Fission/CLI/Command/Register.hs
```diff
@@ -0,0 +1,77 @@
...
+                  $ Fission.Auth.register
+
+  logDebug $ displayShow registerResult
+
+  case registerResult of
+    Right user -> do 
+      logDebug $ displayShow user
+      let username = encodeUtf8 $ user ^. User.interplanetaryFissionUsername
+      let password = encodeUtf8 $ Security.unSecret $ user ^. User.interplanetaryFissionPassword
+      let auth = BasicAuthData username password
```
expede: Don't need a `let` on line! I mean, it works, but often is arranged like so:

```haskell
let
  username = encodeUtf8 $ user ^. User.interplanetaryFissionUsername
  password = encodeUtf8 $ Security.unSecret $ user ^.
  auth = BasicAuthData username password
```

i.e. "we're defining a bunch of local things here!"

library/Fission/CLI/Command/Register.hs
```diff
@@ -0,0 +1,77 @@
...
+      Auth.write auth >> CLI.Success.putOk "Registered & logged in"
```
expede: Since we're in a `do`, may as well put these on separate lines

Issue #116, opened 2019-09-26
Add /ipfs/peer (or similar)


expede:
# Summary

## Problem

Sometimes the CLI isn't able to connect to our IPFS instance

### Impact

Files may not be routable for an IPFS-native user / IPFS node. If they're connected to a common peer, then no problem, but we can't guarantee that.

### Server
Either Fission or the client should have the file, so the easiest thing to do is to just connect directly. 
1. Add an endpoint like `/ipfs/peerid` or similar that returns the IPv4 line of `ipfs id` on the server.
2. Update login to return `peers`

### CLI:
1. On login return the peer address
2. Write peer address to `fission.yml`
3. Use address in `ipfs swarm connect`
4. If there is a failure retrieve the address from `/peerid` and try again

PR #125, opened 2019-10-04
Add /ipfs/peers route to API

expede:
## Problem
The CLI need to know our servers external peer address so that they can connect to our swarm

## Solution
- [x] Add a new method that retrieves all externally accessible peers from `ipfs id`
- [x] Expose that data via a new `/ipfs/peers` route
- [x] Expose that data via a field in the response of the `/provision` route

## Questions for reviewers
1. Should this be a public or private route?
2. Given refactors in https://github.com/fission-suite/web-api/pull/123 the `peers` attribute on the `/provsion` response will have to be moved. Should I wait for that PR?

### Notes
1. While this is related to this issue #116 it doesn't close it as we still have to modify the CLI


expede:
@bechurch fixed the core issue!

expede:
The problem was `GET /ipfs/{cid}`, which was set to match first. So when we requested `/ipfs/peer` is was taking that as `/ipfs/{cid="peer"}`. Flipped the routes and all is well 👍

library/Fission/IPFS/Client.hs
```diff
@@ -42,6 +42,7 @@ cat   :: Text            -> ClientM File.Serialized
 pin   :: Text            -> ClientM Pin.Response
 unpin :: Text -> Bool     -> ClientM Pin.Response
 
+-- QUESTION: WHere the hell does this get called?
```
expede: Here's an example use: https://github.com/fission-suite/web-api/blob/master/library/Fission/CLI/Pin.hs#L43

library/Fission/IPFS/Client.hs
```diff
@@ -42,6 +42,7 @@ cat   :: Text            -> ClientM File.Serialized
...
 add :<|> cat
     :<|> pin
```
expede: 👆 Which uses this part of the API

library/Fission/IPFS/Client.hs
expede: (the `IPFS.Client.pin hash` part)

library/Fission/Internal/Debug.hs
```diff
@@ -0,0 +1,36 @@
...
+    _herokuID       = Hku.ID       "HEROKU_ID"
+    _herokuPassword = Hku.Password "HEROKU_PASSWORD"
+
+    _ipfsPath    = "/usr/local/bin/ipfs"
+    _ipfsURL     = IPFS.URL $ BaseUrl Http "localhost" 5001 ""
+    _ipfsTimeout = IPFS.Timeout 3600
+
+    _pgConnectInfo = PGConnectInfo "localhost" 5432 "web_api" Nothing Nothing Nothing
+
+    _host = "mycoolapp.io"
```
expede: I'm going to break out a couple of these in another branch. Can totes see this being useful. There's a couple things that I can do to make it faster than this full version.

library/Fission/Internal/Orphanage/OctetStream.hs
```diff
@@ -0,0 +1,14 @@
...
+import           RIO
+import qualified RIO.ByteString.Lazy as Lazy
+
+import Servant
+
+instance MimeRender OctetStream a => MimeRender OctetStream [a] where
+  mimeRender proxy values = "["<> meat <>"]"
+    where
+      meat :: Lazy.ByteString
+      meat =  Lazy.intercalate "," $ mimeRender proxy <$> values
```
expede: 🥪

library/Fission/User/Selector.hs
```diff
@@ -20,13 +20,13 @@ import Fission.User.Role
...
-secretDigest'  :: Selector User SecretDigest
-insertedAt'    :: Selector User UTCTime
-modifiedAt'    :: Selector User UTCTime
+-- userID'        :: Selector User (ID User)
+-- role'          :: Selector User Role
+-- active'        :: Selector User Bool
+-- herokuAddOnID' :: Selector User (Maybe (ID Heroku.AddOn))
+-- secretDigest'  :: Selector User SecretDigest
+-- insertedAt'    :: Selector User UTCTime
+-- modifiedAt'    :: Selector User UTCTime
```
expede: I think this commenting out is left over from a pairing session. The compiler is giving us a warning about missing signatures

library/Fission/Web/IPFS.hs
```diff
@@ -70,4 +72,16 @@ authed :: HasLogFunc        cfg
 authed usr = CID.allForUser usr
         :<|> Upload.add usr
         :<|> Pin.server usr
-        :<|> DAG.put usr
+        :<|> DAG.put usr -- Question: why is the function naming convention so all over the place?
+                         -- Answer: it's not! This is what they're called in IPFS!
```
expede: A fuller answer:

1. This isn't MVC — it's a tree of handlers. Think more Node than Rails.
2. We _could_ structure this as MVC, but it would mainly add boilerplate
  * I'm down to do that if it would help!
3. `server` is like a `controller`, but can be used as a standalone server for testing, handed to clients, &c. The full server is a series of handlers composed with `:<|>` the same way lists are composed with `:` (i.e. `|` in Elixir). Combine two servers with `:<|>`, and you get another server.

That said, if it would help to call it `controller` or `handler`, we can do that, too. Controllers work differently, so that may be additionally confusing.

A lot of these functions are named after what they do, which gives us reuse. The following example adds a layer of direction:

```haskell
module Fission.Storage.IPFS.CID

all = ...

allForUser usr = ...

----------

module Fission.Web.IPFS.CID.Controller where

index = CID.all

---------------

module Fission.Web.IPFS.User.CID.Controller where

index user = CID.allForUser user

------------

module Fission.Web.IPFS where

cidRoutes user = CID.index :<|> User.CID.index user

authed = cidRoutes
    :<|> ...
```

This _does_ give us a familiar interface (CRUD actions), but just adds a bunch of indirection to what could have just been calls to the underlying functions directly.

I'm open to having a conversation about that; it's tradeoffs. In either case we're _just composing functions_ with `:<|>`

library/Fission/Web/IPFS/Peer.hs
```diff
@@ -0,0 +1,61 @@
...
+        => HasLogFunc        cfg
+        => RIOServer         cfg API
+get = do
+  logError "STARTED FUNCTION"
+  getExternalAddress >>= \case
+  -- logInfo $ "GET ENDPOINT: " <> displayShow res
+    Right peers -> return peers
+    Left err -> do
+      logDebug "hellow world"
+      Web.Err.throw err
```
expede: With the change to route order; this is actually getting hit now 👍

library/Fission/IPFS/Peer.hs
```diff
@@ -54,5 +61,43 @@ connect peer@(Peer peerID) = IPFSProc.run ["swarm", "connect"] (UTF8.textToLazyB
...
+  & Text.unpack
+  & extractIPfromPeerAddress
+  & maybe "" Text.pack
+  & Net.IPv4.decode
+  <&> Net.IPv4.reserved
+  & maybe False not
+
+
+filterExternalPeers :: [Peer] -> [Peer]
+filterExternalPeers = Data.List.filter (isExternalIPv4 . peer)
```
bechurch: @expede Where would a good place to put util functions?

library/Fission/Internal/Debug.hs
bechurch: Are we still thinking about doing this? if so do you want me to kill these changes?

library/Fission/IPFS/Peer.hs
```diff
@@ -54,5 +62,45 @@ connect peer@(Peer peerID) = IPFSProc.run ["swarm", "connect"] (UTF8.textToLazyB
...
+  <&> Data.List.head
+
+isExternalIPv4 :: Text -> Bool
+isExternalIPv4 ip = ip
+  & T.unpack
+  & extractIPfromPeerAddress
+  & maybe "" T.pack
+  & Net.IPv4.decode
+  <&> Net.IPv4.reserved
+  & maybe False not
```
expede: Trying to get this to just pipe the whole way down is an interesting approach — I can't say that I even dislike it. It's Haskell with an Elixir accent, which I'm clearly not against at the author of Witchcraft 🔮

For science, below is more typical phrasing. The Haskell community really focuses on semantics (what a thing is) over mechanics (how it's done). Pipes really put you [into a mechanical frame of mind](https://youtu.be/fxe5NzHDxpk?t=1438).

There's a lot going on here, and it chunks into a couple phases. This is a more typical implementation:

WIP WIP WIP switching to editor

```haskell
isExternalIPv4 ip = maybe False not maybeIsReserved
  where
    maybeIsReserved :: Maybe Bool
    maybeIsReserved = IPv4.reserved <$> normalized
    
    normalized :: Maybe IPv4
    normalized = IPv4.decode . Text.pack ipAddress

    ipAddress = extractIPFromPeerAddress $ Text.unpack ip

isExternalIPv4 ip =
  case extractIPFromPeerAddress $ Text.unpack ip
```

library/Fission/IPFS/Peer.hs
```diff
@@ -3,14 +3,21 @@ module Fission.IPFS.Peer
...
 import qualified RIO.ByteString.Lazy as Lazy
 import qualified RIO.Text            as Text
 import           RIO.Process (HasProcessContext)
 
+import           Net.IPv4
+import           Text.Regex
+
 import           Data.Has
+import           Data.List hiding (all)
+import qualified Data.Text as T
```
expede: Protip: you have these functions already exported from `RIO.Text` (which I believe just re-exports `Data.Text`). As a general pattern, I tend to recommend using `Text` instead of `T` for clarity sake.

library/Fission/IPFS/Peer.hs
```diff
@@ -3,14 +3,21 @@ module Fission.IPFS.Peer
...
   , fission
+  , getExternalAddress
   ) where
 
 import           RIO hiding (all)
 import qualified RIO.ByteString.Lazy as Lazy
 import qualified RIO.Text            as Text
 import           RIO.Process (HasProcessContext)
 
+import           Net.IPv4
```
expede: Consider `import qualified Net.IPv4 as IPv4`. It's shorter, but importantly less noisy / more focused on the goal

```haskell
-- Before --

Net.IPv4.decode

-- After --

IPv4.decode
```

library/Fission/IPFS/Peer.hs
expede: Could alternately use [`Data.IPv4.isMatchedTo`](https://www.stackage.org/haddock/lts-14.7/iproute-1.7.7/Data-IP.html#v:isMatchedTo) (different library, relies on bitmasking).

library/Fission/IPFS/Peer.hs
```diff
@@ -54,5 +62,45 @@ connect peer@(Peer peerID) = IPFSProc.run ["swarm", "connect"] (UTF8.textToLazyB
...
+  , Has IPFS.BinPath cfg
+  , Has IPFS.Timeout cfg)
+  => m (Either Error [Peer])
+getExternalAddress = IPFSProc.run' ["id"] >>= \case
+    (ExitFailure _ , _, err) ->
+      return $ Left $ UnknownErr $ "BOOM:" <> UTF8.textShow err
+
+    (ExitSuccess , rawOut, _) -> do
+      let ipfsInfo = JSON.decode rawOut
+      let filteredAddrs = filterExternalPeers . maybe [] _addresses $ ipfsInfo
```
expede: 💅 

`let` on each line is completely fine, but it tends to be broken into a grouped assignment section:

```haskell
(ExitSuccess , rawOut, _) -> 
  let
    ipfsInfo = JSON.decode rawOut
    filteredAddrs = filterExternalPeers . maybe [] _addresses $ ipfsInfo

  logInfo $ displayShow ipfsInfo
  return $ Right filteredAddrs
```

It's basically saying "here's my variables ... and now here's how they're used"

library/Fission/IPFS/Peer.hs
```diff
@@ -54,5 +62,45 @@ connect peer@(Peer peerID) = IPFSProc.run ["swarm", "connect"] (UTF8.textToLazyB
...
+filterExternalPeers :: [Peer] -> [Peer]
+filterExternalPeers = Data.List.filter (isExternalIPv4 . peer)
+
+getExternalAddress :: (MonadIO m
+  , MonadReader cfg m
+  , HasProcessContext cfg
+  , HasLogFunc cfg
+  , Has IPFS.BinPath cfg
+  , Has IPFS.Timeout cfg)
+  => m (Either Error [Peer])
```
expede: 💅 

The linter will probably complain (and if it doesn't we should configure it to).

Either:

```haskell
getExternalAddress :: ( MonadRIO          cfg m
                      , HasProcessContext cfg
                      , HasLogFunc        cfg
                      , Has IPFS.BinPath  cfg
                      , Has IPFS.Timeout  cfg
                      ) 
                   => m (Either Error [Peer])

-- Or --

getExternalAddress :: MonadRIO          cfg m
                   => HasProcessContext cfg
                   => HasLogFunc        cfg
                   => Has IPFS.BinPath  cfg
                   => Has IPFS.Timeout  cfg
                   => m (Either Error [Peer])
```

I tend to use the second one because it's compact, but the first does make a clearer visual distinction between constraints and types.

library/Fission/User/Query.hs
```diff
@@ -9,5 +9,5 @@ import Fission.User.Types
 
 -- | Find a user by their account secret
 bySecret :: Text -> Row s User -> Col s Bool
-bySecret secret user = user `is'` #_active
-                   .&& user ! #_secretDigest .== text secret
+bySecret secret user = (user `is'` #_active)
+                   .&& (user ! #_secretDigest .== text secret)
```
expede: If we're going for parsing clarity, we can also do this:

```haskell
bySecret :: Text -> Row s User -> Col s Bool
bySecret secret user = isActive .&& matchesSecret
  where
    isActive      = user `is'` #_active
    matchesSecret = user ! #_secretDigest .== text secret
```

I don't mind either way; just outlining your options 😁  They're both totally readable.

library/Fission/IPFS/Peer.hs
```diff
@@ -54,5 +62,45 @@ connect peer@(Peer peerID) = IPFSProc.run ["swarm", "connect"] (UTF8.textToLazyB
   (ExitFailure _ , _, _) -> Left $ CannotConnect peer
   (ExitSuccess   , _, _) -> Right ()
 
+peerAddressRe :: Regex
+peerAddressRe = mkRegex "^/ip[46]/([a-zA-Z0-9.:]*)/"
+
+extractIPfromPeerAddress :: String -> Maybe String
+extractIPfromPeerAddress peer = peer
+  & matchRegex peerAddressRe
+  <&> Data.List.head
```
expede: This `head :: [a] -> a` function is one of the reasons that people have started using custom Preludes. The problem with `head` is that it's _partial_: it's `undefined` for `head []`. `RIO` has this function under `RIO.Partial`, which is essentially guilt-tripping you into not using it unless you're _sure_ that you won't run into it.

Since we're arleady in `Maybe`, we can rewrite this like in my comment below on `isExternalIPv4`:

```haskell
extractIPfromPeerAddress :: String -> Maybe String
extractIPfromPeerAddress peer = do
  matchedString <- matchRegex peerAddressRgx peer
  headMaybe matchedString

-- or shorter, if you're comfortable with >>= --

extractIPfromPeerAddress :: String -> Maybe String
extractIPfromPeerAddress peer = matchRegex peerAddressRgx peer >>= headMaybe

-- less idiomatic, but closer to the original version --

extractIPfromPeerAddress :: String -> Maybe String
extractIPfromPeerAddress peer = peer & matchRegex peerAddressRgx >>= headMaybe
```

library/Fission/Internal/Debug.hs
expede: Yeah, lemme do that right now

expede: Sorry, got distracted 😅 Here it is now https://github.com/fission-suite/web-api/pull/130

library/Fission/IPFS/Peer.hs
expede: If they're not going to be used elsewhere, you  can leave them in the module and not re-export them, or put it in a `where`. There's nothing wrong with having utility functions in the module itself, but if you wanted to split them out you can either use a generic `*.Util`, or something more semantic like `Fission.IPFS.Peer.External`.

library/Fission/IPFS/Peer.hs
bechurch: for right now the first one to me is both what comes out of my type signature generator and what is more readable (because of the visual difference between `=>` and `->`).

However I like how the second version looks. I would say lets keep the current convention :)

library/Fission/IPFS/Peer.hs
bechurch: Ahh yeah I like this much better ❤️

library/Fission/IPFS/Peer.hs
bechurch: Thanks for this detailed explanation @expede 

Haskell gives us a bunch of ways to write code. Which is nice for expressivity and trying to make code read as easily as possible.

I found this one to be my prefered method
```haskell
isExternalIPv4 :: Text -> Bool
isExternalIPv4 ip = maybe False not isReserved
  where
    isReserved :: Maybe Bool
    isReserved = do
      ipAddress  <- extractIPFromPeerAddress $ Text.unpack ip
      normalized <- IPv4.decode $ Text.pack ipAddress
      return $ IPv4.reserved normalized
```

so I think I'll go with that!

library/Fission/IPFS/Peer.hs
bechurch: > To help build the intuition of how <&> relates to >>=, notice that by switching to >>=, we stay on the happy path and get to use the inner "unwrapped" values. >>= is a superpowered &, where its superpower is keeping you on the happy path (only dealing with the Just values as if they weren't wrapped in a Just).

Thank you! Explicitly I knew that from functors in pictures (see plunger) however in practise I had forgot about it when constructing that pipe from earlier.

Anyhow
```haskell
extractIPfromPeerAddress :: String -> Maybe String
extractIPfromPeerAddress peer = matchRegex peerAddressRgx peer >>= headMaybe
```

reads the best imo

library/Fission/IPFS/Peer.hs
expede: Yes, it's my favourite as well!

bechurch:
@expede This is up and ready for review. I left a description as well as some questions for the team in the original PR comment

library/Fission/IPFS/Peer.hs
```diff
@@ -54,5 +61,46 @@ connect peer@(Peer peerID) = IPFSProc.run ["swarm", "connect"] (UTF8.textToLazyB
...
+-- | Get all external ipfs peer addresses
+getExternalAddress :: MonadRIO          cfg m
+                   => HasProcessContext cfg
+                   => HasLogFunc        cfg
+                   => Has IPFS.BinPath  cfg
+                   => Has IPFS.Timeout  cfg
+                   => m (Either Error [Peer])
+getExternalAddress = IPFSProc.run' ["id"] >>= \case
+    (ExitFailure _ , _, err) ->
+      return $ Left $ UnknownErr $ "BOOM:" <> UTF8.textShow err
```
expede: Leaving the `BOOM` in?

bechurch: Removed!

bechurch:
@expede back up for review

expede:
GH doesn't let me actually approve it until Travis finishes, but consider this as approval. Use your admin privileges while I'm asleep :P

PR #126, opened 2019-10-06
Expand & combine Swagger docs

### LINKED ISSUE #99: Invalid JSON in Swagger Examples
# Summary

## Problem

Swagger docs have broken example(s) with autogenerated Haskell `Show` instances (i.e. not valid JSON)

### Impact

Inaccurate / incomplete docs make it harder to understand how the service will work when interacting with it

## Solution

Write better JSON examples for for Swagger

### Example

Screen Shot 2019-09-13 at 16 06 41

expede:
* Remove `docs` vs `all_docs` distinction
* Add detail to several places in the Swagger docs
* Remove the now superfluous modules
* Get rid of JSON auto-derivation
  * More flexibility
  * Less magic
  * Very few LOC added
  * Removed an entire helper module
* Closes #99


library/Fission/Environment/Types.hs
```diff
@@ -25,8 +25,8 @@ makeLenses ''Environment
 
 instance FromJSON Environment where
   parseJSON = withObject "Environment" \obj -> do
-    _ipfs    <- parseJSON . Object =<< obj .: "ipfs"
-    _storage <- parseJSON . Object =<< obj .: "storage"
-    _web     <- parseJSON . Object =<< obj .: "web"
+    _ipfs    <- obj .: "ipfs"
+    _storage <- obj .: "storage"
+    _web     <- obj .: "web"
```
expede: Aeson is smart enough to know the expected type for these fields, and recursively parse. Totes my bad!

library/Fission/Platform/Heroku/Provision.hs
```diff
@@ -39,11 +38,27 @@ data Request = Request
...
+    ]
+
+instance FromJSON Request where
+  parseJSON = withObject "Heroku.Request" \obj -> do
+    _callbackUrl <- obj .: "callback_url"
+    _name        <- obj .: "name"
+    _plan        <- obj .: "plan"
+    _region      <- obj .: "region"
+    _uuid        <- obj .: "uuid"
+    return Request {..}
```
expede: Yeah, honestly these read well enough, despite the added boilerplate versus deriving them ourselves. Plus we get more flexibility, and the code is clearer than some abstract serializer on a generator, IMO.

library/Fission/Platform/Heroku/UserConfig.hs
```diff
@@ -1,33 +1,63 @@
...
+          ]
+      & example     ?~ toJSON userCfgEx
+      & description ?~
+        "The information that a user needs to know to access this service. Typically sent on provision"
+    where
+      userCfgEx = UserConfig
+        { _url      = Client.BaseUrl Client.Https "runfission.com" 443 ""
+        , _username = "c74bd95b8555275277d4"
+        , _password = Secret "GW0SHByPmY0.y+lg)x7De.PNmJvh1"
+        }
```
expede: Waaaaaaaaaaaaaaay more detail in the Swagger docs

library/Fission/User/CID/Types.hs
```diff
@@ -19,9 +17,4 @@ data UserCID = UserCID
              , Eq
              , Generic
              , SqlRow
-             )
-
-instance ToSchema (ID UserCID) where
-  declareNamedSchema _ =
-     return $ NamedSchema (Just "UserCID")
-            $ mempty & type_ ?~ SwaggerInteger
```
expede: This was unused

library/Fission/Web/IPFS.hs
```diff
@@ -41,11 +39,6 @@ type UnauthedAPI = "cids" :> CID.API
               :<|> Pin.API
               :<|> "dag" :> DAG.API
 
-type SimpleAPI = "cids" :> CID.API
-            :<|> Upload.Simple.API
-            :<|> Pin.API
-            :<|> "dag" :> DAG.API
```
expede: Moved to the only module that cares about it

library/Fission/Web/Routes.hs
```diff
@@ -24,9 +23,6 @@ type API = IPFSRoute
       :<|> AuthRoute
       :<|> PingRoute
 
-type PublicAPI = IPFSRoute
-            :<|> AuthRoute
```
expede: Removed because we just have `/docs` for everything now. No distinction required.

library/Fission/Web/Tls.hs
```diff
@@ -3,8 +3,9 @@ module Fission.Web.Tls
   , url
   ) where
 
-import Network.HTTP.Client
 import RIO
+
+import Network.HTTP.Client
 import Servant.Client
```
expede: Consistent with the rest of the app

library/Fission/Platform/Heroku/Provision.hs
```diff
@@ -135,9 +156,7 @@ instance ToSchema Provision where
         }
 
       cfgEx = Heroku.UserConfig
-        { _interplanetaryFissionUrl      = "https://hostless.dev"
-        , _interplanetaryFissionUsername = "c74bd95b8555275277d4"
-        , _interplanetaryFissionPassword = Secret "GW0SHByPmY0.y+lg)x7De.PNmJvh1"
+        { _url      = Client.BaseUrl Client.Https "runfission.com" 443 ""
```
bechurch: nice!

library/Fission/User/CID/Types.hs
```diff
@@ -19,9 +17,4 @@ data UserCID = UserCID
...
+             )
```
bechurch: 💅 new line at EOF

library/Fission/Web/IPFS.hs
bechurch: Newbie question:
Why is this true:
> keeps this module more server-focused

It seems like this file is fairly server focused?

expede:
@bechurch 

> Newbie question:
> Why is this true:
>
>     keeps this module more server-focused
> 
> It seems like this file is fairly server focused?

Exactly! I moved that chunk of code to the `Fission.Web.Client` module, since it's only existence is to provide a type without the multipart form endpoint for client generation. 

Servant can't autogenerate multipart clients at the moment because they require all kind of extra configuration, memory buffers, and whatnot. Some people are working on it, we don't actually need it (since we're not calling from inside a browser), and multipart only really solves for HTML forms, which we don't care about in our client.

expede:
@bechurch 

> The swagger syntax is easy enough to read however it looks weirdly hard to write. Is there some good docs for swagger with haskell?

Swagger has you write instantiate some nested records that include all of the descriptions, examples, and so on. It's a totally normal `data` type, and all of the types and fields are listed in the [top-level module on Stackage](https://www.stackage.org/haddock/lts-14.7/swagger2-2.4/Data-Swagger.html) (the main type is `Swagger`, which contains others like `Info` and `MimeList`). It's a schema, like nested JSON, and defined by the Swagger2 standard.

HOWEVER, my guess is that you're seeing lenses and going "what are all these funky looking operators? 😱" In short, they're getters and setters that let you work on nested data, enums, `Maybe`s, and so on. They compose nicely, so you can all kinds of mutable-feeling updates in a totally controlled way.

The good news is that once you know the lens library, it's used in a TON of places, so it's not wasted effort. You can really get by with knowing a handful of them, and following the patterns.

## Common Lens Legend

* `&` "and also" (the normal pipe / reverse application operator)

### Getters `.`

* `^.` get / lookup
* `.` go deeper / compose

### Setters `~`

* `.~` set / replace
* `?~` set a nullable field (i.e. a `Maybe`)
* `%~` update with a function (run function on current value and set it to that)
* `+~` add to the current value (e.g. counter)
* `-~` subtract from current value
* `*~` multiple current value by

These read nicely, and let you make deeply nested updates either with composition (`.`) or with a named lens that is a composition of others.

## Why not use record-update syntax?

You totally can! Doing this manually, you end up having to update each nested record recursively, including the updated sub-records as you go. For example:

```haskell
data Person = Person
  { _name    :: Text
  , _address :: Address
  , _pet     :: Maybe Pet
  }

data Pet = Pet
  { _name         :: Text
  , _variety      :: Animal
  , _favouriteToy :: Toy
  }

data Animal = Dog | Cat | Fish

data Toy = Toy 
  { _ name     :: Text
  , _condition :: Condition
  }

data Condition = New | Good | Fair | Bad

-------- NESTED RECORD ----------

owner = Person
  { _name = "Alice"
  , _address = "123 Fake Street"
  , _pet = Just $ Pet 
      { _name = "Fluffy"
      , _variety = Cat
      , _favouriteToy = Toy { _name = "wind-up mouse", _condition = New }
      }
  }

-------- MANUAL UPDATE ----------

wearAndTear = owner { _pet = pet { _favouriteToy = toy { _condition = Fair } } }
  where
    Person { _pet = pet@(Pet { _favouriteToy = toy })} = owner

-------- LENS UPDATE ----------

wearAndTear = owner & pet . favouriteToy . condition .~ Fair

-- or just the function part --

playWithPet = pet . favouriteToy . condition .~ Fair
playWithPet owner -- now the toy is worn down!

-------- LET'S BREAK THAT DOWN ----------

                           Compose path
                      pet.favouriteToy.condition
                                 |
                 Pipe    +--------------+        Replace
                   |     |              |           |
                   v     v              v           v
updateToy' = owner & pet . favouriteToy . condition .~ Fair
               ^       ^         ^           ^          ^
               |       |         |           |          |
               |       +---------+-----------+     Replace with
               |                 |
               |            Nested path
               |     (like in OO dot-notation)
               |
             Initial value    
```



---

Aside: porting `Control.Lens` to `Witchcraft` has been the #1 request that I've gotten (I just don't have the time).

library/Fission/Platform/Heroku/Provision.hs
dholms: Agreed. Much prefer being explicit here

bechurch:
@expede Thanks so much for the Lens Guide 😍

I feel like we should put that somewhere, what do you think about a section in the haskell for TS devs? I'd gladly put that as a chapter before config

bechurch:
@expede @dholms In trying to find the best way to share a proposed finished draft for the HfTS gitbook I accidentally merged it 😬 

https://haskellfortypescriptdevs.fission.codes/appendix-t#common-lens-legend

Let me know if you'd like me to remove or rework it.

p.s. this appendix is linked to from the only page we reference lenses:
https://haskellfortypescriptdevs.fission.codes/chapter-two-global-config#adding-a-new-field-to-the-config

expede:
Screen Shot 2019-10-10 at 18 44 54

library/Fission/CLI/Auth.hs
```diff
@@ -44,6 +43,6 @@ withAuth action = get >>= \case
   Left err -> do
     logError $ displayShow err
     liftIO $ ANSI.setSGR [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Red]
-    UTF8.putText $ Emoji.prohibited <> " Unable to read credentials. Try logging in with "
+    UTF8.putText "🚫 Unable to read credentials. Try logging in with "
```
expede: Turns out that the issue was emacs, not Haskell 😅 Got that fixed, and deleted the `Fission.Emoji` module.

library/Fission/IPFS/Info/Types.hs
```diff
@@ -1,4 +1,11 @@
-module Fission.IPFS.Info.Types (Info (..)) where
+module Fission.IPFS.Info.Types
+  ( Info (..)
+  , id'
+  , publicKey
+  , addresses
+  , agentVersion
+  , protocolVersion
```
expede: It was complaining about us not exporting the lenses. `id` conflicts with `RIO.id :: a -> a` which is a super fundamental function that really anyone would assume to be available... so switching to `id'`

PR #127, opened 2019-10-07
Flow 🌊

expede:
I've switch a few modules to use the [`Flow` library](https://www.stackage.org/haddock/lts-14.7/flow-1.0.18/Flow.html). I've used it before, and even wrote a [variant on it](http://hackage.haskell.org/package/flow-er) a few years ago. It's not _idiomatic_ Haskell, and the `|>` operator conflicts with something in `Control.Lens`, but it does provide a better visual metaphor than `$`, `&`, and `.` by pointing in the direction that the data flows.

Ben and Daniel will be familiar with the syntax from Elixir and Reason, respectively. The one difference versus _specifically Elixir_ is that it pipes to the **last** argument (not the first one).

Give it a look and tell me what you think. It's nonstandard, but it's really whatever makes the team happy, so I thought I'd whip it up.

### DIFF
--- a/fission-web-api.cabal
+++ b/fission-web-api.cabal
@@ -4,7 +4,7 @@ cabal-version: 2.2
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: cc8c6012f648eba2d65544e826ca27b053bc7ec9a49a3e5a25b5fc69bc0db024
+-- hash: b087f32db6030a0db3922527b9b062d816500d4d40a9ef66c384aff137308510
 
 name:           fission-web-api
 version:        1.10.0
@@ -88,6 +88,8 @@ library
       Fission.IPFS.Timeout.Types
       Fission.IPFS.Types
       Fission.IPFS.URL.Types
+      Fission.Log.Lens
+      Fission.Log.Types
       Fission.Monitor
       Fission.Monitor.Types
       Fission.Plan.Types
@@ -181,6 +183,7 @@ library
     , data-has
     , ekg-wai
     , envy
+    , flow
     , fsnotify
     , haskeline
     , http-client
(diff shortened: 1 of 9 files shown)


expede:
> really mean for us?

In code terms, it would mean that you would have to hide the conflicting function in `Control.Lens` every time you import it (or write an internal module that reexports everything except that).

```haskell
import Control.Lens hiding ((|>), (<|))
```

> what couldn't we do?

Nothing 🤷‍♀️ I guess communicate easily with other Haskellers. `.` and `$` are totally ubiquitous.

> how many tutorials would not make sense?

Depends on your bar for "make sense". Almost no one uses Flow, and so no docs, examples, &c use those operators. Reading _any_ other code, you'd have to translate back to `$` and `&`. Other Haskellers would need an adjustment period when reading our code, any talk slides would probaby need to be converted, and so on.

> I mean I'm obviously a fan of the |> operator but thats mostly because I'm familiar with it (even though it does communicate flow much better)

Yeah, I actually agree — it's such a great visual metaphor!

The _only_ thing that I don't like about it in the abstract is that it puts you into an operational frame of mind ("do this step, then this step" rather than "this means that").

> also im a huge fan of the TC39 pipe proposal.
As am I. They were going to add it to TypeScript a few years back, but then delayed when TC39 appeared so they could follow that standard. _Just make it happen already!_ 😛

dholms:
Really dig the visual metaphor here. Communicates a lot more than $/&/. I am a bit worried about modifying the language too much while still learning. I'd hate to get too used to a new way of doing things before I feel comfortable handling data flows in normal Haskell.

> This makes me want to have you stick with $ for a few weeks to get used to it first

I think I actually like this option the best. Learn the rules before we break them 

Also hadn't heard of TC39. Yes plz. Didn't realize how much I wanted this until right now

PR #130, opened 2019-10-08
Development Helpers

expede:
Some development helpers, especially for helping with REL development.

I'm also am getting back into the habit of writing docs:

Screen Shot 2019-10-08 at 21 05 18


package.yaml
```diff
@@ -141,6 +141,7 @@ executables:
     ghc-options:
       - -threaded
       - -flate-specialise
+      - -funbox-strict-fields
```
expede: While I was in there, I've been meaning to enable this flag. It more aggressively unboxes strict fields, like the `Config`.

bechurch: @expede what do you mean by:
> aggressively unboxes strict fields

library/Fission/Internal/Development.hs
```diff
@@ -0,0 +1,127 @@
...
+
+     == Example Use
+
+     > dbPool <- runSimpleApp $ connPool 1 1 3600 _pgConnectInfo
+     > processCtx <- mkDefaultProcessContext
+     > httpManager <- HTTP.newManager HTTP.defaultManagerSettings
+     > logOptions <- logOptionsHandle stdout True
+     > (logFunc, _ :: IO ()) <- newLogFunc $ setLogUseTime True logOptions
+     >
+     > let runSession = run logFunc dbPool processCtx httpManager
```
bechurch: 👍 Nice docs

library/Fission/Internal/Development.hs
```diff
@@ -0,0 +1,127 @@
...
+     > let run' = runRIO cfg { _ipfsPath = "~/Downloads/ipfs" }
+     > run' Fission.IPFS.Peer.all
+     > -- Right ["/ip4/3.215.160.238/tcp/4001/ipfs/QmVLEz2SxoNiFnuyLpbXsH6SvjPTrHNMU88vCQZyhgBzgw"]
+-}
+mkConfig :: IO (Config, IO ())
+mkConfig = do
+  let
+    _herokuID       = Hku.ID       "HEROKU_ID"
+    _herokuPassword = Hku.Password "HEROKU_PASSWORD"
+    _ipfsPath       = "/usr/local/bin/ipfs"
```
bechurch: Seeing a bunch of reused values for config in both `run` and `mkConfig` could we not move them to one record/function called by each?

package.yaml
expede: Haskell is lazy by default: nothing is evaluated until it's used. This is achieved by leaving a placeholder "box" pointer in place (i.e. Schrodinger's box: may be full or empty, but observing it forces it to become full).

You can tell GHC that some data should be strict, and on record fields that's the `!`. However, in many cases that placeholder pointer remains, which is a (small) layer of indirection. Turning this on says "anywhere that you _can_ toss that pointer, do so." The plain value in memory is now "unboxed".

library/Fission/Internal/Development.hs
expede: Yep, sure can! Given that they're placeholders for development, I didn't go back and clean them up, but yeah, fewer surprises are good. The easiest thing would probably be to break them into their own definitions in this module.

```haskell
ipfsPath :: FilePath
ipfsPath = "/usr/local/bin/ipfs" 
```

That kind of thing

expede: Okay, broke this out into `mkConfig` and `mkConfig'`

Convention: `foo'` is either used for inner values, or as "same as the non-tick version, but with some defaults set for you if you're feeling lazy".

Also `foo_` is convention for "foo, but we throw the result away".

expede:
Okay, factored that out. I can has merge?

expede:
Okay, factored that out. I can has merge? @dholms / @bechurch

library/Fission/Internal/Development.hs
bechurch: Hey @expede thanks for breaking those out I had more meant overall across each function in this file

For example `mkConfig` and `run` both use the following values
```haskell
    _herokuID       = Hku.ID       "HEROKU_ID"
    _herokuPassword = Hku.Password "HEROKU_PASSWORD"
    _ipfsPath    = "/usr/local/bin/ipfs"
    _ipfsURL     = IPFS.URL $ BaseUrl Http "localhost" 5001 ""
    _ipfsTimeout = IPFS.Timeout 3600
    _pgConnectInfo = pgConnectInfo
    _host = "mycoolapp.io"
```

could we not extract each of these the same way we do for `pgConnectInfo`

```haskell
pgConnectInfo :: PGConnectInfo
pgConnectInfo = PGConnectInfo "localhost" 5432 "web_api" Nothing Nothing Nothing

ipfsPath :: FilePath
ipfsPath = "/usr/local/bin/ipfs" 

-- &c
```

expede:
@bechurch 🤔 I'm not totally sure what you're hoping to do. Are you hoping to have some default values that you can use in production? Or just some development defaults to pass around single values to functions?

If the last one, then we have a few options for nice helpers, but we probably don't want to have them around for prod work.

expede:
> bnchrch
> whats so bad about abstracting a common default?
> 
> expede
> Nothing's bad about it per se
> Other than it'll be a bit wordier. I think I was initially feeling lazy and didn't see why you'd want those when you can like Heroku.ID "foo"
> But the main reason that I was asking was that there's a few ways to do this
> One is "monomorphic": write some values and export them
> The other is "polymorphic": there's a Default typeclass that lets you define default values for your types
> It's a bit dangerous, since we don't want those finding their way into the production app, and typing default everywhere is such an easy go-to that it could happen
> But then we could put development defaults right next to the record definition

bechurch:
@expede applied that changes. Since I dirtied my hands, @dholms would you give this the final approval :P

expede:
@bechurch Good quetsion! The `.cabal` file is deterministically generated from Stack. Stack is to Cabal what Yarn is to NPM. We wold `gitignore` it, but when this app was Dockerized it needed it to cache dependencies or something. There was some Docker reason. I just created a branch that gitignores it. If that passes CI, I'll merge it in and then we never have to worry about `.cabal` again

bechurch:
Good explanation! I approved that PR. Hopefully it passes CI and we can merge this all in :)

PR #131, opened 2019-10-09
Route53 automation with PUT /dns/{cid}

dholms:
# Problem

There is no way to register domains/update DNS through the web-api

# Solution
- /domain/{cid} POST route
- {username}.hostless.dev points to our gateway (ipfs.runfission.com)
        _dnslink.{username}.hostless.dev points to /ipfs/{cid}


dholms:
Still a WIP! But I'd love some feedback just so I know I'm on the right track.

I have a few questions that have popped up while coding. Mostly style stuff. Down to hop on a call to discuss or hear thoughts here:
- Should `Route53.createChangeRequest` be in monadic context? ie get zoneId from cfg in `registerDomain` and pass as params or leave it how it is? does it matter?
- AWS Config (AWS.Environment.Types) has some types from our code and some from the package. Is this an issue? Should I make alias types for AccessKey/SecretKey?
- ResourceId has no FromJSON instance. I wrote our own ZoneId type instead. Is that the best move or should I write the instance for ResourceId?
- better error handling in `Route53.registerDomain`? You can see [here](http://hackage.haskell.org/package/amazonka-route53-1.6.1/docs/Network-AWS-Route53-ChangeResourceRecordSets.html#t:ChangeResourceRecordSetsResponse) that the response isn't the most descriptive with errors. If anything goes wrong, the route still returns a 500 error. So I think we'd just be intercepting that and cleaning it up?

expede:
Just as I go through the comments and code

> You can see here that the response isn't the most descriptive with errors. If anything goes wrong, the route still returns a 500 error. 

We should be getting a nested record with the following fields:

```haskell
data ChangeResourceRecordSetsResponse = ChangeResourceRecordSetsResponse'
  { _crrsrsResponseStatus :: !Int -- HTTP Status
  , _crrsrsChangeInfo     :: !ChangeInfo
  } 

data ChangeInfo = ChangeInfo'
  { _ciComment     :: !(Maybe Text)
  , _ciId          :: !ResourceId
  , _ciStatus      :: !ChangeStatus
  , _ciSubmittedAt :: !ISO8601
  } 

data ChangeStatus = Insync | Pending
```

They're forcing you to use a custom constructor `changeResourceRecordSetsResponse`, which is typically done to stop you from doing something that the type system can't prevent (or it would be awkward to work with). A toy example of this is even numbers, which _can_ be enforced in types, but it's much easier to do:

```haskell
newtype Even = Even Int

mkEven :: Int -> Maybe Even
mkEven number = if rem number 2
                   then Just (Even number)
                   else Nothing
```

The annoying thing is that because they don't export the fields directly (to force you to use the smart constructor), you need to look them up via some helper (or look directly in the source).

Screen Shot 2019-10-09 at 17 26 16

...

Screen Shot 2019-10-09 at 17 27 21

It's actually fairly well documented, but there's a level of indirection here that I'm not thrilled with 😒

fission-web/Main.hs
```diff
@@ -38,13 +40,19 @@ main = do
   let
     Storage.Environment {..} = env ^. storage
     Web.Environment     {..} = env ^. web
+    AWS.Environment     {..} = env ^. aws
```
expede: 👍 Nice

fission-web/Main.hs
```diff
@@ -38,13 +40,19 @@ main = do
...
     _herokuPassword = Hku.Password . encodeUtf8 $ manifest ^. Hku.api ^. Hku.password
 
     _ipfsPath    = env ^. ipfs . binPath
     _ipfsURL     = env ^. ipfs . url
     _ipfsTimeout = env ^. ipfs . IPFS.timeout
+    
+    _awsAccessKey = env ^. aws . AWS.accessKey
+    _awsSecretKey = env ^. aws . AWS.secretKey
+    _awsZoneId    = env ^. aws . AWS.zoneId
+    _awsDomain    = env ^. aws . AWS.domain
```
expede: We're brought all of these into scope already above...

```haskell
AWS.Environment     {..} = env ^. aws
```

...so we don't need to do all of this lensing.

## Option 1

Remove the above `AWS.Environment {..}` line

## Option 2

```haskell
_awsAccessKey = _accessKey
_awsSecretKey = _secretKey
_awsZoneId    = _zoneId
_awsDomain    = _domain
```

library/Fission/AWS/Route53.hs
```diff
@@ -0,0 +1,77 @@
...
+    dnslinkUrl = "_dnslink" <> baseUrl
+    dnslink    = "dnslink=/ipfs/" <> (unaddress cid)
+
+  baseReq    <- createChangeRequest Cname baseUrl "ipfs.runfission.com"
+  dnslinkReq <- createChangeRequest Txt dnslinkUrl $ wrapQuotes dnslink
+
+  liftIO $ runResourceT . runAWST env . within NorthVirginia $ do
+    void $ send(baseReq)
+    void $ send(dnslinkReq)
+    return $ AWS.Domain baseUrl
```
expede: I know that this is a WIP, but I feel like we could break something like this into a helper:

```haskell
-- Helper --

runAWS = liftIO . runResourceT . runAWST

-- Use --

runAWS env $ within NorthVirginia do
  send baseReq
  send dnsLinkReq
  return $ AWS.Domain baseURL
```

expede: For context, we use `runResourceT` in the DB, too, but you never have to deal with it directly because it's tucked away in a helper.

library/Fission/AWS/Route53.hs
```diff
@@ -0,0 +1,77 @@
...
+
+createEnv :: MonadRIO           cfg m
+           => Has AWS.AccessKey cfg
+           => Has AWS.SecretKey cfg
+           => m(Env)
+createEnv = do
+  AccessKey accessKey <- Config.get
+  SecretKey secretKey <- Config.get
+  liftIO $ newEnv $ FromKeys (AccessKey accessKey) (SecretKey secretKey)
```
expede: 👍 Nice

library/Fission/AWS/Route53.hs
```diff
@@ -0,0 +1,77 @@
...
+    recordSet = resourceRecordSet domain recordType
+    updated   = addValue recordSet content
+    changes   = changeBatch $ toNonEmpty [change Upsert updated]
+    zone      = ResourceId zoneId
+  return $ changeResourceRecordSets zone changes
+
+addValue :: ResourceRecordSet -> Text -> ResourceRecordSet
+addValue recordSet value = do
+  let record = resourceRecord value
+  (recordSet & rrsResourceRecords ?~ toNonEmpty [record]) & rrsTTL ?~ (300 ::Natural)
```
expede: This should read a bit better:

```haskell
recordSet & rrsTTL ?~ (300 ::Natural)
          & rrsResourceRecords ?~ pure record
```

library/Fission/AWS/Route53.hs
```diff
@@ -0,0 +1,77 @@
...
+
+wrapQuotes :: Text -> Text
+wrapQuotes txt = "\"" <> txt <> "\""
```
expede: Good function! This may be useful in `Fission.Internal.UTF8`:

```haskell
wrapIn :: Text -> Text -> Text
wrapIn txt wrapper = wrapper <> txt <> wrapper

-- then at the call site --

wrapIn "hello world" "\""

-- or infix that reads like English --

"hello world" `wrapIn` "\""
```

library/Fission/AWS/Types.hs
```diff
@@ -0,0 +1,44 @@
...
+
+import RIO
+
+import Data.Aeson
+import Data.Swagger (ToSchema (..))
+import Servant
+
+import qualified Fission.Internal.UTF8 as UTF8
+
+newtype Domain = Domain { getDomain :: Text }
```
expede: At risk of being pedantic, we may want to be explicit here: `DomainName`. There are other things with "domain" in the name that are part of DNS: "network domain", "domain controller", and so on

library/Fission/AWS/Types.hs
```diff
@@ -0,0 +1,44 @@
...
+           , Generic
+           , Show
+           , Ord
+           )
+  deriving anyclass ( ToSchema )
+  deriving newtype  ( IsString )
+
+instance FromJSON ZoneId where
+  parseJSON = withText "AWS.ZoneId" \txt ->
+    ZoneId <$> parseJSON (String txt)
```
expede: This file is lit 🔥👍

library/Fission/Config/Types.hs
```diff
@@ -40,6 +42,10 @@ data Config = Config
   , _dbPool         :: !DB.Pool
   , _herokuID       :: !Heroku.ID
   , _herokuPassword :: !Heroku.Password
+  , _awsAccessKey   :: !AWS.AccessKey
+  , _awsSecretKey   :: !AWS.SecretKey
+  , _awsZoneId      :: !AWS.ZoneId
+  , _awsDomain      :: !AWS.Domain
```
expede: 👍

library/Fission/Config/Types.hs
```diff
@@ -58,6 +64,10 @@ instance Show Config where
     , "  _dbPool         = " <> show _dbPool
     , "  _herokuID       = " <> show _herokuID
     , "  _herokuPassword = " <> show _herokuPassword
+    , "  _awsAccessKey   = " <> show _awsAccessKey
+    , "  _awsSecretKey   = HIDDEN"
+    , "  _awsZoneId      = " <> show _awsZoneId
+    , "  _awsDomain      = " <> show _awsDomain
```
expede: 👍

library/Fission/Web.hs
```diff
@@ -94,3 +106,4 @@ server host' = Web.Swagger.server host'
           :<|> const Heroku.server
           :<|> const Auth.verify
           :<|> pure Ping.pong
+          :<|> Domain.server
```
expede: 🚀

library/Fission/Web/Domain.hs
```diff
@@ -0,0 +1,31 @@
+module Fission.Web.Domain
```
expede: Maybe `Fission.Web.DNS`? 🤔

library/Fission/Web/Domain.hs
```diff
@@ -0,0 +1,31 @@
...
+server :: Has AWS.AccessKey cfg
+       => Has AWS.SecretKey cfg
+       => Has AWS.ZoneId    cfg
+       => Has AWS.Domain    cfg
+       => User
+       -> RIOServer         cfg API
+server User { _userID } cid = do
+  let subdomain = User.hashID _userID
+  res <- Route53.registerDomain subdomain cid
+  return res
```
expede: I know that this is a WIP, but I'm going to mention it anyways.

The linter will also give you this as a suggestion: any time you see this pattern

```haskell
do
  -- ...
  foo <- runBar
  return foo
```

you can skip the ceremony and just write

```haskell
do
 -- ...
 runBar
```

expede: I'm not sure if desugaring makes it clearer or not here, but:

```haskell
runBar >>= \foo -> return foo

-- same as --

runBar
```

GHC is smart enough to rewrite this for you, so there's not runtime impact. One of the advantages of all this pure function stuff is the compiler can be VERY aggressive with rewriting functions to be more efficient ⚡

expede: TL;DR

```haskell
server User { _userID } cid = do
  let subdomain = User.hashID _userID
  Route53.registerDomain subdomain cid
```

Could also be:

```haskell
server User { _userID } cid = Route53.registerDomain subdomain cid
  where subdomain = User.hashID _userID
```

But same/same

library/Fission/Web/Routes.hs
```diff
@@ -41,3 +43,7 @@ type IPFSRoute  = IPFSPrefix :> IPFS.API
 type IPFSPrefix = "ipfs"
 
 type PingRoute = "ping" :> Ping.API
+
+type DomainRoute = "domain" 
+                  :> BasicAuth "existing user" User
+                  :> Domain.API
```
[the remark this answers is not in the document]
expede: 🤔 We may want to factor out `BasicAuth "existing user" User`. Perhaps:

```haskell
module Fission.Web.Auth
-- ...
type ExistingUser = BasicAuth "existing user" User

----------

type AuthRoute = "auth"
               :> "verify"
               :> Auth.ExistingUser
               :> Auth.Verify.API


type DomainRoute = "domain" 
                  :> Auth.ExistingUser
                  :> Domain.API
```

Thoughts?

library/Fission/Web/Domain.hs
```diff
@@ -0,0 +1,31 @@
...
+import           Fission.Web.Server
+import           Fission.User        as User
+import           Fission.IPFS.CID.Types
+
+import qualified Network.AWS.Auth    as AWS
+import qualified Fission.AWS.Types   as AWS
+import qualified Fission.AWS.Route53 as Route53
+
+type API = Capture "cid" CID
+        :> Post    '[PlainText, OctetStream] AWS.Domain
```
expede: This should be a `PUT`, no? We have the ID, we're not creating a new one

expede:
@dholms This is looking really good! This is not the world's most straightforward task: this library is huge, Route53 is a _beast_ (AWS in general, really). You're rolling up your sleeves and jumping in and that's super awesome 🎉🎉🎉🎉🎉🎉

library/Fission/AWS/Route53.hs
dholms: Great call. much more readable

library/Fission/Web/Domain.hs
dholms: Well we're upserting a DNS record. So in some cases we're creating a new DNS record

library/Fission/Web/Domain.hs
dholms: Yeah I had that thought yesterday as well. This confirms it so I'll change it out :ok_hand:

library/Fission/Web/Domain.hs
dholms: That last version :heart_eyes: 

Also haven't been using a linter :grimacing: Gonna get that set up now. [hlint](https://hackage.haskell.org/package/hlint) seems to be the go to?

library/Fission/Web/Routes.hs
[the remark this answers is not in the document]
dholms: Big fan. I'll do the same with `"heroku add-on api"` for consistency

library/Fission/AWS/Route53.hs
dholms: I'm having an awful time with the types on this one... May need to run it by you at some point

expede: @dholms sure thing!

library/Fission/Web/Domain.hs
expede: Sure, but that's a VERY common misconception about `PUT` and `POST`

> It's quite possible, valid and even preferred in some occasions, to use PUT to create resources, or use POST to update resources.

[Source](http://restcookbook.com/HTTP%20Methods/put-vs-post/)

* `POST`
  * You don't have a primary key yet
  * Calling multiple times may yield different results
* `PUT`
  * You know the key of the thing you're updating
  * Idempotent

It's not about of the record has been persisted previously; it's about how you reference the thing you're talking about. We know the key of the DNS record: it's their username! I believe that this would be idempotent, too, no?

A little reference table:

HTTP Method | Idempotent | Safe
-- | -- | --
OPTIONS | yes | yes
GET | yes | yes
HEAD | yes | yes
PUT | yes | no
POST | no | no
DELETE | yes | no
PATCH | no | no

[Source](http://restcookbook.com/HTTP%20Methods/idempotency/)

library/Fission/AWS/Route53.hs
```diff
@@ -0,0 +1,73 @@
...
+    return . parseStatus $ res ^. crrsrsResponseStatus
+
+createEnv :: MonadRIO           cfg m
+           => Has AWS.AccessKey cfg
+           => Has AWS.SecretKey cfg
+           => m(Env)
+createEnv = do
+  AccessKey accessKey <- Config.get
+  SecretKey secretKey <- Config.get
+  liftIO $ newEnv $ FromKeys (AccessKey accessKey) (SecretKey secretKey)
```
expede: Can also give a type hint to avoid unwrapping / rewrapping

```haskell
accessKey :: AccessKey <- Config.get
secretKey :: SecretKey <- Config.get
liftIO $ newEnv $ FromKeys accessKey secretKey
```

library/Fission/AWS/Route53.hs
```diff
@@ -0,0 +1,73 @@
...
+
+addValue :: ResourceRecordSet -> Text -> ResourceRecordSet
+addValue recordSet value = recordSet & rrsTTL ?~ 300
+                                     & rrsResourceRecords ?~ pure(resourceRecord value)
+
+parseStatus :: Int -> Maybe ServerError
+parseStatus status
+  | status >= 500 = Just err500 { errBody = "Unkown AWS Error" }
+  | status >= 400 = Just err404 { errBody = "Could not locate AWS Resource" }
+  | otherwise = Nothing
```
expede: 👍

library/Fission/Web/Domain.hs
dholms: The more you know! Yup should definitely be PUT

library/Fission/Web/DNS.hs
```diff
@@ -0,0 +1,47 @@
...
+       => Has AWS.DomainName cfg
+       => User
+       -> RIOServer         cfg API
+server User { _userID } cid = do 
+  domain :: AWS.DomainName <- Config.get
+  let
+    username = User.hashID _userID
+    baseUrl = username <> AWS.getDomainName domain
+    dnslinkUrl = "_dnslink." <> baseUrl
+    dnslink = "dnslink=/ipfs/" <> unaddress cid
```
expede: protip:

```haskell

server User { _userID } (CID hash) = do
-- ...
dnslink = "dnslink=/ipfs/" <> hash
```

library/Fission/Web/Routes.hs
```diff
@@ -29,10 +31,14 @@ type UserRoute = "user" :> User.API
 
 type HerokuRoute = "heroku"
                    :> "resources"
-                   :> BasicAuth "heroku add-on api" ByteString
+                   :> Auth.HerokuAddOnAPI
```
expede: I literally said out loud to myself "oooh, that's good"

expede:
Checkpointing before pushing rebase 
Git hash: `21ee546`

library/Fission/AWS/Environment/Types.hs
```diff
@@ -0,0 +1,46 @@
...
+    , "  _domainName = " <> show _domainName
+    , "}"
+    ]
+
+instance FromJSON Environment where
+  parseJSON = withObject "AWS.Environment" \obj -> do
+    _accessKey  <- obj .:? "accessKey"  .!= ""
+    _secretKey  <- obj .:? "secretKey"  .!= ""
+    _zoneId     <- obj .:? "zoneId"     .!= ""
+    _domainName <- obj .:? "domainName" .!= ""
```
expede: If these fields are actually nullable, they should be `Maybe`s in the data declaration, not empty strings in the parser.

library/Fission/AWS/Route53.hs
```diff
@@ -0,0 +1,73 @@
...
+           -> Text
+           -> m(Maybe ServerError)
+registerDomain recordType domain content = do 
+  logDebug $ "Updating DNS record at: " <> displayShow domain
+  env <- createEnv
+  req <- createChangeRequest recordType domain content
+
+  liftIO . runResourceT . runAWST env . within NorthVirginia $ do
+    res <- send req
+    return . parseStatus $ res ^. crrsrsResponseStatus
```
expede: Ohhhhh, you haven't pushed your changes from earlier 😅 This is going to not be fun with the rebase. I may take a swing at recreating it manually here.

library/Fission/AWS.hs
```diff
@@ -0,0 +1,50 @@
...
+  secretKey :: SecretKey <- Config.get
+  liftIO $ newEnv $ FromKeys accessKey secretKey
+
+validate :: ChangeResourceRecordSetsResponse -> Either ServerError ChangeResourceRecordSetsResponse
+validate changeSet
+  | status >= 500 = Left err500 { errBody = "Unkown AWS Error" }
+  | status >= 400 = Left err404 { errBody = "Could not locate AWS Resource" }
+  | otherwise = Right changeSet
+  where
+    status = changeSet ^. crrsrsResponseStatus
```
expede: @dholms I moved this to this module

library/Fission/AWS.hs
```diff
@@ -0,0 +1,50 @@
...
+ensureContent
+  :: MonadRIO   cfg m
+  => MonadThrow     m
+  => Exception err
+  => m (Either err ChangeResourceRecordSetsResponse)
+  -> m ChangeResourceRecordSetsResponse
+ensureContent runRequest = do
+  errOrResp <- runRequest
+  resp      <- ensureM errOrResp
+  ensureM $ validate resp
```
expede: Multi-level error handling

library/Fission/AWS/Route53.hs
```diff
@@ -0,0 +1,58 @@
...
+               -> Text
+               -> m (Either ServerError ChangeResourceRecordSetsResponse)
+registerDomain recordType domain content = do
+  logDebug $ "Updating DNS record at: " <> displayShow domain
+  env <- createEnv
+  req <- createChangeRequest recordType domain content
+
+  withAWS env NorthVirginia $ do
+    res <- send req
+    return $ validate res
```
expede: Something in me wants to make this shorter

```haskell
 withAWS env NorthVirginia . pure $ validate =<< send req
```

But the current version reads more clearly than that, so 👍

library/Fission/AWS/Types.hs
```diff
@@ -0,0 +1,44 @@
...
+
+import RIO
+
+import Data.Aeson
+import Data.Swagger (ToSchema (..))
+import Servant
+
+import qualified Fission.Internal.UTF8 as UTF8
+
+newtype DomainName = DomainName { getDomainName :: Text }
```
expede: 👍

library/Fission/Web/DNS.hs
```diff
@@ -0,0 +1,48 @@
...
+        :> PutAccepted '[PlainText, OctetStream] AWS.DomainName
+
+server :: HasLogFunc         cfg
+       => Has AWS.AccessKey  cfg
+       => Has AWS.SecretKey  cfg
+       => Has AWS.ZoneId     cfg
+       => Has AWS.DomainName cfg
+       => User
+       -> RIOServer         cfg API
+server User { _userID } (CID hash) = do
```
expede: Protip: you can pattern match on `CID hash` here instead of using `unaddress` in the body of the function

library/Fission/AWS/Environment/Types.hs
dholms: hmmm what do you think is the best move here @expede? All these values are required for the AWS api to work, but I guess the server can still make sense to run without hooking into AWS. So make them `Maybe`s and then just send errors if someone tries to hit an AWS route without having these env vars?

expede:
```haskell
data Environment = Environment
  { _accessKey  :: !(Maybe AWS.AccessKey)  -- ^ AWS Access Key
  , _secretKey  :: !(Maybe AWS.SecretKey)  -- ^ AWS Secret Key
  , _zoneId     :: !(Maybe AWS.ZoneId)     -- ^ AWS Hosted Zone
  , _domainName :: !(Maybe AWS.DomainName) -- ^ AWS Domain
  }

makeLenses ''Environment

instance Show Environment where
  show Environment {..} = intercalate "\n"
    [ "Environment {"
    , "  _accessKey  = " <> show _accessKey
    , "  _secretKey  = HIDDEN"
    , "  _zoneId     = " <> show _zoneId
    , "  _domainName = " <> show _domainName
    , "}"
    ]

instance FromJSON Environment where
  parseJSON = withObject "AWS.Environment" \obj -> do
    _accessKey  <- obj .:? "accessKey"  
    _secretKey  <- obj .:? "secretKey"  
    _zoneId     <- obj .:? "zoneId"     
    _domainName <- obj .:? "domainName" 

    return $ Environment {..}
```

library/Fission/AWS.hs
```diff
@@ -0,0 +1,52 @@
...
+          => m Env
+createEnv = do
+  accessKey :: AccessKey <- Config.get
+  secretKey :: SecretKey <- Config.get
+  liftIO $ newEnv $ FromKeys accessKey secretKey
+
+validate :: ChangeResourceRecordSetsResponse -> Either ServerError ChangeResourceRecordSetsResponse
+validate changeSet =
+  if status >= 300
+    then Left $ toServerError status
```
expede: 🤔 300s are redirects and caching. I guess it can't hurt to make that an error, because the HTTP client should have hit that first and I would not expect a 300 to get this far into the app.

library/Fission/AWS/Environment/Types.hs
```diff
@@ -0,0 +1,46 @@
...
+    , "  _domainName = " <> show _domainName
+    , "}"
+    ]
+
+instance FromJSON Environment where
+  parseJSON = withObject "AWS.Environment" \obj -> do
+    _accessKey  <- obj .: "accessKey"
+    _secretKey  <- obj .: "secretKey"
+    _zoneId     <- obj .: "zoneId"
+    _domainName <- obj .: "domainName"
```
expede: 👍

library/Fission/AWS/Types.hs
```diff
@@ -0,0 +1,44 @@
...
+  mimeRender _ = UTF8.textToLazyBS . getDomainName
+
+instance MimeRender OctetStream DomainName where
+  mimeRender _ = UTF8.textToLazyBS . getDomainName
+
+newtype ZoneId = ZoneId { getZoneId :: Text }
+  deriving ( Eq
+           , Generic
+           , Show
+           , Ord
```
expede: Does this need to be orderable (i.e. `ZoneId foo > ZoneId bar`?) Slapping an `Ord` on the instance can't hurt, but these add a bit of compile overhead. We've been moving fast and throwing instance generation at everything, but may be good to start getting them under control

library/Fission/Environment/Types.hs
```diff
@@ -14,11 +15,13 @@ import Control.Lens (makeLenses)
 import qualified Fission.IPFS.Environment.Types    as IPFS
 import qualified Fission.Storage.Environment.Types as Storage
 import qualified Fission.Web.Environment.Types     as Web
+import qualified Fission.AWS.Environment.Types     as AWS
 
 data Environment = Environment
   { _ipfs    :: !IPFS.Environment    -- ^ IPFS configuration
   , _storage :: !Storage.Environment -- ^ Storage/DB configuration
   , _web     :: !Web.Environment     -- ^ Web configuration
+  , _aws     :: !AWS.Environment     -- ^ Web configuration
```
expede: Web config -> AWS config

expede:
Rebased

library/Fission/AWS/Environment/Types.hs
```diff
@@ -0,0 +1,46 @@
...
+    , "  _domainName = " <> show _domainName
+    , "}"
+    ]
+
+instance FromJSON Environment where
+  parseJSON = withObject "AWS.Environment" \obj -> do
+    _accessKey  <- obj .: "access_key"
+    _secretKey  <- obj .: "secret_key"
+    _zoneID     <- obj .: "zone_id"
+    _domainName <- obj .: "domain_name"
```
expede: FYI, made these look like standard YAML

expede: And added them to the `env.yaml.example`

expede:
Okay, tested it myself locally. Works maybe _too_ well 😉

Merging...

PR #134, opened 2019-10-10
update dns Up and Watch cmds from CLI

dholms:
# Problem
CLI does not currently register domains after pinning content

# Solution
After pinning content to remote node, send a `/dns` PUT request with the pinned CID to update DNS


dholms:
Biggest question here is how I'm doing error handling. Can't tell if it's smart or over-engineered.

Also tried to search around for how to put clickable links in the terminal, but didn't see much on first pass. Any ideas?

expede:
@dholms I previously deleted the `Fission.Emoji` module. Turns out that you _can_ just use emoji directly, but emacs was messing them up. I added the emoji package to emacs and they suddenly worked, so no need for `Emoji.foo`

expede:
This could use a rebase

library/Fission/AWS/Types.hs
```diff
@@ -30,6 +31,12 @@ instance MimeRender PlainText DomainName where
 instance MimeRender OctetStream DomainName where
   mimeRender _ = UTF8.textToLazyBS . getDomainName
 
+instance MimeUnrender PlainText DomainName where
+  mimeUnrender _proxy bs =
+    case decodeUtf8' $ Lazy.toStrict bs of
+      Left err  -> Left $ show err
+      Right txt -> Right $ DomainName txt
```
expede: This is readable, so purely for educational purposes:

If we didn't have two constructors to get inside of here, we could have used `fmap`, right?

```haskell
fmap DomainName . decodeUtf8' $ Lazy.toStrict bs -- assuming always `Right`
```

There's a class that's like `Functor`, but for this situation: [`Bifunctor`](https://www.stackage.org/haddock/lts-14.8/base-4.12.0.0/Data-Bifunctor.html), which defined [`bimap`](https://www.stackage.org/haddock/lts-14.8/base-4.12.0.0/Data-Bifunctor.html#v:bimap)

```haskell
bimap show DomainName . decodeUtf8' $ Lazy.toStrict bs
```

That does the same as the `case` version. Now to be clear, I use the `case` version essentially always, because it's very easy to see what's happening. Bifunctor

library/Fission/CLI/Auth.hs
```diff
@@ -35,10 +45,14 @@ cachePath = do
...
-withAuth :: (MonadRIO cfg m, HasLogFunc cfg) => (BasicAuthData -> m ()) -> m ()
+withAuth :: (MonadRIO cfg m, HasLogFunc cfg)
+         => (BasicAuthData -> m (Either ClientError a))
+         -> m (Either CLIError a)
 withAuth action = get >>= \case
   Right auth ->
-    action auth
+    action auth >>= \case
+      Right result -> return $ Right result
+      Left err -> return . Left . CLIError . UTF8.textShow $ err
```
expede: If you want to, those `return`s can be moved before the `case`. This is specifically because it's happening on both branches:

```haskell
action auth >>= pure . \case
  Right result -> Right result
  Left err -> Left . CLIError . UTF8.textShow $ err
```

Which, per the `Bifunctor` comment above could also be:

```haskell
action auth >>= pure . first (CLIError . UTF8.textShow)
```

...but we're not playing code golf 🏌️‍♀️

library/Fission/CLI/Auth.hs
```diff
@@ -47,3 +61,4 @@ withAuth action = get >>= \case
     liftIO $ ANSI.setSGR [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Blue]
     UTF8.putText "fission-cli login"
     liftIO $ ANSI.setSGR [ANSI.Reset]
+    return . Left . CLIError . UTF8.textShow $ err
```
expede: Looks like we use `return . Left . CLIError . UTF8.textShow $ err` twice in this function; it may be a sign that this is a reusable chunk:

```haskell
foo = Left . CLIError . UTF8.textShow
```

library/Fission/CLI/DNS.hs
```diff
@@ -0,0 +1,52 @@
...
+    -> BasicAuthData
+    -> m (Either ClientError AWS.DomainName)
+update cid@(CID hash) auth = do
+  logDebug $ "Updating DNS to " <> display hash
+
+  Client.Runner runner <- Config.get
+  update' runner auth cid >>= \case
+    Right domain -> do
+      CLI.Success.dnsUpdated $ AWS.getDomainName domain
+      return $ Right domain
```
expede: Protip:

```haskell
Right wrapped@(DomainName domain) -> do
  CLI.Success.dnsUpdated domain
  return $ Right wrapped
```

Or with better naming. You can pattern match is what I'm saying 😛

expede:
> Biggest question here is how I'm doing error handling. Can't tell if it's smart or over-engineered.

@dholms 

Yeah, it's not so much over-engineered as it's _noisy._ In Haskell that's a good sign that you're missing an abstraction. In almost all cases, we should be able to get automated behaviour if we're writing something repetitive — the main question is if it's worth the effort (hint: it is in this case IMO).

I took some time and wrote some instances (and a couple helper functions). This is how this reads now, and yet behaves the exactly same way as before:

```haskell
up = runCLI_ do
  logDebug "Starting single IPFS add locally"
  dir <- liftIO getCurrentDirectory
  cid <- liftE $ IPFS.addDir dir
  liftE $ Auth.withAuth (CLI.Pin.run cid)
  liftE $ Auth.withAuth (CLI.DNS.update cid)
```

library/Fission/AWS/Types.hs
dholms: oh nice that's cool ^^
Very useful to see it done both ways, but yeah with the `case` version it's definitely more immediately apparent what's going on

library/Fission/CLI/Auth.hs
dholms: HLint doesn't like that first option (still compiles fine), and recommends:
```haskell
    \case
      Right result -> Right result
      Left err -> Left . CLIError . UTF8.textShow $ err
      <$> action auth
```
which I'm not crazy about because the flow isn't super clear

library/Fission/CLI/Auth.hs
```diff
@@ -14,11 +15,20 @@ import qualified System.Console.ANSI as ANSI
 
 import qualified Data.Yaml as YAML
 import           Servant
+import           Servant.Client
 
 import           Fission.Internal.Constraint
 import           Fission.Internal.Orphanage.BasicAuthData ()
 import qualified Fission.Internal.UTF8 as UTF8
 
+newtype CLIError = CLIError { message :: Text }
```
expede: For consistency, let's:

1. Move this to it's own `*.Error` module
2. More structured errors, if that's useful

I'm going to just... do this? At worst, we can roll it back

library/Fission/CLI/Auth.hs
```diff
@@ -35,10 +45,14 @@ cachePath = do
   home <- getHomeDirectory
   return $ home </> ".fission.yaml"
 
-withAuth :: (MonadRIO cfg m, HasLogFunc cfg) => (BasicAuthData -> m ()) -> m ()
+withAuth :: (MonadRIO cfg m, HasLogFunc cfg)
+         => (BasicAuthData -> m (Either ClientError a))
+         -> m (Either CLIError a)
```
expede: Nice 👍

library/Fission/CLI/Display/Success.hs
```diff
@@ -12,11 +13,16 @@ import qualified Fission.Internal.UTF8 as UTF8
...
 putOk :: MonadIO m => Text -> m ()
 putOk msg = do
   liftIO $ ANSI.setSGR [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Green]
   UTF8.putText $ "✅ " <> msg <> "\n"
   liftIO $ ANSI.setSGR [ANSI.Reset]
+
+dnsUpdated :: MonadIO m => Text -> m ()
+dnsUpdated domain = do
+  UTF8.putText "📝 DNS Updated. Check out your site at: \n"
+  UTF8.putText $ "🔗 " <> domain  <> "\n"
```
expede: 😍  Beautiful ✨

library/Fission/Web/DNS/Client.hs
```diff
@@ -0,0 +1,15 @@
...
+
+import Servant
+import Servant.Client
+
+import           Fission.IPFS.CID.Types
+import qualified Fission.Web.Routes as Routes
+import qualified Fission.AWS.Types as AWS
+
+update :: BasicAuthData -> CID -> ClientM AWS.DomainName
+update = client (Proxy :: Proxy Routes.DNSRoute)
```
expede: EOF newline (🚫 ↩️)

expede:
Superseded by #156

PR #148, opened 2019-10-15
#139 Fission down

bechurch:
A similar PR may already be submitted!
Please search among the [Pull request](../) before creating one.

Thanks for submitting a pull request! Please provide enough information so that others can review your pull request:

For more information, see the `CONTRIBUTING` guide.

**Summary**

This PR fixes/implements the following **bugs/features**

* [x] Adds a `down` command to download objects from ipfs
* [x] Provide a waiting message during down

Explain the **motivation** for making this change. What existing problem does the pull request solve?

Exposes more ipfs functionality through fission for a better user experience.

**Closing issues**

closes #139

PR #150, opened 2019-10-16
Fission Down ⬇️

expede:
Making a few changes to #148 after the changes in #149

### DIFF
--- a/library/Fission/CLI.hs
+++ b/library/Fission/CLI.hs
@@ -26,10 +26,10 @@ cli :: MonadRIO    cfg m
     => Has Client.Runner cfg
     => Has IPFS.BinPath  cfg
     => Has IPFS.Timeout  cfg
-    => cfg
-    -> IO ((), m ())
-cli cfg =
-  simpleOptions version description detail (pure ()) do
+    => m ((), m ())
+cli = do
+  cfg <- ask
+  liftIO $ simpleOptions version description detail (pure ()) do
     Login.command    cfg
     Register.command cfg
     Up.command       cfg
--- a/library/Fission/CLI.hs
+++ b/library/Fission/CLI.hs
@@ -7,6 +7,8 @@ import           Data.Has
 
 import           Options.Applicative.Simple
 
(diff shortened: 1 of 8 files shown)


expede:
Okay, so looks like Ben got around to integrating those changes into his PR. Will review this and see what's worth porting over.

library/Fission/CLI.hs
```diff
@@ -7,33 +7,37 @@ import           Data.Has
...
     => HasLogFunc        cfg
     => HasProcessContext cfg
     => Has Client.Runner cfg
     => Has IPFS.BinPath  cfg
     => Has IPFS.Timeout  cfg
-    => cfg
-    -> IO ((), m ())
-cli cfg =
-  simpleOptions version description detail (pure ()) do
+    => m ()
```
expede: Let's just simplify this type to something more like what you'd expect

library/Fission/CLI/Command/Down.hs
```diff
@@ -31,23 +32,21 @@ command cfg =
     "down"
     "pull a ipfs or ipns object down to your system"
     (\cid -> runRIO cfg $ down cid)
-    (strArgument $ metavar "ContentID" <> help "The CID of the IPFS object you want to download")
+    (strArgument $ mconcat
+      [ metavar "ContentID"
+      , help "The CID of the IPFS object you want to download"
+      ])
```
expede: Breaking these onto multiple lines to make it feel like a list of options

library/Fission/CLI/Command/Login.hs
```diff
@@ -26,7 +26,7 @@ import qualified Fission.CLI.Display.Error   as CLI.Error
 import qualified Fission.CLI.Display.Wait    as CLI.Wait
 
 -- | The command to attach to the CLI tree
-command :: MonadUnliftIO m
+command :: MonadIO m
```
expede: Doesn't need to be `MonadUnliftIO`. Compiler was complaining.

PR #169, opened 2019-10-19
add whoami cli command

dholms:
# Problem
Users have to check `~/.fission.yaml` to determine who they're currently logged in as

# Solution
Add `whoami` command that prints user's username

Closes https://github.com/fission-suite/web-api/issues/161

### DIFF
--- a/library/Fission/CLI/Command/Whoami.hs
+++ b/library/Fission/CLI/Command/Whoami.hs
@@ -1,0 +1,48 @@
+-- | Whoami command
+module Fission.CLI.Command.Whoami (command, whoami) where
+
+import           RIO
+import           RIO.ByteString
+
+import           Options.Applicative.Simple (addCommand)
+import           Servant
+
+import qualified Data.ByteString.Char8 as BS
+import qualified System.Console.ANSI as ANSI
+
+import           Fission.Internal.Constraint
+import qualified Fission.Internal.UTF8 as UTF8
+
+import qualified Fission.CLI.Auth as Auth
+import           Fission.CLI.Config.Types
+
+
+-- | The command to attach to the CLI tree
+command :: MonadIO m
+        => HasLogFunc cfg
+        => cfg
+        -> CommandM (m ())
+command cfg =
+  addCommand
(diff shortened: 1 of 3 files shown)


library/Fission/CLI/Command/Whoami.hs
```diff
@@ -0,0 +1,48 @@
...
+      Auth.couldNotRead
+
+    Right auth -> do
+      let username = basicAuthUsername auth
+      UTF8.putText "💻 Currently logged in as: "
+
+      liftIO $ ANSI.setSGR [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Blue]
+      putStr $ BS.append username "\n"
+
+      liftIO $ ANSI.setSGR [ANSI.Reset]
```
expede: Why not:

```haskell
    Right auth -> do
      let username = 
      UTF8.putText "💻 Currently logged in as: "

      liftIO $ ANSI.setSGR [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Blue]
      ----------- THIS LINE -----------
      putStr $ basicAuthUsername auth <> "\n"

```

library/Fission/CLI/Command/Whoami.hs
```diff
@@ -0,0 +1,48 @@
...
+
+import qualified Fission.CLI.Auth as Auth
+import           Fission.CLI.Config.Types
+
+
+-- | The command to attach to the CLI tree
+command :: MonadIO m
+        => HasLogFunc cfg
+        => cfg
+        -> CommandM (m ())
```
expede: Not for this PR, but FYI I'm going to write a helper:

```haskell
type MonadCmd cmd = forall m . MonadIO m => CommandM (m cmd)
```

so then we can write:

```haskell
command :: HasLogFunc cfg => cfg -> MonadCmd ()
```

Or something with better naming

library/Fission/CLI/Command/Whoami.hs
expede: Also EOF newline

PR #170, opened 2019-10-20
Fission CLI logout

bechurch:
**Summary**
Adds a `logout` command to the CLI that removes the `.fission.yaml` file

### DIFF
--- a/library/Fission/CLI/Auth.hs
+++ b/library/Fission/CLI/Auth.hs
@@ -4,6 +4,7 @@ module Fission.CLI.Auth
   , withAuth
   , write
   , couldNotRead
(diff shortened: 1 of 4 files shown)


library/Fission/CLI/Auth.hs
```diff
@@ -46,6 +47,12 @@ couldNotRead = do
   UTF8.putText "fission-cli login"
 
   liftIO $ ANSI.setSGR [ANSI.Reset]
+-- | Removes the users config file
+-- removeConfigFile :: MonadIO m => m FilePath
```
expede: 🔪☠️

PR #171, opened 2019-10-20
New syntax: Proxy @value

expede:
> Now that I have some time, I'm going through and cleaning up some syntax, writing helpers, &c

# Problem

We have a lot of `Proxy :: Proxy Foo` boilerplate

# Solution

Swap that out to `Proxy @Foo`


expede:
Rejoice, for pipes are likely coming shortly 😉

expede:
I did this back in Berlin when I had a spare 10 minutes. Just rebased it. Really simple change in syntax: `Proxy :: Proxy API` can now be written as `Proxy @API`, which I feel like reads cleaned / less boilerplate. Thoughts @dholms / @bechurch ?

PR #174, opened 2019-10-25
Pairing on 116

expede:
WIP

### DIFF
--- a/library/Fission/CLI/Auth.hs
+++ b/library/Fission/CLI/Auth.hs
@@ -57,11 +57,11 @@ removeConfigFile = do
 
 withAuth :: MonadRIO   cfg m
          => HasLogFunc cfg
-         => (UserConfig -> m (Either ClientError a))
+         => (BasicAuthData -> m (Either ClientError a))
          -> m (Either SomeException a)
 withAuth action = get >>= \case
   Right auth ->
-    action auth >>= pure . \case
+    action (toBasicAuth auth) >>= pure . \case
       Right result -> Right result
       Left err -> Left $ toException err
 
--- a/library/Fission/CLI/Command/Login.hs
+++ b/library/Fission/CLI/Command/Login.hs
@@ -62,11 +62,17 @@ login = do
       let auth = BasicAuthData username $ BS.pack password
 
       authResult <- Cursor.withHidden
-                 . CLI.Wait.waitFor "Verifying your credentials"
-                 . runner
-                 $ IPFS.Client.peers (IPFS.Client.request auth)
+                  . CLI.Wait.waitFor "Verifying your credentials"
+                  . runner
+                  $ IPFS.Client.peers (IPFS.Client.request auth)
 
       case authResult of
+        Left err ->
+          CLI.Error.put err "Authorization failed"
+
+        Right [] ->
+          CLI.Error.put_ "Unable to find peers"
+
         Right peers -> do
           let writeTo = UserConfig { username = username
                                    , password = BS.pack password
@@ -75,6 +81,3 @@ login = do
 
           Auth.write writeTo
           CLI.Success.putOk "Logged in"
-
-        Left err ->
-          CLI.Error.put err "Authorization failed"
--- a/library/Fission/CLI/Command/Login.hs
+++ b/library/Fission/CLI/Command/Login.hs
@@ -6,7 +6,7 @@ import           RIO.ByteString
 
 import qualified Data.ByteString.Char8 as BS
 import           Data.Has
-import           Data.List.NonEmpty
+import           Data.List.NonEmpty as NonEmpty
 
 import           Options.Applicative.Simple (addCommand)
 import           Servant
@@ -15,7 +15,6 @@ import           System.Console.Haskeline
 import qualified Fission.Config as Config
 import           Fission.Internal.Constraint
 
-import           Fission.Web.User.Client  as User.Client
 import           Fission.Web.IPFS.Client  as IPFS.Client
 import qualified Fission.Web.Client.Types as Client
 
@@ -26,8 +25,6 @@ import qualified Fission.CLI.Display.Cursor  as Cursor
 import qualified Fission.CLI.Display.Success as CLI.Success
 import qualified Fission.CLI.Display.Error   as CLI.Error
 import qualified Fission.CLI.Display.Wait    as CLI.Wait
-import qualified Fission.IPFS.Peer.Types as Peer
-import qualified RIO.ByteString.Lazy as Lazy
 
 -- | The command to attach to the CLI tree
 command :: MonadIO m
@@ -71,9 +68,13 @@ login = do
 
       case authResult of
         Right peers -> do
-          let writeTo = UserConfig {username = username
-                                    , password = (BS.pack password)
-                                    , peers = fromList peers}
+          let writeTo = UserConfig { username = username
+                                   , password = BS.pack password
+                                   , peers    = NonEmpty.fromList peers
+                                   }
 
-          Auth.write writeTo >> CLI.Success.putOk "Logged in"
-        Left  err -> CLI.Error.put err "Authorization failed"
+          Auth.write writeTo
+          CLI.Success.putOk "Logged in"
+
+        Left err ->
+          CLI.Error.put err "Authorization failed"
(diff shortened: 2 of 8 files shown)


library/Fission/CLI/Pin.hs
```diff
@@ -45,65 +44,29 @@ run :: MonadRIO          cfg m
...
-  -- CLI.Success.live hash
-  -- return cid
-
-  -- case foo' of
-  --   Left err -> return $ Left err
-  --   Right pr -> do
-  --     IPFS.Peer.connect pr
-
-  --     bar <- pin runner userConfig cid
+  liftE $ IPFS.Peer.connect IPFS.Peer.fission
```
expede: It's kind of hard to see in the diff, so here it is in plain:

```haskell
run :: MonadRIO          cfg m
    => MonadUnliftIO         m
    => HasLogFunc        cfg
    => HasProcessContext cfg
    => Has Client.Runner cfg
    => Has IPFS.BinPath  cfg
    => Has IPFS.Timeout  cfg
    => CID
    -> NonEmpty IPFS.Peer
    -> BasicAuthData
    -> m (Either SomeException CID)
run cid@(CID hash) peers auth = runExceptT do
  logDebug $ "Remote pinning " <> display hash
  Client.Runner runner <- Config.get

  liftE $ IPFS.Peer.connect IPFS.Peer.fission

  IPFS.Peer.connect $ head $ peers
  liftE $ pin runner cid auth

  CLI.Success.live hash
  return cid
```

library/Fission/CLI/Command/Login.hs
```diff
@@ -70,9 +68,13 @@ login = do
 
       case authResult of
         Right peers -> do
-          let writeTo = UserConfig {username = username
-                                    , password = (BS.pack password)
-                                    , peers = peers}
+          let writeTo = UserConfig { username = username
+                                   , password = BS.pack password
+                                   , peers    = NonEmpty.fromList peers
```
expede: `NonEmpty` is a data type like `[]`, but always has a value in it (i.e. `head` is safe to use without a `Maybe`). My putting this upstream, we can always assume that the value is there

expede: Really we should be using `nonEmpty` here, since it's total 🤔

library/Fission/CLI/Command/Register.hs
```diff
@@ -101,16 +103,12 @@ register' = do
...
-              username = username,
-              password = password,
-              peers = peers
-            }
+            Provision { _peers, _username, _password = Secret pass } = user
+            auth = UserConfig
+              { username = encodeUtf8 _username
+              , password = encodeUtf8  pass
+              , peers    = fromList _peers
+              }
```
expede: > -- Question: Theres got to be a simpler way to convert

This is one simpler phrasing 👆

library/Fission/CLI/Command/Up.hs
```diff
@@ -54,7 +55,8 @@ up Up.Options {..} = handleWith_ Error.put' do
   logDebug $ "Starting single IPFS add locally of " <> displayShow absPath
 
   unless dnsOnly do
-    void . liftE . Auth.withAuth $ CLI.Pin.run cid
+    config <- liftE Auth.get
+    void . liftE $ CLI.Pin.run cid (peers config) (toBasicAuth config)
```
expede: We do these exact two lines in a couple of places. May be worth DRYing up.

library/Fission/CLI/Auth.hs
```diff
@@ -57,11 +57,11 @@ removeConfigFile = do
 
 withAuth :: MonadRIO   cfg m
          => HasLogFunc cfg
-         => (UserConfig -> m (Either ClientError a))
+         => (BasicAuthData -> m (Either ClientError a))
          -> m (Either SomeException a)
 withAuth action = get >>= \case
   Right auth ->
-    action auth >>= pure . \case
+    action (toBasicAuth auth) >>= pure . \case
```
expede: This was a WIP branch, and stuff was being messed around intentionally. Totally get that 👍 Changing this back made life much easier.

This function no longer applied only to authentication. Here's the type signature on `master`:

```haskell
withAuth :: MonadRIO   cfg m
         => HasLogFunc cfg
         => (BasicAuthData -> m (Either ClientError a))
         -> m (Either SomeException a)
```

Changing `BasicAuthData` to `UserConfig` made life much harder, and a lot of functions read in a weird way. For example, `Auth.get` now tells you about `peers`.

We have a few options here. One is to move `Auth.get` to `UserConfig.get`. Another (preferred) way is to parse `BasicAuthData` and `NonEmpty Peers` from the same file, but do it in two passes to get two data types. We can either run the two effects, or read the file in and parse the pure data.

library/Fission/CLI/Command/Login.hs
```diff
@@ -64,15 +62,22 @@ login = do
...
+                  . CLI.Wait.waitFor "Verifying your credentials"
+                  . runner
+                  $ IPFS.Client.peers (IPFS.Client.request auth)
 
       case authResult of
+        Left err ->
+          CLI.Error.put err "Authorization failed"
+
+        Right [] ->
+          CLI.Error.put_ "Unable to find peers"
```
expede: Fail if we have no peers

expede: ...actually maybe we want to reconnect IPFS in this scenario?

library/Fission/CLI/Pin.hs
expede: Breaking down a few of the changes:

# 1. `NonEmpty`

We shouldn't get this far into the function if we have no peers. In general the pattern in Haskell is to push any ambiguities as close to the application boundary as possible. IMO this is a good pattern to apply to any language. Here we do this here: https://github.com/fission-suite/web-api/pull/174/files#diff-07ef7e40bb53d7443ccd27f6e7dbe01dR69-R80 , and make it a non-empty list (`NonEmpty a`)

# 2. No More `Maybe`

Thanks to that `NonEmpty`, we no longer need to pattern match on `maybeHead`

# 3. `ExceptT`

Setup: Change the monad that we're running in to do automated error handling by sticking a `runExcepT` at the beginning. Now `>>=` uses the instance from `ExceptT` because it's the outermost one.

Payoff: Take anything in the form `m (Either err val)` and slap a `liftE` on the front. This does the railroad programming / error handling for us.

library/Fission/CLI/Command/Watch.hs
```diff
@@ -112,21 +116,20 @@ handleTreeChanges timeCache hashCache watchMgr cfg dir =
...
-      return $ Left err
-
-    Right _ ->
-      Auth.withAuth $ CLI.DNS.update cid
+pinAndUpdateDNS cid = runExceptT do
+  config <- liftE Auth.get
+  let auth = toBasicAuth config
+
+  _ <- liftE $ CLI.Pin.run cid (peers config) auth
+  liftE $ CLI.DNS.update cid auth
```
expede: This seems gratuitous, but maybe that's me?

library/Fission/CLI/Display/Error.hs
```diff
@@ -1,24 +1,33 @@
...
   [ "Something went wrong. Please try again or file a bug report with "
   , "Fission support at https://github.com/fission-suite/web-api/issues/new"
   ]
+
+-- | Display a given error to the user
+put_ :: MonadIO m => Text -> m ()
+put_ msg = do
+  liftIO $ ANSI.setSGR [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Red]
+  UTF8.putText $ "🚫 " <> msg <> "\n"
+  liftIO $ ANSI.setSGR [ANSI.Reset]
```
expede: We've wanted this for a while

library/Fission/CLI/Pin.hs
```diff
@@ -25,17 +26,15 @@ import qualified Fission.Web.Client      as Client
...
 import Control.Monad.Except
 
-maybeToEither:: Maybe val -> err -> Either err val
-maybeToEither maybeA err = case maybeA of
-  Just val -> Right val
-  Nothing  -> Left err
+-- maybeToEither:: Maybe val -> err -> Either err val
+-- maybeToEither maybeA err = case maybeA of
+--   Just val -> Right val
+--   Nothing  -> Left err
```
expede: Don't need this any more

library/Fission/CLI/Command/Up.hs
expede: Actually, we're doing this because it's no longer just auth in that data structure. I recommend splitting them, as per another comment, and then we can go back to `withAuth`.

PR #176, opened 2019-10-28
Remove CLI

expede:
The CLI has been moved to its own repo: https://github.com/fission-suite/cli


README.md
```diff
@@ -145,7 +145,7 @@ Live monitoring dashboard
 
 ```
 export MONITOR=true
-stack run fission-web
+stack run
```
expede: `stack` no longer has to disambiguate, so `stack run` is sufficient. We have the `Makefile`, too, so this won't matter to the internal team too much.

package.yaml
```diff
@@ -156,25 +153,6 @@ executables:
...
-      - -flate-specialise
-      - -rtsopts
-      - -with-rtsopts=-N
-      - -with-rtsopts=-T
-      - -optc-ffast-math
-      - -optc-O3
-      - -O2
-    # uncomment to build a static binary on Linux
-    # cc-options: -static
-    # ld-options: -static -pthread
```
expede: We also don't care about static binaries for the web API 👍

PR #177, opened 2019-10-28
Dont return data from register endpoint

bechurch:
**Summary**
We shouldn't return sensitive data from the register endpoint. Or any data at all 😅 

This PR:
1. Changes `/register` to return `()` instead of `User.Provision`
1. Removes the Register Client file. (This has been readded to cli in an upcoming pr)


library/Fission/Web/User/Create.hs
```diff
@@ -57,11 +52,7 @@ server (User.Registration username password email) = do
   ensureContent $ registerDomain Route53.Cname baseUrl "ipfs.runfission.com"
   ensureContent $ registerDomain Route53.Txt dnsLinkUrl $ dnsLinkTxt `wrapIn` "\""
 
-  return User.Provision
-    { _url      = url
-    , _username = username
-    , _password = Secret password
-    }
+  return ()
```
expede: 👍

expede:
That would be me being slopping in splitting out the CLI. I _knew_ it seemed too easy 😆

PR #181, opened 2019-10-30
Fix #140: Heroku Deprovision 406

### LINKED ISSUE #140: Heroku Deprovision 406
# Summary

## Problem

Deprovision requests to Heroku _fully succeed in the UI_, but the server is returning a 406.

### Impact

Probably noting other than some ugly logs, but Heroku has asked us to look into it. We should make sure that those keys actually do get invalidated, and items unpinned. It's probably the most involved web handler in the entire project 🤷‍♀️

## Solution

* Check that it's actually working
* Write some tests

https://devcenter.heroku.com/articles/building-an-add-on#the-deprovisioning-request

expede:
# Problem

Tracking issue: #140 

The Heroku add-on API has been recording 406 ("Not Acceptable") errors on de-provisioning requests. It looks like the 406 is actually on their end. We're missing two details from their docs (see next section).

# Solution

## Content Type
Return a `Content-Type: application/vnd.heroku-addons+json; version=3`

## HTTP Status Code

Return a 410 on subsequent requests after 24 hours... or something? I dunno it's not totally clear. Using a 410 as failure when a record isn't found appears to be an acceptable response 🤷‍♀️ We _could_ update the add-on to track if it has been used in the past, and disambiguate between the 410 and 404, but that seems like more work than it's worth to make an external API happy.

### DIFF
--- a/library/Fission/Web/Error.hs
+++ b/library/Fission/Web/Error.hs
@@ -64,6 +64,7 @@ ensure :: MonadRIO   cfg m
        => HasLogFunc cfg
        => MonadThrow     m
        => Display       err
+       => Exception     err
        => ToServerError err
        => Either err a
        -> m a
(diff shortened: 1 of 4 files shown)


expede:
Update some style

PR #189, opened 2019-11-05
Longer HTTP Timeout

expede:
Increase timeout for both clients and requesting from IPFS node to 30 minutes.

We were experiencing issues where large pins were timing out, and propagating through to the CLI.

This is already running on prod as a quick fix. Merging to master now for consistency.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -67,18 +67,27 @@ main = do
 
     let
       Web.Port port' = _port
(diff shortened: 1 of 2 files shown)


expede:
@dholms 

Yeah, they totally should be env vars! Let's merge this for (quick fix), and I'll open a new PR later today that does the env var

PR #192, opened 2019-11-08
DNS Swagger Docs

dholms:
## Problem
Missing swagger documentation on dns route. It just says `default`

## Solution
Add high-level swagger documentation to route

### DIFF
--- a/library/Fission/Web/Swagger.hs
+++ b/library/Fission/Web/Swagger.hs
@@ -26,7 +26,8 @@ server :: Host -> RIOServer cfg API
 server = hoistServer (Proxy @API) fromHandler . swaggerSchemaUIServer . docs
 
 docs :: Host -> Swagger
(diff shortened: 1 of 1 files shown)


library/Fission/Web/Swagger.hs
```diff
@@ -68,6 +69,10 @@ ping :: Swagger -> Swagger
 ping = makeDocs (Proxy @Web.PingRoute)
   ["Ping" & description ?~ "Check for liveness"]
 
+dns :: Swagger -> Swagger
+dns = makeDocs (Proxy @Web.DNSRoute)
+  ["DNS" & description ?~ "Interact with DNS on AWS Route53"]
```
expede: 🤔 We probably don't need to be this specific. Let's say we later switch provider. "DNS management" or something may be sufficient.

PR #193, opened 2019-11-08
Sentry Integration

### LINKED ISSUE #179: Logging with Sentry
# Summary

## Problem

Logs with `nohup.out` are... insufficient. They're very easy to overwrite, we have to SSH in the server to read them, searching logs with `grep` leaves a lot to be desired.

## Solution

Integrate Sentry

# Detail

* Cookbook Guide: https://docs.servant.dev/en/stable/cookbook/sentry/Sentry.html
* Library: http://hackage.haskell.org/package/raven-haskell

expede:
# Problem

Inspecting errors by hand is a terrible experience:

* Don't get notified
* Digging through logs can be difficult

# Solution

* [x] Set up Sentry
* [x] Add Sentry key to prod
* [x] Create Sentry team account
* [x] Invite team

# Detail

* Closes #179
* We're not doing this as middleware because we'd like to capture more info than gets currently exposed through WAI

Screenshot from testing:

Screen Shot 2019-11-09 at 18 35 10


app/Main.hs
```diff
@@ -59,23 +61,29 @@ main = do
...
-      runner         = if env ^. web . Web.isTLS then runTLS tlsSettings' else runSettings
-      condDebug      = if env ^. web . Web.pretty then id else logStdoutDev
-
-    when (env ^. web . Web.monitor) Monitor.wai
-    liftIO . runner settings
-           . CORS.middleware
-           . condDebug
-           =<< Web.app
+
+      _logFunc = baseLogger <> condSentryLogger
```
expede: Did you know that loggers are composable? I didn't! Pretty neat 👍

PR #194, opened 2019-11-08
Added POST /user/reset_password route

dholms:
## Problem
No way for users to reset their password

## Impact 
If a user exposes their password/api keys, then they can't cycle their credentials

## Solution
Add a POST /user/reset_password route
- optional password param
- if not included then it defaults to a random password

Closes https://github.com/fission-suite/web-api/issues/191

### DIFF
--- a/library/Fission/User/Mutate.hs
+++ b/library/Fission/User/Mutate.hs
@@ -1,6 +1,7 @@
 module Fission.User.Mutate
   ( create
   , createWithHeroku
+  , updatePassword
   ) where
 
 import           RIO
@@ -16,14 +17,16 @@ import           Fission.Internal.Constraint
 import           Fission.Internal.Orphanage.ID ()
 
 import           Fission.Timestamp as Timestamp
+import           Fission.Random    as Random
 
 import qualified Fission.Platform.Heroku.AddOn as Heroku
 import qualified Fission.Platform.Heroku.Types as Heroku
 
 import           Fission.User.Role
 import           Fission.User.Types
-import qualified Fission.User.Table as Table
-import qualified Fission.User.Mutate.Error as Error
+import qualified Fission.User.Table          as Table
+import qualified Fission.User.Mutate.Error   as Error
+import qualified Fission.User.Password.Types as User
 
 -- | Create a new, timestamped entry
 create :: MonadRIO    cfg m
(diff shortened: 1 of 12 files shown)


library/Fission/Web/User/Password/Reset.hs
```diff
@@ -0,0 +1,31 @@
...
+import           Servant
+
+import           Fission.Web.Server
+import qualified Fission.Web.Error    as Web.Err
+
+import           Fission.User                 as User
+import qualified Fission.User.Password.Types  as User
+
+type API = ReqBody '[JSON] User.Password
+        :> Post '[JSON] User.Password
```
expede: Per REST, this should be a `PUT`. I wish that they had named it better; `PUT` vs `POST` is confusing.

library/Fission/User/Mutate.hs
```diff
@@ -71,6 +74,29 @@ create' username password email herokuUUID = do
       logInfo $ "Inserted user " <> display uID
       return $ Right uID
 
+updatePassword :: MonadRIO    cfg m
+               => MonadSelda      m
+               => HasLogFunc  cfg
+               => ID User
+               -> User.Password
+               -> m (Either Error.Create User.Password)
```
expede: 💅 Question for ya: how do you feel about this style:

```haskell
updatePassword 
  :: ( MonadRIO    cfg m
     , MonadSelda      m
     , HasLogFunc  cfg
     )
  => ID User
  -> User.Password
  -> m (Either Error.Create User.Password)
```

This is both a more common style, and after a few months and two onboardings, I think possibly clearer than the row of `=>`. Very clear where the constraints are versus arguments.

Dropping the `::` to the next line is newer, but starting to catch on in the community, too. It makes indentation more consistent.

I was thinking about us slowly changing to this style as we go. Any thoughts?

library/Fission/User/Mutate.hs
```diff
@@ -71,6 +74,29 @@ create' username password email herokuUUID = do
...
+    Right secretDigest -> do
+      update Table.users 
+        (\user -> user ! #_userID .== literal userID)
+        (\user -> user `with` [#_secretDigest := literal secretDigest])
+      logInfo $ "Updated password for user " <> display userID
+      return . Right . User.Password $ Just password
+
+defaultPassword :: MonadRIO cfg m
+                => Maybe Text
+                -> m (Text)
```
expede: 💅

* No need for the brackets around `(Text)`
* No need for the brackets around `(Just pw)` in this case
* This doesn't need to be full `MonadRIO`; `MonadIO` will do
  * `MonadRIO` is for when you need to pass around some config

```haskell
defaultPassword :: MonadIO m => Maybe Text -> m Text
defaultPassword Just pw = return pw
defaultPassword Nothing = liftIO $ Random.text 200
```

library/Fission/Web/User/Password/Reset.hs
dholms: Yup you're right. I often mix those up

library/Fission/User/Mutate.hs
dholms: Already left my thoughts in the other PR :wink: but yes big fan

library/Fission/User/Mutate.hs
dholms: Good point on 1 & 3. Compiler freaks out without brackets around `Just pw` though: `Equations for ‘defaultPassword’ have different numbers of arguments`

PR #195, opened 2019-11-10
Fission.Prelude |> and pipes |> no underscores |> Restyle

expede:
I was stuck in a couple of airports today, delayed flights, unable to concentrate. So I did those style updates we've been chatting about.

It's both a lot of LOC, and yet fewer than I expected.

# `Fission.Prelude`

We've been using `RIO`, but are also importing the same modules over and over. It's not uncommon to set up a project common re-export module, called a `Prelude`. This branch has one at:

```haskell
import Fission.Prelude
```

It imports `RIO`, `Data.Has`, and a few more things. No need to import those when you use `Fission.Prelude`.

# Pipes

`$` and `&` are now hidden, replaced with `<|` and `|>` respectively. There's also `<.` and `.>`, but have also left `.` in there, too. `.` reads very well IMO, in application order.

The flow works more like Elm's than Elixir's. The good news is that Elm is widely accepted as having nice syntax, but may feel less familiar to Ben. Elixir's pipes thread a "subject" in the "first" position. In Elm/Haskell, the "subject" tends to be the last argument, because of auto-currying, which means that `<|` actually sees a lot of use in Elm (and I'd assume for us, too).

> Brooke's PLT-Nerd Side Commentary
> ---------------------------------
> Actually Elixir is the odd one out here, across languages. Pipes are super nice, but the first-argument threading is a consequence of Elixir having variadic functions.

## Constructor Order

Data constructors are functions. Listing them in the same sequence as you destructure them seems to make the most sense.

```haskell
return <| Left <| UnknownErr <| UTF8.textShow err
```

# No More `makeLenses`

We weren't really using our own lenses.

## `instance Has foo Config`

This branch switches to the `lens` construction. It's pretty straightforward once you know the goal. `lens` is a higher order function that composes a getter and setter in a special way. It get used like this:

```haskell
instance Has Host Config where
  hasLens = lens host \cfg newHost -> cfg { host = newHost }
```

Think of `lens` like a barely fancier version of:

```haskell
data Lens record fieldValue = Lens 
  { getter :: record -> fieldValue
  , setter :: record -> fieldValue -> record
  }
```

It's a small amount of boilerplate. Just like how we moved to manually writing `ToJSON` and `FromJSON` instances.

## Accessors

Instead of `foo ^. bar . baz`, to do nested _lookups_ we can use the same pipe oprator as above. For example:

```haskell
-- app/Main.hs

  let
    Storage.Environment {..} = env |> storage
    Web.Environment     {..} = env |> web
    AWS.Environment     {..} = env |> aws

    -- ...

    ipfsPath    = env |> ipfs |> binPath
    ipfsURL     = env |> ipfs |> url
    ipfsTimeout = env |> ipfs |> IPFS.timeout
```

## Record Fields Underscores

Since we're not deriving lenses anymore, we don't need to prefix record fields with underscores anymore. Reads slightly nicer.

# Constraint Style

There are two valid styles. There's a movement to dropping the `::` to the next line, and the other style is actually quite readable in this style:

```haskell
pin
  :: ( Has HTTP.Manager  cfg
     , Has IPFS.URL      cfg
     , MonadSelda   (RIO cfg)
     , HasLogFunc        cfg
     )
  => ID User
  -> RIOServer cfg PinAPI
pin userID = -- ...
```

i.e.

```haskell
functionName
  :: ( constraintA
     , constraintB
     )
  => arg1
  -> arg2
  -> output
```

# Selda Selectors

We enabled `OverloadedLabels` syntax a while back. This means that we no longer need special SQL selectors. Those modules are gone, and we just use a `#` before the field name to get a selector. For example:

```haskell
byUsername username user = user `is'` #active
                       .&& user ! #username .== text username
```


library/Fission/AWS/Route53.hs
```diff
@@ -3,6 +3,7 @@ module Fission.AWS.Route53
   , registerDomain
   ) where
 
+import Flow
```
dholms: Any reason to not use `Fission.Prelude` here? or is that just an oversight?

expede:
> Any reason for that or have they just not been switched over yet?

Purely just me missing them 😅 Did this between (delayed) flights on zero sleep, so totally possible that there are modules have yet to be converted.

expede:
I walked @dholms through this on a call yesterday. @bechurch, lemme know if you'd like a walkthrough as well, though it's pretty much all in the PR description.

library/Fission/IPFS/Client.hs
```diff
@@ -7,6 +7,7 @@ module Fission.IPFS.Client
   , unpin
   ) where
 
+import           Flow
```
bechurch: Should be a Prelude import now right?

library/Fission/Timestamp.hs
```diff
@@ -6,6 +6,7 @@ module Fission.Timestamp
   , (<@)
   ) where
 
+import Flow
```
bechurch: should be Import Prelude?

library/Fission/IPFS/Client.hs
expede: It should be, yes 👍

expede:
Okay, I actually wen through _all the files_, and believe that's it for (illegitimate) `import RIO`s. Make the diff quiet a bit bigger, which... makes sense 😛

PR #196, opened 2019-11-11
Separate executable-specific code into `app`

expede:
I've been thinking a bit about how we'd get that PostgreSQL dependency out of the CLI. There's a few options, but really the one that makes the most sense is to take things that are specific to our implementation, and move them to `app`. It would be nice to keep the rest of the library (especially Servant types) available, and this approach achieves that.

Selda can have any number of DBs plugged into it. In the approach in this PR, we leave it abstract, and plug in `PG` under `app`. We then move the `selda-postgresql` dependency to the app.

It also moves the `*.Environment.Types` files to `app`. These are very specific to this exact app, and not, say, an internal development environment.

---

The other thing that I'd like to do at some point is break out a bunch of interfaces into typeclasses. Right now we need to thread through a BUNCH of `Has foo cfg`. If you look at anything with `MonadSelda (RIO cfg)` on it, you won't see `Has DB.Pool`. This is because our `instance MonadSelda (RIO cfg)` assumes that you `Has DB.Pool cfg`. It's a _very_ common pattern to break out all of your interfaces into little DSLs (e.g. `MonadIPFS` would give you `cat` and `pin`). The the constraints would tell us more about what the function can _do_ rather than what it _needs_. Doing this seems like a natural extension to this PR, since all of the `instance`s that would run the library code would like in `app`.


app/Fission/Config/Types.hs
```diff
@@ -12,19 +12,19 @@ import qualified Fission.Storage.Types         as DB
 import qualified Fission.Platform.Heroku.Types as Heroku
 import qualified Network.AWS.Auth              as AWS
 import qualified Fission.AWS.Types              as AWS
-import           Fission.Internal.Orphanage.PGConnectInfo ()
+-- import           Fission.Internal.Orphanage.PGConnectInfo ()
 
 -- | The top level 'Fission' application 'RIO' configuration
-data Config = Config
+data Config dbBackend = Config
```
expede: Parameterized this type by the kind of database backend we plug in over in `app`.

package.yaml
```diff
@@ -144,6 +143,7 @@ executables:
     source-dirs: app
     dependencies:
       - fission-web-api
+      - selda-postgresql
```
expede: No longer a dependency for the app. This frees anything downstream (read: the Fission CLI) from the `lib-pg` dependency.

library/Fission/Types.hs
```diff
@@ -1,9 +1,9 @@
...
+module Fission.Types where -- (Fission) where
 
 import RIO
 
-import Fission.Config.Types
+-- import Fission.Config.Types
 
 -- | Top-level application type
-type Fission = RIO Config
+-- type Fission be = RIO (Config be)
```
expede: We're not actually using this file at the moment

bechurch:
Leaving a note here that we'll want to update out CLI build to omit the postgres install!

expede:
Oh wow. Rather than attempt the _super painful_ rebase that this looks like it'll be, I'll just re-create the branch 😅 It's not that much work once you know what needs to be done

expede:
Closing in favour of #203

PR #199, opened 2019-11-12
More idiomatic password reset without Maybe

expede:
I need to run to a meeting like... now!

But @dholms this is a slightly more idiomatic version. It's more code, but also a cleaner, more reusable, &c. I can walk you through it later in our one-on-one 👍

---

> ...Later...

Welp, I merged the last PR, and then realized that the `Maybe` was at the wrong layer 😅

### DIFF
--- a/library/Fission/User/Password/Types.hs
+++ b/library/Fission/User/Password/Types.hs
@@ -5,7 +5,7 @@ import RIO
 import Data.Aeson
 import Data.Swagger
 
-newtype Password = Password { password :: Maybe Text }
+newtype Password = Password { password :: Text }
   deriving          ( Eq
                     , Generic
                     , Show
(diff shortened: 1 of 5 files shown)


library/Fission/User/Password/Types.hs
```diff
@@ -5,19 +5,15 @@ import RIO
 import Data.Aeson
 import Data.Swagger
 
-newtype Password = Password { password :: Maybe Text }
+newtype Password = Password { password :: Text }
```
expede: The data model itself is not a `Maybe`

library/Fission/User/Password/Types.hs
```diff
@@ -5,19 +5,15 @@ import RIO
...
   deriving          ( Eq
                     , Generic
                     , Show
                     )
   deriving anyclass ( ToSchema )
 
 instance ToJSON Password where
-  toJSON (Password password) = 
-    Object [("password", maybe Null String password)]
+  toJSON (Password password) = toJSON $ String password
```
expede: Also not an object

library/Fission/Web/User/Password/Reset/Types.hs
```diff
@@ -0,0 +1,25 @@
...
+import RIO
+
+import Control.Lens
+
+import Data.Aeson
+import Data.Swagger
+
+import qualified Fission.User.Password.Types as User
+
+newtype Reset = Reset { maybePassword :: Maybe User.Password }
```
expede: Here's the web payload

PR #200, opened 2019-11-12
add toJSON instance for User.Password.Reset

dholms:
## Problem
Can't use Servant client without ToJSON instance for `User.Password.Reset`

## Solution
Add ToJSON instance


expede:
Please hold for #195

PR #202, opened 2019-11-13
Basic text functions in `Fission.Prelude`

expede:
When reviewing a CLI PR, I realized that it may be helpful to have a printing function with automatic newlines. The base `Prelude` has `putStrLn`, which `RIO` eschews for some reason. In any case, here's some helpful functions by default to make it easier to default to printing each statement on a new line.

### DIFF
--- a/library/Fission/Internal/UTF8.hs
+++ b/library/Fission/Internal/UTF8.hs
@@ -12,11 +12,11 @@ module Fission.Internal.UTF8
   , wrapIn
   ) where
 
+import           Flow
+import           RIO
 import qualified RIO.ByteString      as Strict
 import qualified RIO.ByteString.Lazy as Lazy
 import qualified RIO.Text            as Text
-
-import Fission.Prelude hiding (encode)
 
 class Textable a where
   encode :: a -> Either UnicodeException Text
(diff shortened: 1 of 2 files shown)


library/Fission/Internal/UTF8.hs
```diff
@@ -10,12 +12,12 @@ module Fission.Internal.UTF8
   , wrapIn
   ) where
 
+import           Flow
+import           RIO
```
expede: Can't use `Fission.Prelude` because it forms a dependency cycle 🤷‍♀️

library/Fission/Internal/UTF8.hs
```diff
@@ -57,5 +59,9 @@ stripN n = Text.dropEnd i . Text.drop i
 putText :: MonadIO m => Text -> m ()
 putText = Strict.putStr . encodeUtf8
 
+-- | Helper for printing 'Text' to a console with a newline at the end
+putTextLn :: MonadIO m => Text -> m ()
+putTextLn txt = putText <| txt <> "\n"
```
expede: Now we can write things like:

```haskell
import Fission.Prelude
-- ...
do
  putTextLn "Hello world"
  putTextLn "This is on a newline"
```

PR #203, opened 2019-11-13
Move PG dependency from `library` to `app`

expede:
Liberate the `fission-web` library from the tyranny of PostgreSQL. This makes the downstream CLI have zero dependencies, which is amazing 🎉

Things to note:

* `data Config` is now `data Config dbBackend`
  * Doesn't change _anything_ for you day to day, because we've been using typeclasses
* Some modules are now in `app` instead of `library`
  * Only ones that are tightly coupled to `app/Main.hs` _specifically_
  * Again, no changes required to your normal coding routine
  * Keep as much in `library` as possible, as usual
* This _technically_ breaks backwards compatibility, so bumped to version `2.0.0`
  * No longer exporting PG-specific modules
  * Please let me know if I'm being too pedantic!


package.yaml
```diff
@@ -1,5 +1,5 @@
 name: fission-web-api
-version: '1.20.0'
+version: '2.0.0'
```
dholms: :tada: :tada: :tada:

expede:
Pretty sure Travis is failing because it's caching an old version of the Stack resolver. But maybe not... we'll have to see in the morning. If it passes and someone wants to merge, feel free. Goodnight 👋

PR #206, opened 2019-11-13
Return s3 node on /ipfs/peers

dholms:
## Problem
GET /ipfs/peer only returns web-api node address

## Impact
When a user pins something, the s3 node has to through discover the user's node through the web-api node. Sometime's this takes longer than it should. We can skip that connectivity period by returning the s3 node address as well

##Solution
return s3 node address & web-api node address on GET /ipfs/peer

Closes https://github.com/fission-suite/web-api/issues/205

### DIFF
--- a/library/Fission/Web/IPFS/Peer.hs
+++ b/library/Fission/Web/IPFS/Peer.hs
@@ -22,5 +22,5 @@ get
      )
   => RIOServer cfg API
 get = getExternalAddress >>= \case
-  Right peers -> return <| [IPFS.Peer.fission ] <> peers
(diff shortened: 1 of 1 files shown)


library/Fission/Web/IPFS/Peer.hs
```diff
@@ -22,5 +22,5 @@ get
      )
   => RIOServer cfg API
 get = getExternalAddress >>= \case
-  Right peers -> return peers
+  Right peers -> return <| [IPFS.Peer.fission ] <> peers
```
expede: You can also write this as `return (IPFS.Peer.fission : peers)` 😉

PR #209, opened 2019-11-19
Recursive add with optional ignores

dholms:
## Problem
The current `addDir` function uploads all files. It's impossible to "exclude" certain files.

### Impact
Files that you don't want shared (for instance, .env or .Ds_Store can end up being uploaded to IPFS.
This is a prerequisite to solving https://github.com/fission-suite/cli/issues/31

## Solution
Add an additional function that recursively adds a dir to IPFS by adding individual files, and manually creating DAG nodes. Allow files to be excluded by providing a list of file globs for "ignored" files/folders.


library/Fission/IPFS/DAG/Link/Types.hs
```diff
@@ -0,0 +1,21 @@
...
+  , name :: IPFS.Name
+  , size :: Integer
+  } deriving (Show, Eq, Generic)
+
+
+instance ToJSON Link where
+  toJSON (Link cid name size) =
+    Object [ ("Name", String . T.pack <| unName name)
+           , ("Size", Number  <| fromIntegral size)
+           , ("Cid", Object [("/", String <| unaddress cid)])
```
dholms: looks a little weird, but that's how ipfs does dag links :man_shrugging:

library/Fission/Storage/IPFS/Add.hs
```diff
@@ -0,0 +1,167 @@
...
+import qualified Fission.IPFS.Process        as IPFS.Proc
+import           Fission.IPFS.Error          as IPFS.Error
+import           Fission.IPFS.Types          as IPFS
+import           Fission.IPFS.DAG.Node.Types as DAG
+import           Fission.IPFS.DAG.Link       as DAG.Link
+
+import           Fission.Storage.IPFS.DAG as DAG
+import qualified Fission.Storage.IPFS.Pin as IPFS.Pin
+
+addRaw ::
```
dholms: Not sure if people saw my msg on discord about the syntax. I decided to write this PR with `::` on the same line as the function name. This fixes my syntax highlighting issues, but I also just think it's clearer because it's immediately apparent that that line is the start of a function type. I figured I'd do it with this PR to see how everyone felt about it (and make my life easier :wink:). Not sure how strong of convention this is, so let me know how you feel about it!

library/Fission/Storage/IPFS/Add.hs
```diff
@@ -0,0 +1,167 @@
...
+  ( MonadRIO cfg m
+  , HasProcessContext cfg
+  , HasLogFunc cfg
+  , Has IPFS.BinPath cfg
+  , Has IPFS.Timeout cfg
+  )
+  => [Glob.Pattern]
+  -> FilePath
+  -> m (Either IPFS.Error.Add IPFS.CID)
+addDir ignored path = doesFileExist path >>= \case
```
dholms: removed the old `addDir` is there any time when we'd want to do this natively? I tested it out and got the exact same hashes natively & through this

library/Fission/Storage/IPFS/Add.hs
```diff
@@ -0,0 +1,167 @@
...
+        _  -> return . Left <| firstErr links
+    _  -> return . Left <| firstErr results
+
+removeIgnored :: [Glob.Pattern] -> [FilePath] -> [FilePath]
+removeIgnored ignored files = List.filter (not . matchesAny ignored) files
+
+matchesAny :: [Glob.Pattern] -> FilePath -> Bool
+matchesAny globs path = List.any (\x -> Glob.match x path) globs
+              
+firstErr :: [Either a b] -> a
```
dholms: Thoughts on this? Given a list of potentially multiple errors, we break and return the first in the list

library/Fission/Storage/IPFS/Add.hs
```diff
@@ -0,0 +1,167 @@
...
+  files <- listDirectory path
+  let toAdd = removeIgnored ignored files
+  results <- mapM (addDir ignored . combine path) toAdd 
+  case lefts results of
+    [] -> do
+      let names = IPFS.Name <$> toAdd
+      links <- zipWithM DAG.Link.create (rights results) names
+      case lefts links of
+        [] -> DAG.putNode Node
+              { dataBlock = "CAE="
```
dholms: Never did find out what this mysterious data is :sob:. Still looking around. Parsing w/ the multiformats library gave me an error :thinking:

package.yaml
```diff
@@ -91,6 +91,7 @@ dependencies:
   - flow
   - filepath
   - fsnotify
+  - Glob
```
dholms: I hate that this package is capitalized :unamused:

library/Fission/Storage/IPFS/Add.hs
dholms: `CAE=` == `\u0008\u0001`
https://medium.com/@php.laboratory/you-are-right-that-mysterious-u0008-u0001-came-from-protobuf-its-serialized-directory-1-from-dc3a42b59db8

This is from protobufs and `CAE=` just means "directory". I guess we can leave this hardcoded for now? It's the only type of IPLD node we need to manually generate atm. If/when we start adding other types of IPLD nodes, we could look into implementing protobufs for haskell or switching over to a different dag format

dholms: Actually [dag-pb](https://github.com/ipld/go-ipld-deprecated) seems to be deprecated. But it's still the default when doing a recursive add. Some experimenting with `dag-cbor` didn't work with the gateway, but there's the possibility that I'm doing something incorrectly there

library/Fission/Storage/IPFS/Add.hs
expede: On a mechanical note, won't this fail at runtime if there are no `Left`s?

library/Fission/Storage/IPFS/Add.hs
expede: The convention isn't super strong / just emerging. This is fine IMO 👍

library/Fission/Storage/IPFS/Add.hs
expede: Have you considered using something like `foldM` instead of `mapM` above? If you want to fail on the first `Left`, that may be a way to do it rather than deconstructing a mix of `Either`s to see if there's ant `Left`s

library/Fission/Storage/IPFS/Add.hs
expede: I can't see this commit anymore apparently, but from memory: we don't do it natively because this module assumes that we're on the EC2 instance, and actual storage happens on the S3 node.

dholms: ah sorry, I think "natively" was the wrong word to use here. I just meant our recurse vs IPFS' recurse. In other words, manually creating DAG nodes vs `ipfs add -r ...`

Neither one should affect that datastores on the back end. And `addDir` is never actually called in the Web API, only the CLI. This is because files to be added through the API, are serialized and added as raw data. Directories are never crawled.

library/Fission/Storage/IPFS/Add.hs
dholms: In the only place I'm using it, I check to make sure that some `lefts` is not empty. But I shouldn't rely on a check like that to keep this from blowing up.

I reimplemented this using `foldM`. I think it definitely makes it easier to grok. Let me know what you think! 

Is there a way to fail and exit `foldM` after encountering your first `Left`? Right now I"m just pattern matching and returning a `Left`.

library/Fission/IPFS/DAG/Link/Types.hs
expede: 🤷‍♀️

library/Fission/Storage/IPFS/Add.hs
expede: >  And addDir is never actually called in the Web API, only the CLI

We really need to split this into its own library. It's starting to get confusing.

package.yaml
expede: I know, right?! Especially for something that's fairly widely used 😭

library/Fission/Storage/IPFS/Add.hs
```diff
@@ -0,0 +1,178 @@
...
+  -> m (Either IPFS.Error.Add Node)
+foldResults _ _ (Left err) _ = return <| Left err
+foldResults path ignored (Right node) filename = do
+  addDir ignored (path </> filename) >>= \case
+    Left err ->  return <| Left err
+    Right cid ->
+      DAG.Link.create cid (IPFS.Name filename) >>= \case
+      Left err -> return <| Left err
+      Right link ->
+        return <| Right <| node { links = link:(links node) }
```
expede: Moving the conversation here from the `Outdated` comment (which is thus hidden). This looks great! Yes, the correct thing to do to "exit early" is

```haskell
foldResults _ _ (Left err) _ = return <| Left err
```

library/Fission/Storage/IPFS/Add.hs
```diff
@@ -0,0 +1,178 @@
...
+  => IPFS.Ignored
+  -> FilePath
+  -> m (Either IPFS.Error.Add IPFS.CID)
+walkDir ignored path = do
+  files <- listDirectory path
+  let toAdd = removeIgnored ignored files
+  let f = foldResults path ignored
+  foldM f (Right <| Node { dataBlock = "CAE=", links = [] }) toAdd >>= \case
+    Left err -> return <| Left err
+    Right node -> DAG.putNode node
```
expede: This is _totally fine as is_, but have you considered breaking out the `Node` into a named variable for legibility?

```haskell
walkDir ignored path = do
  files <- listDirectory path

  let
    toAdd   = removeIgnored ignored files
    reducer = foldResults path ignored
    seed    = Right <| Node 
                { dataBlock = "CAE="
                , links     = [] 
                }

  foldM reducer seed toAdd >>= \case
    Left  err  -> return <| Left err
    Right node -> DAG.putNode node
```

dholms: Love it :ok_hand:

PR #211, opened 2019-11-20
Gateway env

dholms:
## Problem
- gateway not in env
- `domain_name` starts with a `.`

## Solution
- create env variable
- add `.` to `domain_name` in code instead of env file
- refactor `registerDomain` function to eliminate WET code

A bit of clarification on variables: 
1) `ipfs.gateway` is our ipfs gateway (ie `ipfs.runfission.com`). This gets set as the cname value for any domain a user registers
2) `aws.domain_name` is the base domain name that we use for user's domain names (ie `fission.name`)

The issue (https://github.com/fission-suite/web-api/issues/162) was initially misframed because we were using `demo.runfission.com` as our `aws.domain_name` so both domains could be derived from the same base (`runfission.com`), meaning we didn't need two separate variables.

Closes https://github.com/fission-suite/web-api/issues/162

### DIFF
--- a/app/Fission/Environment/IPFS/Types.hs
+++ b/app/Fission/Environment/IPFS/Types.hs
@@ -8,12 +8,14 @@ data Environment = Environment
   { url     :: !IPFS.URL     -- ^ IPFS client URL (may be remote)
   , timeout :: !IPFS.Timeout -- ^ IPFS timeout in seconds
   , binPath :: !IPFS.BinPath -- ^ Path to local IPFS binary
+  , gateway :: !IPFS.Gateway -- ^ Domain Name of IPFS Gateway
   } deriving Show
 
 instance FromJSON Environment where
   parseJSON = withObject "IPFS.Environment" \obj -> do
     timeout <- obj .:? "timeout" .!= 3600
     binPath <- obj .:? "binPath" .!= "/usr/local/bin/ipfs"
+    gateway <- obj .:? "gateway" .!= "ipfs.runfission.com"
     url     <- obj .:  "url" >>= parseJSON . String
 
     return <| Environment {..}
--- a/library/Fission/AWS.hs
+++ b/library/Fission/AWS.hs
@@ -1,6 +1,5 @@
 module Fission.AWS
   ( createEnv
-  , ensureContent
   , validate
   , withAWS
   ) where
@@ -41,16 +40,3 @@ validate changeSet =
 
   where
     status = changeSet ^. crrsrsResponseStatus
-
--- | Ensure that a request completed, and that the status code is not in an error range
-ensureContent
-  :: ( MonadRIO   cfg m
-     , MonadThrow     m
-     , Exception err
-     )
-  => m (Either err ChangeResourceRecordSetsResponse)
-  -> m ChangeResourceRecordSetsResponse
-ensureContent runRequest = do
-  errOrResp <- runRequest
-  resp      <- ensureM errOrResp
-  resp |> validate |> ensureM
--- a/library/Fission/AWS/Route53.hs
+++ b/library/Fission/AWS/Route53.hs
@@ -41,12 +41,10 @@ registerDomain username (CID hash) = do
     dnslinkUrl = "_dnslink." <> baseUrl
     dnslink    = "dnslink=/ipfs/" <> hash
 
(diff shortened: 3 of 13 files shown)


library/Fission/AWS.hs
```diff
@@ -41,16 +40,3 @@ validate changeSet =
 
   where
     status = changeSet ^. crrsrsResponseStatus
-
--- | Ensure that a request completed, and that the status code is not in an error range
-ensureContent
```
dholms: The resp already gets validated in `Route53.changeRecord`

And now `registerDomain` returns `m (Either ServerError AWS.DomainName)`. We can just `ensureM` in the route function.

If I'm missing something here please let me know! But with the refactor this function seems redundant

app/Fission/Environment/IPFS/Types.hs
```diff
@@ -8,12 +8,14 @@ data Environment = Environment
   { url     :: !IPFS.URL     -- ^ IPFS client URL (may be remote)
   , timeout :: !IPFS.Timeout -- ^ IPFS timeout in seconds
   , binPath :: !IPFS.BinPath -- ^ Path to local IPFS binary
+  , gateway :: !IPFS.Gateway -- ^ Domain Name of IPFS Gateway
```
expede: How is this different from `url`?

dholms: `url` is the url of the http interface for some node (ie `localhost:5001`)
`gateway` is the url of the gateway that we want to point dns records to (ie `ipfs.runfission.com`)

dholms:
When testing/deploying to prod, make sure to update your `env.yaml`!

You need to 
- add `ipfs.gateway` (ie `ipfs.runfission.com`)
- delete the `.` at the front of `aws.domain_name` (ie `fission.name`)

app/Fission/Environment/IPFS/Types.hs
expede: Sorry, I'm clearly coming at this without enough context. What feature does this enable?

dholms: actually doesn't enable a feature. This is fixing https://github.com/fission-suite/web-api/issues/162 (https://trello.com/c/I5ETfd7L/826-fix-gateway-in-env)

the issue was slightly misframed which I addressed in this comment: https://github.com/fission-suite/web-api/issues/162#issuecomment-556254891

library/Fission/AWS/Route53.hs
```diff
@@ -3,30 +3,66 @@ module Fission.AWS.Route53
...
+registerDomain username (CID hash) = do
+  gateway :: IPFS.Gateway   <- Config.get
+  domain  :: AWS.DomainName <- Config.get
+
+  let
+    baseUrl    = username <> "."<> AWS.getDomainName domain
+    dnslinkUrl = "_dnslink." <> baseUrl
+    dnslink    = "dnslink=/ipfs/" <> hash
+
+  logDebug <| displayShow ("~~" <> (wrapIn dnslink "\"") <> "~~")
```
expede: Totally fine as-is, but cool trick to make a function infix (reads like pseudo-English)

```haskell
logDebug <| displayShow ("~~" <> dnslink `wrapIn` "\"" <> "~~")
```

library/Fission/Config/Types.hs
```diff
@@ -73,6 +75,10 @@ instance Has IPFS.Timeout (Config db) where
   hasLens = lens ipfsTimeout \cfg newIPFSTimeout ->
     cfg { ipfsTimeout = newIPFSTimeout }
 
+instance Has IPFS.Gateway (Config db) where
+  hasLens = lens ipfsGateway \cfg newIPFSGateway ->
+    cfg { ipfsGateway = newIPFSGateway }
```
expede: 👍

library/Fission/Web/DNS.hs
```diff
@@ -5,53 +5,33 @@ module Fission.Web.DNS
...
-
-  dnslink
-    |> wrapIn' "\""
-    |> registerDomain Route53.Txt dnslinkUrl
-    |> ensureContent
-
-  return <| AWS.DomainName baseUrl
-
-wrapIn' :: Text -> Text -> Text
-wrapIn' a b = wrapIn b a
```
expede: Love it when we delete code 🥳

PR #212, opened 2019-11-26
Ipfs refactor

dholms:
## Problem
IPFS modules are all in `fission-web-api`. These would be useful as a separate library. This would also decrease our CLI's reliance on `fission-web-api`

## Solution
- Break out IPFS modules into their own library (https://github.com/fission-suite/ipfs-haskell)
- Create IPFS monads for ease of use


dholms:
Outstanding questions/issues:

**Best way to deal with code that's need by both `ipfs` and `fission-web-api`?**
We briefly discussed this but got distracted before reaching a resolution. Some examples include
- `Fission.Prelude`
- `Internal.UTF8`
- various orphans

**What all should be exported in `Network.IPFS` (ie the base package)?**
Right now I only export the classes. Do we want to export types (from `IPFS.Types`) / functions (like `add`, `pin`) as well? My current thought is export types but not functions. Although this means either maintaining an annoying import/export chain or getting rid of `IPFS.Types` altogether.

**What should I do about `getTimeout` on `MonadLocalIPFS`?**
This feels like a hack, but maybe it's okay?
`Network.IPFS.Get.getFile` needs `Has IPFS.Timeout cfg` as a constraint. Since this is a constraint on `MonadLocalIPFS` shouldn't it implicitly have `Has IPFS.Timeout cfg` as a constraint

**`MonadLocalIPFS m` vs `MonadLocalIPFS (RIO cfg)`**
Related to above^^ I kind of get the different between these just from context, but I'm not quite grokking what's going on. You said `(RIO cfg)` == `m`. Do you jsut use `RIO cfg` when other constraints need to reference the `cfg`?

**Error handling**
I didn't change much in terms of error handling, because our current error setup for processes was already returning everything we needed to process the error in the library code (exit code, stdout, stderr). Am I missing something here?

**Move instances**
Instances for the monad classes are still sitting in the library. I'm going to give moving these a go, but might need a hand /second set of eyes

expede:
> **Move instances**
> ~Instances for the monad classes are still sitting in the library. I'm going to give moving these a go, but might need a hand /second set of eyes~
Moved these instances from the library to the server. Thought it was going to be more complex than it was for some reason. Let me know if they look good!

Nope, that's perfect 👍 That should be easy -- yay abstraction!

expede:
> **Error handling**
> I didn't change much in terms of error handling, because our current error setup for processes was already returning everything we needed to process the error in the library code (exit code, stdout, stderr). Am I missing something here?

I'll take a deeper dive into the code, but those are very tied to the `RIO.Process` system. We should have errors more specific to IPFS, which get returned in the `class`. I'll take a look at what's there now, but wanted to get to this down between meetings.

app/Fission/Internal/Orphanage/RIO.hs
```diff
@@ -20,3 +25,35 @@ instance Has (DB.Pool PG) cfg => MonadSelda (RIO cfg) where
...
+  , Has HTTP.Manager cfg
+  )
+  => MonadRemoteIPFS (RIO cfg) where
+    runRemote query = do
+      IPFS.URL url            <- Config.get
+      manager :: HTTP.Manager <- Config.get
+      url
+        |> mkClientEnv manager
+        |> runClientM query
+        |> liftIO
```
expede: Yeah, this file belongs in the app. People use `RIO` in many different ways. Totally get that you need this to test development, though

expede: You could also create a `newtype` for testing

expede:
Context:

> **Best way to deal with code that's need by both `ipfs` and `fission-web-api`?**
> We briefly discussed this but got distracted before reaching a resolution. Some examples include

> - `Fission.Prelude`

I'd say copy/paste the bits that you need for the library. We don't need something shared, especially going forward (may create new dependencies, &c)

> - `Internal.UTF8`

See above

> - various orphans

Ideally the orphans in `Network.IPFS` whould be `newtype`s, not orphans. The whole orphan thing that we have going on is a mix of pragmatism and hubris (when I was writing everything myself).

expede:
> **What should I do about `getTimeout` on `MonadLocalIPFS`?**
> This feels like a hack, but maybe it's okay?
> `Network.IPFS.Get.getFile` needs `Has IPFS.Timeout cfg` as a constraint. Since this is a constraint on `MonadLocalIPFS` shouldn't it implicitly have `Has IPFS.Timeout cfg` as a constraint

It can be a constraint on the `instance`. I'd put it there, not on the `class`.

expede:
> **Error handling**
> I didn't change much in terms of error handling, because our current error setup for processes was already returning everything we needed to process the error in the library code (exit code, stdout, stderr). Am I missing something here?

A think missing a bit, yeah. `(exitCode, stdout, stderr)` is _pretty concrete_. In terms of the `class`, think about if we were going to mock it out in tests and not actually hit IPFS. You'd want some kind of semantic error type to deal with. `(exitCode, stdout, stderr)` is a bit low-level/granular. What we want as an interface is probably something more along the lines of

```haskell
data Error
  | Timeout CID
  | PeerUnavailable Peer
  = Unknown RawMessage

type RawMessage = Text
```

expede:
Oh, actually the last comment is for a general class. Yes, the current version actually makes sense since we're assuming that we're shelling out.

If you wanted to do a combined general class, it would look something like this:

```haskell
class Monad m => MonadIPFS m where
  cat   :: CID  -> m (Either GetError   File)
  add   :: File -> m (Either AddError   CID)
  pin   :: CID  -> m (Either PinError   ())
  unpin :: CID  -> m (Either UnpinError ())

class Monad m => MonadIPFSPeer
  listPeers :: m [Peer]
  connect   :: Peer -> m Bool
```

Where implement would have to define each function, but can rely on their own helpers, &c

🤔 Actually, it can't hurt to add that to the library now:

```haskell
instance MonadIPFSLocal m => MonadIPFS m where
  cat = getFileOrDirectory
  -- ...
```

dholms:
Alrighty redid some of the error handling. I went about halfway between the two ideas, so hopefully it's the best of both worlds and not the worst :joy: 

Check out the PR here: https://github.com/fission-suite/ipfs-haskell/pull/2

I don't think we want to make a general combined class `MonadIPFS` or have the same methods on both the remote & local classes. I believe we discussed the issues with this at some point last week. There are a good deal of methods in the web-api that use both classes, and I think we'll end with confusion because of that. Also we're going to have a hard time getting the apis of the two to match because of things like multipart form uploads etc.

Anyways, I refactored the error handling code. Now `MonadLocalIPFS` only has once instance. Most of the error handling needs more context (ie "what cmd gave this error?"), but some is more general and can be taken care of in this instance (ie timeouts). This also fixes the `getTimeout` issue that I was asking about above^^

LMK if it looks like I got it about right. I think the main thing is striking the right balance between error handling in the instance method vs in the implementation methods

app/Fission/Internal/Orphanage/RIO.hs
```diff
@@ -3,20 +3,64 @@
...
+      let opts' = ("--timeout=" <> show secs <> "s") : opts
+
+      runProc readProcess ipfs (byteStringInput arg) byteStringOutput opts' >>= \case
+        (ExitSuccess, contents, _) ->
+          return <| Right contents
+        (ExitFailure _, _, stdErr)
+          | Lazy.isSuffixOf "context deadline exceeded" stdErr -> 
+              return . Left <| Process.Timeout secs
+          | otherwise ->
+            return . Left <| Process.UnknownErr stdErr
```
expede: 🤔 Any thoughts about doing a `swarm connect` before every action right in the `instance`? Just port in the kludge from the CLI and automate away the details so we don't have to look at the hack and feel bad 🤣

app/Fission/Internal/Orphanage/RIO.hs
```diff
@@ -3,20 +3,64 @@
...
+  , Has HTTP.Manager cfg
+  )
+  => MonadRemoteIPFS (RIO cfg) where
+    runRemote query = do
+      IPFS.URL url            <- Config.get
+      manager :: HTTP.Manager <- Config.get
+      url
+        |> mkClientEnv manager
+        |> runClientM query
+        |> liftIO
```
expede: 🤔 Any thoughts about doing a swarm connect before every action right in the instance? Just port in the kludge from the CLI and automate away the details so we don't have to look at the hack and feel bad 🤣

library/Fission/Internal/Orphanage/ToServerError.hs
```diff
@@ -0,0 +1,47 @@
...
+
+    500 -> err500
+    501 -> err501
+    502 -> err502
+    503 -> err503
+    504 -> err504
+    505 -> err505
+
+    n -> error <| show n <> " is not an error status code"
+
```
expede: Wait, why is this an orphan?

expede: I see. I moved to the helper lib 🤔

[TL;DR think "separation of concerns"]

I would personally have the class live in the application, not in the IPFS lib. The IPFS lib needs to know which error to throw, but doesn't need a generic method to convert into HTTP errors, right? If you need to set IPFS errors to Servant errors, we can do that in the module where `ToServantError` is defined.

Am I missing something?

library/Fission/Web/Heroku.hs
```diff
@@ -110,13 +104,12 @@ provision Request {uuid, region} = do
...
-     , HasLogFunc        cfg
-     , Has HTTP.Manager  cfg
-     , Has IPFS.URL      cfg
-     )
+deprovision ::
+  ( MonadSelda      (RIO cfg)
+  , MonadMask       (RIO cfg)
+  , MonadRemoteIPFS (RIO cfg)
+  , HasLogFunc           cfg
+  )
```
expede: Awwwwwwwwww yeeeeeeees. We'd need to implement [one more class](https://www.stackage.org/haddock/lts-14.16/monad-logger-0.3.30/Control-Monad-Logger.html#t:MonadLogger), but check this out:

```haskell
instance HasLogFunc cfg => MonadLogger (RIO cfg) where
  monadLoggerLog _loc _src level msg =
    logger <- Config.get
    logger txt level -- or something

deprovision ::
  ( MonadSelda      m
  , MonadMask       m
  , MonadRemoteIPFS m
  , MonadLogger     m
  )
  => ServerT DeprovisionAPI m
deprovision uuid' = do
  let err = Web.Err.ensureMaybe err410 -- HTTP 410 is specified by the Heroku AddOn docs
  let err = Web.Err.ensureMaybe err410

Comment (expede) [reply library/Fission/Web/Heroku.hs]: i.e. no more tracking `cfg` all over the place. Constraints == capabilities.

File: app/Fission/Internal/Orphanage/RIO.hs (hunk 3 above)
Comment (dholms) [reply app/Fission/Internal/Orphanage/RIO.hs]: :thinking: :thinking: 
We wouldn't want that in the web-api would we? I dig the idea for in the cli when we write these instances tho

Comment (expede) [reply app/Fission/Internal/Orphanage/RIO.hs]: Aren't we still having issues where the web-api randomly can't find the S3 node? The reconnection script may not be frequent enough or... something? I dunno. I had this flagged in the back of my brain for a few weeks, so that info may be stale. Thoughts?

File: library/Fission/Internal/Orphanage/ToServerError.hs (hunk 4 above)
Comment (dholms) [reply library/Fission/Internal/Orphanage/ToServerError.hs]: yeah that makes sense.

Would it be bad practice to put all the `ToServantError` orphans in the module where the class is defined?

File: library/Fission/Web/Heroku.hs (hunk 5 above)
Comment (dholms) [reply library/Fission/Web/Heroku.hs]: awwww yiiisss

File: app/Fission/Internal/Orphanage/RIO.hs (hunk 3 above)
Comment (dholms) [reply app/Fission/Internal/Orphanage/RIO.hs]: ah right that's true. yup I'll go ahead and do it in both then

File: library/Fission/Internal/Orphanage/ToServerError.hs (hunk 4 above)
Comment (expede) [reply library/Fission/Internal/Orphanage/ToServerError.hs]: # Short Answer

It's not an orphan if you define it in the same module as the `class`. Let's do that 👍

# Long Answer

"Orphans" only refer to a specific edge case. By default, Haskell wants you to define `instance`s _either_ where the `data` is defined, or where the `class` is defined.

Orphans arise when you have data from a library that you don't control, and a class from a library that you don't control, and you want to implement them together without wrapping in a `newtype`.

Comment (dholms) [reply library/Fission/Internal/Orphanage/ToServerError.hs]: ahhh got it. I thought it was specifically where the `data` was defined. Nice that makes sense :+1:

Comment (expede) [reply library/Fission/Internal/Orphanage/ToServerError.hs]: FYI, rounded out that answer a bit more

File: app/Fission/Internal/Orphanage/RIO.hs (hunk 3 above)
Comment (dholms) [reply app/Fission/Internal/Orphanage/RIO.hs]: Tagging so this doesn't get lost (@expede)

So in order to connect, you need to be in a `MonadLocalIPFS`. I'm not sure we want to attach that as a constraint on our `MonadRemoteIPFS` instance. Thoughts?

Comment (expede) [reply app/Fission/Internal/Orphanage/RIO.hs]: That's a good point 🤔 You could do something along these lines:

```haskell
class Monad m => MonadRemoteIPFS m where
  runRemote :: ClientM a -> m(Either ClientError a)
  ipfsConnect :: Peer -> m (Either ConnectionError ())
  ipfsAdd   :: Lazy.ByteString -> m (Either ClientError CID)
  -- and so on ...
  ipfsAdd raw = runRemote <| IPFS.Client.add raw
  ipfsConnect _ = return () -- Noop
  -- and the rest of the defaults

instance 
  ( Has IPFS.URL     cfg
  , Has HTTP.Manager cfg
  , Has Peer         cfg
  , MonadRIO         cfg m
  , MonadLocalIPFS m
  ) 
  => MonadRemoteIPFS m where
    runRemote = do
      Peer peerID             <- Config.get
      IPFS.URL url            <- Config.get
      manager :: HTTP.Manager <- Config.get

      Local.connectPeer peerID

      url
        |> mkClientEnv manager
        |> runClientM query
        |> liftIO
```

library/Fission/Web.hs
```diff
@@ -42,26 +42,22 @@ import qualified Fission.Web.Heroku            as Heroku
...
+  , Has AWS.AccessKey    cfg
+  , Has AWS.SecretKey    cfg
+  , Has AWS.ZoneID       cfg
+  , Has AWS.DomainName   cfg
+  , Has Heroku.ID        cfg
+  , Has Heroku.Password  cfg
+  , HasLogFunc           cfg
+  , MonadSelda      (RIO cfg)
+  )
+  => RIO cfg Application
```
expede: Yay getting shorter!!

library/Fission/Web/Error.hs
```diff
@@ -58,6 +62,40 @@ instance ToServerError Int where
...
+    AddErr           addErr -> toServerError addErr
+    GetErr           getErr -> toServerError getErr
+    LinearizationErr linErr -> toServerError linErr
+
+instance ToServerError Get.Error where
+  toServerError = \case
+    Get.InvalidCID txt          -> err422 { errBody = UTF8.textToLazyBS txt }
+    Get.UnexpectedOutput _ -> err500 { errBody = "Unexpected IPFS result" }
+    Get.UnknownErr _         -> err500 { errBody = "Unknown IPFS error" }
+    (Get.TimedOut (CID hash) _) ->
```
expede: Does this need the outer brackets?

dholms: nope that was an artifact from.... something

expede:
[Classy IPFS in web-api](https://trello.com/c/aTq98GUs/868-classy-ipfs-in-web-api)

PR #213, opened 2019-11-26
Fix Swagger Docs to allow for codegeneration

### LINKED ISSUE #207: API docs at runfission.com generate malformed curl commands
>I tried running the `PUT /dns/{cid}` command from the Swagger API but got an error due to >malformed URL. This is the `curl` command that the Web UI generated:
>
```
curl -X PUT "https://"https://runfission.com"/dns/QmaYSP6h3AvqKCrsS9hvV6Zp7ZoKQTchJ737ezXxcNJj31" -H  "accept: text/plain;charset=utf-8"
```
>
>At the top of the page it says `[ Base URL: "https://runfission.com" ]`, so that's likely the issue.
>
>I fixed the URL and ran it from the command-line and it worked \o/

_Originally posted by @agentofuser in https://github.com/fission-suite/cli/issues/9#issuecomment-554090623_

bechurch:
## Summary

* [x] Host being rendered with quotes
* [x] Host being rendered with scheme
* [x] OpenAPI refs containing spaces
* [x] Removes an invalid/redundant file upload param specification

Fixing these makes our openapi spec valid again

## Closing issues

closes #207

## After Merge

### DIFF
--- a/library/Fission/Internal/Orphanage/MultipartForm.hs
+++ b/library/Fission/Internal/Orphanage/MultipartForm.hs
@@ -2,6 +2,7 @@
 
 module Fission.Internal.Orphanage.MultipartForm () where
 
+import Data.Swagger
 import Network.HTTP.Media.MediaType
 import Servant
 import Servant.Multipart
@@ -10,7 +11,17 @@ import Servant.Swagger.Internal
 
 import Fission.Prelude
 
+
 instance HasSwagger api => HasSwagger (MultipartForm Mem (MultipartData Mem) :> api) where
   toSwagger _ = Proxy @api
-             |> toSwagger
-             |> addConsumes ["multipart" // "form-data"]
+      |> toSwagger
+      |> addConsumes ["multipart" // "form-data"]
+      |> addParam param
+    where
+      param = mempty
+        |> name .~ "file"
+        |> required ?~ True
+        |> description ?~ "mempty |> type_ .~ SwaggerFile"
+        |> schema .~ ParamOther (mempty
+            |> in_ .~ ParamFormData
+            |> paramSchema .~ (mempty |> type_ ?~ SwaggerFile))
--- a/library/Fission/Internal/Orphanage/MultipartForm.hs
+++ b/library/Fission/Internal/Orphanage/MultipartForm.hs
@@ -13,15 +13,23 @@ import Fission.Prelude
 
 
 instance HasSwagger api => HasSwagger (MultipartForm Mem (MultipartData Mem) :> api) where
-  toSwagger _ = Proxy @api
+  toSwagger _ =
+    Proxy @api
       |> toSwagger
       |> addConsumes ["multipart" // "form-data"]
       |> addParam param
     where
-      param = mempty
-        |> name .~ "file"
-        |> required ?~ True
-        |> description ?~ "A file to upload (may also be multipart/form-data)"
-        |> schema .~ ParamOther (mempty
-            |> in_ .~ ParamFormData
-            |> paramSchema .~ (mempty |> type_ ?~ SwaggerFile))
+      param =
+        mempty
+          |> name        .~ "file"
+          |> required    ?~ True
+          |> description ?~ "A file to upload (may also be multipart/form-data)"
+          |> schema      .~ ParamOther fileParam
+
+      fileParam =
+        mempty
+          |> in_         .~ ParamFormData
+          |> paramSchema .~ swaggerFileType
+
+      swaggerFileType =
+        mempty |> type_ ?~ SwaggerFile
--- a/library/Fission/Internal/Orphanage/MultipartForm.hs
+++ b/library/Fission/Internal/Orphanage/MultipartForm.hs
@@ -2,7 +2,6 @@
 
 module Fission.Internal.Orphanage.MultipartForm () where
 
-import Data.Swagger
 import Network.HTTP.Media.MediaType
 import Servant
 import Servant.Multipart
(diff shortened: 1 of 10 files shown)


library/Fission/Internal/UTF8.hs
```diff
@@ -55,6 +57,10 @@ stripN n = Text.dropEnd i . Text.drop i
     i :: Int
     i = fromIntegral n
 
+-- | Helper for showing nytestring without quotes
+stripQuotes :: Char8.ByteString -> [Char]
```
dholms: This is probably fine as is. I would either
- rename this something that follows the pattern `stripN...` since this doesn't actually "strip quotes" it strips one char on each side
- or since this function is only used once, just compose functions in `Fission.Web.Types` instead of writing a new function

library/Fission/Internal/Orphanage/MultipartForm.hs
```diff
@@ -15,9 +14,3 @@ instance HasSwagger api => HasSwagger (MultipartForm Mem (MultipartData Mem) :>
   toSwagger _ = Proxy @api
              |> toSwagger
              |> addConsumes ["multipart" // "form-data"]
-             |> addParam param
```
dholms: Just curious why we removed the param here. Seems like this route isn't very useful without an uploaded file param

expede:
FYI I fixed some of the issues over at #220 -- double quotes and HTTPS prefixing in particular

library/Fission/Internal/Orphanage/MultipartForm.hs
```diff
@@ -11,13 +11,17 @@ import Servant.Swagger.Internal
...
 
+
 instance HasSwagger api => HasSwagger (MultipartForm Mem (MultipartData Mem) :> api) where
   toSwagger _ = Proxy @api
-             |> toSwagger
-             |> addConsumes ["multipart" // "form-data"]
-             |> addParam param
+      |> toSwagger
+      |> addConsumes ["multipart" // "form-data"]
+      |> addParam param
```
expede: 💅

If we're going to move the pipelne to that position, let's also drop the `Proxy @api` to the next line as well. I would also say this in Elixir 😛

```haskell
  toSwagger _ =
    Proxy @api
      |> toSwagger
      |> addConsumes ["multipart" // "form-data"]
      |> addParam param
```

library/Fission/Internal/Orphanage/MultipartForm.hs
```diff
@@ -11,13 +11,17 @@ import Servant.Swagger.Internal
...
       param = mempty
-           |> name        .~ "file"
-           |> description ?~ "A file to upload (may also be multipart/form-data)"
-           |> required    ?~ True
+        |> name .~ "file"
+        |> required ?~ True
+        |> description ?~ "A file to upload (may also be multipart/form-data)"
+        |> schema .~ ParamOther (mempty
+            |> in_ .~ ParamFormData
+            |> paramSchema .~ (mempty |> type_ ?~ SwaggerFile))
```
expede: Ooh, yeah, I'd break this out, too:

Option 1

```haskell
   where
      param = 
        mempty
          |> name .~ "file"
          |> description ?~ "A file to upload (may also be multipart/form-data)"
          |> required ?~ True
          |> schema .~ ParamOther (
               mempty
                 |> in_ .~ ParamFormData
                 |> paramSchema .~ (mempty |> type_ ?~ SwaggerFile)
          )
```

Option 2:

```haskell
   where
      param = mempty
        |> name .~ "file"
           |> description ?~ "A file to upload (may also be multipart/form-data)"
        |> required ?~ True
        |> description ?~ "A file to upload (may also be multipart/form-data)"
        |> schema .~ ParamOther paramSchema

      paramSchema = 
        mempty
          |> in_ .~ ParamFormData
          |> paramSchema .~ fileType

      fileType = 
        mempty |> type_ ?~ SwaggerFile
```

library/Fission/Internal/UTF8.hs
```diff
@@ -14,9 +14,9 @@ module Fission.Internal.UTF8
 
 import           Flow
 import           RIO
-import qualified RIO.ByteString      as Strict
-import qualified RIO.ByteString.Lazy as Lazy
-import qualified RIO.Text            as Text
+import qualified RIO.ByteString             as Strict
+import qualified RIO.ByteString.Lazy        as Lazy
+import qualified RIO.Text                   as Text
```
expede: Why?

library/Fission/Web/Swagger.hs
```diff
@@ -39,10 +38,10 @@ docs host' =
     |> user
 
 app :: HasSwagger api => Proxy api -> Web.Host -> Swagger
-app proxy (Web.Host (BaseUrl { baseUrlHost })) =
+app proxy appHost =
   proxy
     |> toSwagger
-    |> host               ?~ Host baseUrlHost Nothing
+    |> host               ?~ Host (Web.getHostName appHost) Nothing
```
expede: Does that unwrap deeply enough?

EDIT

Oh, I see. There's a helper function. That confused me on first pass, becasue the name is so similar. I'm not sure if `master` changed under you (uses pattern matching) 🤔 Another version could be:

```haskell
host ?~ Host (appHost |> Web.getHostName |> baseUrlHost) Nothing
```

or rename

```haskell
getRawHost :: Host -> String
```

library/Fission/Web/Types.hs
```diff
@@ -17,7 +19,7 @@ newtype Host = Host { getHost :: BaseUrl }
                    )
 
 instance Show Host where
-  show = show . UTF8.stripNBS 1 . encode
+  show = Char8.unpack . UTF8.stripNBS 1 . encode
```
expede: These may be equivalent 🤔

expede: Whatever, they do the same thing 👍 No need to change your PR

library/Fission/Internal/UTF8.hs
bechurch: Oh, I believe I had aligned them with a longer import I later removed.

bechurch:
@expede Updated with your suggestions

PR #215, opened 2019-11-28
Change RIO_VERBOSE -> DEBUG

### LINKED ISSUE #198: Switch envvar `RIO_VERBOSE` to `DEBUG` or similar
## Problem

`RIO_VERBOSE` is the default from `RIO`. It's also annoyingly long to type, and hard for team members to remember.

## Solution

Switch to `DEBUG=true`, or similar

bechurch:
## Summary
Changes the `RIO_VERBOSE` env var to `DEBUG`

## Closing issues
closes #198 

## After Merge

### DIFF
--- a/library/Fission/App.hs
+++ b/library/Fission/App.hs
@@ -1,0 +1,21 @@
+-- | Helpers for running the Fission applications
+module Fission.App (runApp) where
+
+import System.Environment (setEnv, unsetEnv)
+import Fission.Environment (getFlagWithDefault)
+import Fission.Prelude
+
+-- | Run the given simple app.
+--   Set `DEBUG` to `True` to enable verbose app logging
+runApp :: RIO SimpleApp a -> IO a
+runApp simpleApp = do
+  isVerbose  <- getFlagWithDefault "DEBUG" False
+  setRioVerbose isVerbose
+  runSimpleApp simpleApp
+
+-- | Sets the env variable `RIO_VERBOSE` to the given value
+--   Note: unsetting `RIO_VERBOSE` is the only way for it to be interpretted as False
+setRioVerbose :: Bool -> IO ()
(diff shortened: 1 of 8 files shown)


library/Fission/Environment.hs
```diff
@@ -59,3 +60,10 @@ getFlag key = do
   mayStr <- lookupEnv key
   let mayVal = truthy . fmap toLower <$> mayStr
   return mayVal
+
+-- | Get a given flag, default to the given value if not found
+getFlagWithDefault :: String -> Bool -> IO Bool
+getFlagWithDefault flag defaultValue =
+  getFlag flag >>= \case
+    Just value -> return value
+    Nothing -> return defaultValue
```
expede: FYI, we have acombinator for fallback values:

```haskell
getFlag flag .!~ fallback
```

expede: If you don't like operators in particular, we can do a named version, but let's keep these orthogonal:

```haskell
fallback :: Monad m => m (Maybe a) -> a -> m a
fallback = (.!~)
-- AKA: fallback mayValue defValue = pure (fromMaybe defValue) <*> mayValue

"TIMEOUT"
|> getFlag
|> fallback 42
```

library/Fission/App.hs
```diff
@@ -0,0 +1,25 @@
...
+import Fission.Environment (getFlagWithDefault)
+import Fission.Prelude
+
+-- | Run the given simple app.
+--   Set `DEBUG` to `True` to enable verbose app logging
+runApp :: RIO SimpleApp a -> IO a
+runApp simpleApp = do
+  isVerbose  <- isDebugEnabled
+  setRioVerbose isVerbose
+  runSimpleApp simpleApp
```
expede: 👍

PR #216, opened 2019-11-28
Move default fission peer to env var

### LINKED ISSUE #146: Move IPFS S3 node address to an env var
NB: Feature requests will only be considered if they solve a pain

# Summary
Move the `fission` const used for storing the address of our s3 node to an env variable. 

## Problem
* Currently if that address changes we will need to update it via a code deploy

### Impact
* We would remove config from code
* We would be closer to use a cluster of `IPFSS3Node`'s

## Solution
1. Move `fission` to `env.yaml` under `ipfs.peer_address`

bechurch:
## Summary
Moves our s3 remote peer address to our config file

## Closing issues

closes #146

## After Merge
* [ ] ENSURE YOU ADD `ipfs.remotePeer: /ip4/3.215.160.238/tcp/4001/ipfs/QmVLEz2SxoNiFnuyLpbXsH6SvjPTrHNMU88vCQZyhgBzgw` to production

### DIFF
--- a/app/Fission/Environment/IPFS/Types.hs
+++ b/app/Fission/Environment/IPFS/Types.hs
@@ -5,17 +5,19 @@ import           Fission.Prelude
 import qualified Fission.IPFS.Types as IPFS
 
 data Environment = Environment
-  { url     :: !IPFS.URL     -- ^ IPFS client URL (may be remote)
-  , timeout :: !IPFS.Timeout -- ^ IPFS timeout in seconds
-  , binPath :: !IPFS.BinPath -- ^ Path to local IPFS binary
-  , gateway :: !IPFS.Gateway -- ^ Domain Name of IPFS Gateway
+  { url        :: !IPFS.URL     -- ^ IPFS client URL (may be remote)
+  , timeout    :: !IPFS.Timeout -- ^ IPFS timeout in seconds
+  , binPath    :: !IPFS.BinPath -- ^ Path to local IPFS binary
+  , gateway    :: !IPFS.Gateway -- ^ Domain Name of IPFS Gateway
+  , remotePeer :: !IPFS.Peer    -- ^ Remote Peer to connect to
   } deriving Show
 
 instance FromJSON Environment where
   parseJSON = withObject "IPFS.Environment" \obj -> do
-    timeout <- obj .:? "timeout" .!= 3600
-    binPath <- obj .:? "binPath" .!= "/usr/local/bin/ipfs"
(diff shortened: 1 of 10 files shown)


app/Fission/Environment/IPFS/Types.hs
```diff
@@ -5,17 +5,19 @@ import           Fission.Prelude
...
   parseJSON = withObject "IPFS.Environment" \obj -> do
-    timeout <- obj .:? "timeout" .!= 3600
-    binPath <- obj .:? "binPath" .!= "/usr/local/bin/ipfs"
-    gateway <- obj .:? "gateway" .!= "ipfs.runfission.com"
-    url     <- obj .:  "url" >>= parseJSON . String
+    timeout    <- obj .:? "timeout" .!= 3600
+    binPath    <- obj .:? "binPath" .!= "/usr/local/bin/ipfs"
+    gateway    <- obj .:? "gateway" .!= "ipfs.runfission.com"
+    url        <- obj .:  "url" >>= parseJSON . String
+    remotePeer <- obj .:  "remotePeer"
```
dholms: should we default this to our peerID?

bechurch: So i'm not sure which way we want to fall on this.

Personally I'm uneasy about having default config values for things that could easily change (Like urls, ip addresses, certain types of keys). Simply because it lets us be less explicit in our configuration files. Which if these values change in the future we may be left having to read code to figure out how to override a default to get a production application back online.

@expede @dholms does that make sense?

dholms: well the .env file would take precedence over in code values. So we could default `remotePeer` to `blahblahblah` and it would still work fine as long as we have an actual value in our .env file. This would just be for if the .env file is screwed up for some reason and the `remotePeer` value gets left out. Altho that _should_ never happen so it probably isn't a big deal

bechurch: I think I'm going to leave it without a default. I would hope that we could always look at the `env.yaml` and know what the application is configured to use.

Its mostly to guard against the following cases:
1. you wanted to change the `remotePeer` so you added the `remotePeer` to `env.yaml` however you mispelled it as `remotPeer`. Everything would run fine and you wouldn't know that the config change didnt occur.
2. we build a test environment, including a test s3node. You forget to override in local development and now your pinning test files to s3 production.

Holding off on merging till after lunch in case theres an over my dead body :)

PR #220, opened 2019-11-30
Fix Swagger docs cURL generator & domain example

expede:
This PR introduces a handful of Swagger doc changes.

* Focus on production version
  * Remove HTTP from scheme drop-down
  * No longer append the port (assume 443)
* Generate valid `curl` commands
  * Fix hostname (had extra quotes around it before)
  * Fix docs on domain names
  * Add example to domain 

# Sane URLs

## Before

Screen Shot 2019-11-30 at 22 07 05

## After

Screen Shot 2019-11-30 at 21 56 22

# Fix Auto-generated Domain Docs

## Before

Screen Shot 2019-11-30 at 22 06 44

## After

Screen Shot 2019-11-30 at 21 56 28

### DIFF
--- a/library/Fission/AWS/DomainName/Types.hs
+++ b/library/Fission/AWS/DomainName/Types.hs
@@ -1,6 +1,6 @@
 module Fission.AWS.DomainName.Types (DomainName (..)) where
 
-import           Data.Swagger (ToSchema (..))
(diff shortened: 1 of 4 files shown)


expede:
Fixes #207

PR #221, opened 2019-12-02
Use Persistent DB Backend instead of Selda

icidasset:
This replaces the [Selda](https://www.selda.link/) library with the [Persistent](https://hackage.haskell.org/package/persistent) library.


library/Fission/User/Model
```diff
@@ -0,0 +1,13 @@
...
+  herokuAddOnId HerokuAddonId Maybe
+  role Role
+  secretDigest Text
+  username Text
+
+  insertedAt UTCTime
+  modifiedAt UTCTime
+
+  UniqueSecretDigest secretDigest
+  UniqueUsername username
```
expede: Oh nice! I didn't realize that you could do it this way 🎉 I'd only veer seen the "one big file" approach. Does this still enforce FKs and whatnot?

Also, how do you feel about calling these like `Model.entity`, or `Model.persistent`, or... something? Just feels weird without a file extension

library/Fission/User/Types.hs
```diff
@@ -1,40 +1,36 @@
...
+--
+-- instance Digestable (ID User) where
+--   digest = digest . UTF8.textShow
+--
+-- instance ToSchema (ID User) where
+--   declareNamedSchema _ =
+--     mempty
+--       |> type_ ?~ SwaggerInteger
+--       |> NamedSchema (Just "UserID")
+--       |> pure
```
expede: 🤔 Without having tried it, my assumption is to use the primary key:

```haskell
instance ToSchema (UserId) where
  declareNamedSchema _ = -- probably the same as before
```

library/Fission/User/Model
icidasset: It should enforce the FKs yeah. There's some links in the collapsed examples section in the [docs](https://www.stackage.org/haddock/lts-14.4/persistent-template-2.6.0/Database-Persist-TH.html#v:persistManyFileWith).

Will rename to `Model.entity` 👍

library/Fission/User/Types.hs
expede: If you want to go down the rabbit hole, we may be able to do this generically. From Persistent:

```haskell
class ( PersistField (Key record), ToJSON (Key record), FromJSON (Key record)
      , Show (Key record), Read (Key record), Eq (Key record), Ord (Key record))
  => PersistEntity record where
```

With that in mind:

```haskell
-- maybe need this constraint? No sure: PersistField (Key model) => ...
instance ToSchema (Key model) where
  declareNamedSchema proxy =
    mempty
      |> type_ ?~ SwaggerInteger
      |> NamedSchema (Just (recordName proxy <> "Id"))
      |> pure
```

Buuuuuuuuuuuuuut that may be more complicated 🤷‍♀️ It's flexible at least

DISCLAIMER: I haven't tried that code, so YMMV

library/Fission/Platform/Heroku/Types.hs
```diff
@@ -60,7 +62,8 @@ instance FromJSON Region where
     "amazon-web-services::ap-southeast-2" -> return Sydney
     "amazon-web-services::ap-northeast-1" -> return Tokyo
     "amazon-web-services::us-east-1"      -> return Virginia
-    bad -> fail <| "Invalid region: " <> show bad
+
+    bad -> fail ("Invalid region: " <> show bad)
```
expede: 👍

library/Fission/Prelude.hs
```diff
@@ -29,5 +30,8 @@ import RIO.List                    (headMaybe, intercalate)
 import RIO.Process
 import RIO.Time
 
+andThen :: (m a -> b) -> m a -> m b
+andThen = (<<=)
```
expede: Yeah, this isn't Elm. If anything, it should be called `bind`, but we may just want to use `>>=`. The term `andThen` has a lot of overlap with a BUNCH of other concepts, and bind doesn't _necessarily_ create asequence (it does often, but not always).

library/Fission/Web/Auth.hs
```diff
@@ -45,16 +46,38 @@ user = do
...
+    where_ (byUsername u)
+    return u
+  )
+    |> from
+    |> selectFirst
+    |> andThen (
+      maybe
+        (pure NoSuchUser)
+        checkPassword
+    )
```
expede: 💅 

```haskell
checkUser
  :: MonadDatabase
  => BasicAuthData
  -> RIO cfg (BasicAuthResult User)
checkUser (BasicAuthData authUsername authPassword) user = do
  result <- selectFirst <| from do
    where_ (byUsername user)
    return user

  maybe (pure NoSuchUser) checkpassword result
```

This reads more like English, more like SQL, 

I totally get the urge to turn everything into a pipeline, but I think that it actually hurts readability in this case. Don't be afraid to bind a variable!

---

Alternate, that's also quite readable IMO:

```haskell
checkUser (BasicAuthData authUsername authPassword) user = do
  result <- selectFirst <| from do
    where_ (byUsername user)
    return user

  case result of
    Nothing     -> pure NoSuchUser
    Just dbUser -> checkPasword dbUser
```

library/Fission/Prelude.hs
icidasset: @expede Gotcha 👍  So far example, you would do...

```
a
  |> b
  >>= c
  |> d
```

or:
1. Would you rather use `|> bind` in this scenario?
2. Or avoid this scenario all together and use a `do`, `let` or `where` block?

library/Fission/Web/Auth.hs
icidasset: Yeah you're right, I wrote it this way initially because I thought it'd be more like Ecto, in which the `where` would be in the pipeline instead of the `do` block. But that's not case, so will rewrite 👍

library/Fission/Prelude.hs
expede: In the specific case of the comment on `checkUser`, using a `do` block is clearer. `|>` and ``>>=` operate at different levels, so in the case that you _must_ pipeline, I'd use a `bind`, but also remember that you're likely going to get stuck inside that closure if you need to reuse that variable binding.

For science, the other other way to do it is to use a straight-up `fmap` instead of `bind` or `>>=`:

```haskell
checkUser
  :: MonadDatabase
  => BasicAuthData
  -> RIO cfg (BasicAuthResult User)
checkUser (BasicAuthData authUsername authPassword) =
  from (\user -> do
    where_ (byUsername user)
    return user)
  |> selectFirst
  |> fmap (maybe NoSuchUser checkPassword')

checkPassword' :: User -> BasicAuthResult User
checkPassword' usr =
  if validatePassword (encodeUtf8 <| secretDigest <| usr) password
     then Authorized usr
     else Unauthorized
```

icidasset: That looks clean af 👌 But yeah, I will stick with the previously suggested `selectFirst <| ...` format. Thanks for the science experiment 😄

icidasset:
Closing in favor of #226

PR #225, opened 2019-12-11
Add mock adapter for route53

bechurch:
## Summary
Adds the ability for devs to turn on a route53 mock adapater. This adapter outputs the requested changes to stdout and makes no request to external services.

The adapater can be enabled via the `aws.mock_enabled` field in the `env.yaml` file.

This should allow safer and simpler local development.

## After Merge


app/Fission/Internal/Development.hs
```diff
@@ -132,6 +134,7 @@ mkConfig dbPool processCtx httpManager logFunc = Config {..}
     awsSecretKey  = "SOME_AWS_SECRET_KEY"
     awsZoneID     = "SOME_AWS_ZONE_ID"
     awsDomainName = "SOME_AWS_DOMAIN_NAME"
+    awsMockEnabled = AWS.MockEnabled True
```
icidasset: This is mostly getting me familiar with the code, so excuse my ignorance 😅 

This means that the mock is enable by default in dev, right?
Do we even need an environment variable then, or we do we need mocking in non-test envs as well?
Or is the reasoning more like, why not have an env variable if it's easy to setup?

Also, thinking of testing, do we want mocking to be enabled per external service, or is like a "global" switch good enough?

expede: @icidasset you're absolutely right that this is a code smell 💯

The "right" way to do this is to write a `class MonadAWS`, and use that to do the switching — either with a flag or some `Maybe`s, but all isolated into the top-level `instance` declaration.

We're doing this for now just to get unblocked for local dev. I don't have the time this week to teach that style, so we're going to circle back. I've been eliminating `RIO` pretty aggressively in the DB adapter switch.

bechurch: so @icidasset this function here `mkConfig` is a helper for if you are trying to run code in the repl taht requires config.

So essentially this turns on mocking when you are tinkinging in the repl

I think we will want a per service mock for the most part as we expand? Given that I can rename this to `route53Mock`. what are your thoughts?

library/Fission/AWS/Types.hs
```diff
@@ -2,7 +2,14 @@
...
   ( DomainName (..)
   , ZoneID (..)
+  , MockEnabled (..)
   ) where
 
 import Fission.AWS.DomainName.Types
 import Fission.AWS.Zone.Types
+import Fission.Prelude
+
+newtype MockEnabled = MockEnabled Bool deriving (Show)-- newtype MockEnabled = Bool()
```
dholms: :nail_care: clear out this comment?

Also new line at the end of this file. I added a vscode extension for this cause I kept forgetting :laughing:

library/Fission/AWS/Route53.hs
```diff
@@ -43,12 +46,53 @@ registerDomain username (CID hash) = do
...
+    logDebug mockMessage
+    return (Right mockRecordResponse)
+  where
+    mockMessage = "MOCK: Updating DNS "
+                  <> displayShow recordType
+                  <> " record at: "
+                  <> displayShow domain
+                  <> " with "
+                  <> displayShow content
+    mockTime = parseISODateTime "2011-11-19 18:28:52.607875 UTC"
```
dholms: anything in particular about this time that we need in the mock?

Otherwise, we could use `getCurrentTime` from http://hackage.haskell.org/package/time-1.6.0.1/docs/Data-Time-Clock.html and save the import

app/Fission/Internal/Development.hs
icidasset: @bechurch Ah gotcha. For some reason I keep thinking that this `Development.hs` file is for the dev environment. Probably because I'm so used Phoenix's config files 🤔 Totally makes sense then 👍 

Yeah, per service mocks in dev makes sense (to enable mocking, or not, totally depends on the service I imagine). Also regarding testing, I guess in the tests we'll always have mocks on by default?

PR #226, opened 2019-12-11
Selda -> Persistent/Esqueleto

expede:
No more Selda. Now it's...

# Notes

* The line count change roughly `+1,300 −1,400` is heavily inflated by breaking up or renaming a couple modules, and switching to pipes in a few place.
* Esqueleto is an extension to Persistent that provides access to more complex queries
  * It's very, very common to use them together, and even in the Persistent docs
* (Many) more notes inline


app/Fission/Storage/PostgreSQL/Types.hs
```diff
@@ -0,0 +1,21 @@
+module Fission.Storage.PostgreSQL.Types (ConnectionInfo (..)) where
```
expede: Rename: `Fission.Storage.PostgreSQL.ConnectionInfo.Types`

library/Fission/Platform/Heroku/Types.hs
```diff
@@ -33,12 +33,11 @@ data Region
...
            , Eq
-           , Enum
            , Generic
-           , Bounded
-           , SqlType
            )
 
+derivePersistField "Region"
+
 instance ToJSON Region where
```
expede: Break apart

library/Fission/User/CID.hs
```diff
@@ -1,11 +1,7 @@
...
   ( module Fission.User.CID.Mutation
-  , module Fission.User.CID.Query
-  , module Fission.User.CID.Table
   , module Fission.User.CID.Types
   ) where
 
 import Fission.User.CID.Mutation
-import Fission.User.CID.Query
-import Fission.User.CID.Table
 import Fission.User.CID.Types
```
expede: This can also probably go 💀

library/Fission/User.hs
```diff
@@ -1,14 +1,10 @@
...
   , module Fission.User.Security
-  , module Fission.User.Table
   , module Fission.User.Types
   ) where
 
 import Fission.User.Mutate
-import Fission.User.Query
 import Fission.User.Security
-import Fission.User.Table
 import Fission.User.Types
```
expede: ...not long for this world

library/Fission/User/Types.hs
```diff
@@ -1,38 +1,18 @@
-module Fission.User.Types (User (..))where
+-- FIXME orphans out of the orphanage!
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
expede: Right, this too

expede:
Okay! That should do it 🎉 Adding some notes inline for reviewers. Things still left:

* [ ] Make sure that these changes actually work on the live data
* [ ] Check that all of the new queries behave as intended

app/Fission/Storage/PostgreSQL.hs
```diff
@@ -1,51 +1,36 @@
...
-  rawPool <- liftIO <| createPool (pgOpen pgInfo) seldaClose stripeCount connTTL connsPerStripe
-  logDebug <| "DB pool stats: " <> displayShow rawPool
+import           Fission.Models
 
-  return <| DB.Pool rawPool
+-- | Automagically update the database schema to reflect the latest schema in Fission.Models
+--
+--   Example:
+--
+--   > runOne (runDB updateDBToLatest)
```
expede: ☝️ How to apply the changes to a live system

app/Fission/Environment/Storage/Types.hs
```diff
@@ -1,13 +1,11 @@
...
-import Database.Selda.PostgreSQL
-
 import Fission.Prelude
-import Fission.Internal.Orphanage.PGConnectInfo ()
+import Fission.Storage.PostgreSQL.ConnectionInfo.Types
 
 -- | Configuration for the web application
 data Environment = Environment
-  { pgConnectInfo  :: !PGConnectInfo  -- ^ PostgreSQL configuration
+  { pgConnectInfo  :: !ConnectionInfo  -- ^ PostgreSQL configuration
```
expede: A custom equivalent to `PGConnectInfo`

Lifted verbatim from Steven's PR

app/Fission/Storage/PostgreSQL.hs
```diff
@@ -1,51 +1,36 @@
...
-connPool stripeCount connsPerStripe connTTL pgInfo@(PGConnectInfo {..}) = do
-  logDebug <| "Establishing DB pool for " <> displayShow pgDatabase
+import           Fission.Prelude
+import           Fission.Storage.PostgreSQL.ConnectionInfo.Types
 
-  rawPool <- liftIO <| createPool (pgOpen pgInfo) seldaClose stripeCount connTTL connsPerStripe
-  logDebug <| "DB pool stats: " <> displayShow rawPool
+import           Fission.Models
 
-  return <| DB.Pool rawPool
```
expede: These big chunks of removed code were Selda-specific

app/Main.hs
```diff
@@ -74,14 +77,17 @@ main = do
...
-             . condDebug
-             =<< Web.app
+
+      app <- Web.app
+
+      app
+        |> condDebug
+        |> CORS.middleware
+        |> runner settings
+        |> liftIO
```
expede: I think this reads better. Applying middleware and whatnot.

app/Main.hs
```diff
@@ -41,8 +46,8 @@ main = do
     Web.Environment     {..} = env |> web
     AWS.Environment     {..} = env |> aws
 
-    herokuID       = Hku.ID       <| encodeUtf8 (manifest |> Hku.id)
-    herokuPassword = Hku.Password <| encodeUtf8 (manifest |> Hku.api |> Hku.password)
+    herokuID       = Hku.ID       <| encodeUtf8 <| Hku.id <| manifest
+    herokuPassword = Hku.Password <| encodeUtf8 <| Hku.password <| Hku.api <| manifest
```
expede: Not trying to pun on dot syntax. The theory is that pipe directionality is actually a stronger factor in reading here.

app/Fission/Internal/Development.hs
```diff
@@ -168,5 +178,11 @@ mkConfig' = do
   let cfg = mkConfig dbPool processCtx httpManager logFunc
   return (cfg, close)
 
-pgConnectInfo :: PGConnectInfo
-pgConnectInfo = PGConnectInfo "localhost" 5432 "webapi" Nothing Nothing Nothing
+connectionInfo :: ConnectionInfo
+connectionInfo = ConnectionInfo
+  { pgDatabase = "web_api"
```
expede: Not sure when that got out of sync with what's in my `env.yaml`.

_Please let me know if this is wrong for some reason_

app/Fission/Internal/Orphanage/RIO.hs
```diff
@@ -71,3 +59,6 @@ instance
         |> mkClientEnv manager
         |> runClientM query
         |> liftIO
+
+instance MonadTime (RIO cfg) where
+  currentTime = liftIO getCurrentTime
```
expede: This entire module is going to get burned down in a future PR where we'll introduce the long awaited `newtype Fission a = Fission { unFission :: RIO Config a }`

app/Fission/Internal/Development.hs
expede: ACTUALLY, this function should probably read from the `env.yaml`

library/Fission/AWS/Route53.hs
```diff
@@ -20,9 +20,9 @@ import           Network.IPFS.Gateway.Types
 import           Network.IPFS.CID.Types
 
 registerDomain ::
-  ( MonadRIO           cfg m
+  ( MonadReader        cfg m
```
expede: What happened to `MonadRIO`? Well, we're moving to more generality, so there's far fewer instances when we need both of them together.

library/Fission/Config/Types.hs
```diff
@@ -3,16 +3,21 @@ module Fission.Config.Types (Config (..)) where
...
-import qualified Network.AWS.Auth              as AWS
-import qualified Fission.AWS.Types             as AWS
+import qualified Fission.Platform.Heroku.ID.Types       as Heroku
+import qualified Fission.Platform.Heroku.Password.Types as Heroku
+import qualified Fission.AWS.Types                      as AWS
+
 
 -- | The top level 'Fission' application 'RIO' configuration
-data Config dbBackend = Config
+data Config = Config
```
expede: No longer passing in a configurable backend, because the Persistent's `SqlBackend` will work with any SQL, not just PostgreSQL.

library/Fission/Internal/MonadDB/Class.hs
```diff
@@ -0,0 +1,19 @@
+{-# LANGUAGE UndecidableInstances #-}
```
expede: Unfortunately need to turn off the automatic decidablilty check. It can't mechanically tell that `MonadDB` has fewer members than `MonadIO`, but we know this is the case (because we're writing the `MonadDB`), so it's okay.

library/Fission/Internal/MonadDB/Types.hs
```diff
@@ -0,0 +1,6 @@
+module Fission.Internal.MonadDB.Types (Transaction) where
+
+import Database.Persist.Sql
+import RIO
+
+type Transaction m a = ReaderT SqlBackend m a
```
expede: A nice alias that Steven came up with 😃 👍

library/Fission/Internal/Orphanage/CID.hs
```diff
@@ -0,0 +1,8 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+module Fission.Internal.Orphanage.CID () where
+
+import Database.Persist.TH
+import Network.IPFS.CID.Types
+
+derivePersistField "CID"
```
expede: `derivePersistField` is a macro* that makes `CID` work directly in a DB table.

*Not technically a macro, but very similar idea

library/Fission/Internal/Orphanage/UUID.hs
```diff
@@ -4,12 +4,28 @@
...
+      |> UUID.fromString
+      |> \case
+            Just x  -> Right x
+            Nothing -> Left "Invalid UUID"
+
+  fromPersistValue _ =
+    Left "Not PersistDBSpecific"
+
+instance PersistFieldSql UUID where
+  sqlType _ = SqlOther "uuid"
```
expede: Selda wants `Enum` and `Bounded`; Persistent wants `PersistField` and `PersistFieldSql`. Honestly, I really like the Selda approach here, but _whatever_

library/Fission/Models.hs
```diff
@@ -0,0 +1,72 @@
...
+
+instance Digestable UserId where
+  digest = digest . UTF8.textShow
+
+instance ToSchema UserId where
+  declareNamedSchema _ =
+    mempty
+      |> type_ ?~ SwaggerInteger
+      |> NamedSchema (Just "UserID")
+      |> pure
```
expede: Welcome to the big monolithic models module! We get some niceties by having it set up this way, including automatic migration generation, and it looks like the docs. We always have the option to break them up later. There's some work going into making this module more modular as part of Persistent. Today it's totally doable, but also a bit of a PITA.

library/Fission/Models.hs
```diff
@@ -0,0 +1,72 @@
+{-# OPTIONS_GHC -fno-warn-missing-export-lists #-}
```
expede: Because so many functions and types are generated by template Haskell, this is just easier than trying to enumerate them all 🤷‍♀ There's a lot type-level trickery so it's just easier to trust in the DSL, IMO.

library/Fission/Models.hs
```diff
@@ -0,0 +1,72 @@
...
+
+{-# LANGUAGE NoDeriveAnyClass #-}
```
expede: Conflicted with the TH generator

library/Fission/Platform/Heroku/AddOn/Manifest/Types.hs
```diff
@@ -1,6 +1,7 @@
-module Fission.Platform.Heroku.AddOn.Manifest
+module Fission.Platform.Heroku.AddOn.Manifest.Types
   ( Manifest (..)
-  , API (..)
```
expede: `API` only lives inside of `Manifest`, so don't bother exporting it explicitly.

library/Fission/Prelude.hs
```diff
@@ -15,19 +17,68 @@ module Fission.Prelude
...
+logDebug = logWithoutLoc "" LevelDebug
+
+logWarn :: (ToLogStr msg, MonadLogger m) => msg -> m ()
+logWarn = logWithoutLoc "" LevelWarn
+
+logError :: (ToLogStr msg, MonadLogger m) => msg -> m ()
+logError = logWithoutLoc "" LevelError
+
+logOther :: (ToLogStr msg, MonadLogger m) => LogLevel -> msg -> m ()
+logOther = logWithoutLoc ""
```
expede: Making the `MonadLogger` class play nice with `RIO`'s concrete logger. Renaming so we don't have to make any changes to the code 👍

However, because of how `MonadLogger` works, you will sometimes have to tell these functions what kind of string you're giving it, because it's no longer going to force you to use `Utf8Builder` — but that's also really great in a lot of places! Way less `UTF8.textShow` and whatnot. (This is because it has that `ToLogStr` constraint rather than a concrete type).

library/Fission/User/CID/Mutation.hs
```diff
@@ -4,63 +4,64 @@ module Fission.User.CID.Mutation
...
-  mayID <- insertUnless Table.userCIDs (eqUserCID userID hash)
-            [UserCID def userID hash <@ now]
+  -> m (Maybe UserCIDId)
+create userId cid = runDBNow \now -> do
+  mayID <- insertUnique UserCID
+    { userCIDUserFK     = userId
+    , userCIDCid        = cid
+    , userCIDInsertedAt = now
+    , userCIDModifiedAt = now
+    }
```
expede: Figured that this would read better

expede: Also note that the fields are all now prefixed by the table's name. That's generated by Persistent.

library/Fission/User/CID/Mutation.hs
```diff
@@ -4,63 +4,64 @@ module Fission.User.CID.Mutation
...
-createX uID (fmap IPFS.CID.unaddress -> hashes) = do
-  results <- query do
-    uCIDs <- select Table.userCIDs `suchThat` inUserCIDs uID hashes
-    return <| uCIDs ! #cid
-
-  now <- liftIO getCurrentTime
+createX userId hashes = runDBNow \now -> do
+  existingCIDs <- select <| from \userCID -> do
+    where_ (userCID ^. UserCIDCid `in_` valList hashes)
+    return (userCID ^. UserCIDCid)
```
expede: Selectors are now data

library/Fission/User/Mutation/Error.hs
```diff
@@ -1,4 +1,4 @@
-module Fission.User.Mutate.Error (Create (..)) where
+module Fission.User.Mutation.Error (Create (..)) where
```
expede: Making these module name consistent

library/Fission/User/Role/Types.hs
```diff
@@ -0,0 +1,15 @@
+module Fission.User.Role.Types (Role (..)) where
```
expede: Broke into a `.Types` module. This has been bothering me for _weeks_

library/Fission/Web.hs
```diff
@@ -7,106 +7,108 @@ module Fission.Web
...
+  , Has IPFS.Gateway    cfg
+  , Has IPFS.Peer       cfg
+  , Has Web.Host        cfg
+  , Has AWS.AccessKey   cfg
+  , Has AWS.SecretKey   cfg
+  , Has AWS.ZoneID      cfg
+  , Has AWS.DomainName  cfg
+  , Has Heroku.ID       cfg
+  , Has Heroku.Password cfg
+  , HasLogFunc          cfg
```
expede: This list is currently _longer_. What the heck, right? Well, currently this function is contending with needing to describe a concrete `RIO` _and_ abstract capabilities like `MonadTime`. Hence some oddities like `MonadTime (RIO cfg)`.

library/Fission/Web/Auth.hs
```diff
@@ -1,30 +1,30 @@
...
-
-context :: Proxy (BasicAuthCheck User ': BasicAuthCheck ByteString ': '[])
-context = Proxy
+  -> (forall a . m a -> Handler a)
+  -> ServerT api m
+  -> ServerT api Handler
+server api = hoistServerWithContext api context
+  where
+    context :: Proxy Checks
+    context = Proxy
```
expede: I reworked this a bit, to make it work more nicely with the `Monad*` capability style.

library/Fission/Web/Auth.hs
```diff
@@ -1,30 +1,30 @@
...
+  ( Checks
+  , server
   , basic
   , user
   , checkUser
   ) where
 
 import Crypto.BCrypt
-import Database.Selda
+import Database.Esqueleto
```
expede: Protip: you generally only want to import `Database.Esqueleto`, since it re-exports a big chunk of `Database.Persist`.

library/Fission/Web/Auth/Types.hs
```diff
@@ -4,13 +4,14 @@ module Fission.Web.Auth.Types
...
+import Database.Esqueleto
 import Servant (BasicAuth)
 
 import Fission.Prelude
-import Fission.User.Types
+import Fission.Models
 
 -- | Authorization check for a normal user
-type ExistingUser = BasicAuth "existing user" User
+type ExistingUser = BasicAuth "existing user" (Entity User)
```
expede: `Entity` means "with the primary key, too". Persistent splits out the stuff that you can construct to insert a new row from the ID, and then just wraps existing rows in `Entity` to make space for the ID.

library/Fission/Web/DNS.hs
```diff
@@ -3,35 +3,36 @@ module Fission.Web.DNS
...
+  , MonadUnliftIO          m
+  , MonadThrow             m
   )
-  => User
-  -> RIOServer cfg API
-server User { username } cid = 
-  registerDomain username cid
-    >>= Web.Err.ensureM
+  => Entity User
+  -> ServerT API m
```
expede: Now that we're using the `Monad*` capabilities, we're not specialized to RIO. There's no reason for `RIOServer` to exist anymore, so we're back to the native `ServerT api m`.

For context: 

```haskell
newtype RIOServer cfg api = ServerT api (RIO cfg)
```

library/Fission/Web/Heroku.hs
```diff
@@ -1,152 +1,29 @@
...
-      logError <| "Unable to unpin CID: " <> display ipfsMsg
-      return ()
-
-    Right _ ->
-      return ()
-
-  return NoContent
+  => ServerT API m
+server = Provision.create
+    :<|> Deprovision.destroy
```
expede: Split these into their own files. They're the beefiest handlers in the project, and this file was getting long. Plus, it could be good to break out some handlers and whatnot if they continue to get more complex.

library/Fission/User/CID/Mutation.hs
icidasset: FYI, that was one of the reasons I did the separate models, so we didn't need to have the prefixed record fields (ie. you can disable the prefixes).

app/Fission/Storage/PostgreSQL.hs
expede: The same as would happen in any other system.

_However_ because persistent refuses to _remove_ tables/rows/etc, it actually may just be totally safe to stick right at the top of the server initialization 🤔 Good point

bechurch: Im more on the side of automatically migrating the database on deploy. If things are going wrong at this step for us I think thats more a sign thats something wrong with our steps leading up to deploy rather than our system configuration itself

Sidenote:
> persistent refuses to remove tables/rows/etc
Wait really? Whys that? I understand that might be beneficial to never have to worry about accidentally destroying data. However the trade offs dont see worth it

library/Fission/Web/Auth.hs
```diff
@@ -1,30 +1,30 @@
...
+  :: HasServer api Checks
   => Proxy api
-  -> cfg
-  -> RIOServer cfg api
-  -> Server api
-server api cfg = hoistServerWithContext api context (toHandler cfg)
-
-context :: Proxy (BasicAuthCheck User ': BasicAuthCheck ByteString ': '[])
-context = Proxy
+  -> (forall a . m a -> Handler a)
```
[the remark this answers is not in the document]
expede: Same note?

We actually do need this type to be existential:

```haskell
    • Couldn't match type ‘m a -> Handler a’
                     with ‘forall x. m0 x -> n0 x’
      Expected type: (m a -> Handler a)
                     -> ServerT api m -> ServerT api Handler
        Actual type: (forall x. m0 x -> n0 x)
                     -> ServerT api m0 -> ServerT api n0
```

library/Fission/Web/Auth.hs
```diff
@@ -35,22 +35,48 @@ basic unOK pwOK = BasicAuthCheck (pure . check')
...
+checkUser ::
+  ( MonadDB     m
+  , MonadLogger m
+  )
+  => BasicAuthData
+  -> m (BasicAuthResult (Entity User))
 checkUser (BasicAuthData username password) = do
-  mayUser <- findOne <| select User.users `suchThat` User.byUsername (decodeUtf8Lenient username)
-  maybe (pure NoSuchUser) checkPassword mayUser
+  users <- runDB <| select <| from \usr -> do
```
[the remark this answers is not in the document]
expede: Yup; will do 👍

library/Fission/Web/Auth.hs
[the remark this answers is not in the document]
icidasset: Yeah, I had a similar comment here: https://github.com/fission-suite/web-api/pull/226#discussion_r358257850

I wondered if it might be useful to note down, in a code comment, why we have the `forall`.

app/Main.hs
bechurch: Right pipes ftw

app/Fission/Storage/PostgreSQL.hs
expede: The idea is that `migarteAll` is _so_ automatic that it's easy to accidentally bork critical production server data. If you want to drop a table, rename a column, &c, then you need to move to a more manual process like [`persistent-migration`](https://hackage.haskell.org/package/persistent-migration)

library/Fission/Web/Auth.hs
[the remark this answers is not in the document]
expede: Sure, I'll throw a comment in here 👍  It's a standard part of the type system, but it's not in a lot of place in the code, so makes sense.

app/Fission/Storage/PostgreSQL.hs
bechurch: > migarteAll is so extremely automatic

I gotcha, that makes sense then. Its not obvious that if you remove a a field from a type that it will remove any related column.

Whats your thoughts on `persistent-migration` then? something we will add?

What are your thoughts on

library/Fission/Web/Heroku/Deprovision.hs
```diff
@@ -0,0 +1,96 @@
...
+  addOnIds <- runDB <| select <| from \herokuAddOn -> do
+    where_ (herokuAddOn ^. HerokuAddOnUuid ==. val uuid')
+    limit 1
+    return herokuAddOn
+
+  Entity addOnId _ <- ensureOne addOnIds
+
+  -- users associated with those add ons
+
+  userIds <- runDB <| select <| from \user -> do
```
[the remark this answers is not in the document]
expede: Discussed in Discord. We're going to leave these for now, and play with Esqueleto's slightly different `selectFirst` experience the next time we need to write a query like this.

app/Fission/Storage/PostgreSQL/ConnectionInfo/Types.hs
```diff
@@ -0,0 +1,36 @@
...
+  , pgPort     :: Int
+  , pgUsername :: Maybe Text
+  , pgPassword :: Maybe Text
+  }
+
+instance Show ConnectionInfo where
+  show ConnectionInfo {..} =
+    mconcat
+      [ "host="    <> Text.unpack pgHost
+      , " port="   <> show pgPort
```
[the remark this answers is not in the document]
expede: Okay, yeah, that makes sense 👍

expede:
Needs a rebase, obvs. But I want to get any substantive changes in before doing that.

expede:
Saving position before epic rebase

Removing debugging code
23303920b0cd99819492076234c58dd755ca0be0

expede:
Alright! And that's a completed rebase 🎉

@bechurch & @dholms I'd love your reviews when you get a sec

app/Fission/Internal/Development.hs
```diff
@@ -31,14 +39,17 @@ import           Fission.App (runApp)
...
+  logOptions <- logOptionsHandle stdout True
+  processCtx  <- mkDefaultProcessContext
+  httpManager <- HTTP.newManager HTTP.defaultManagerSettings
+
   withLogFunc (setLogUseTime True logOptions) \logFunc -> do
-    dbPool      <- runApp <| connPool 1 1 3600 pgConnectInfo
-    processCtx  <- mkDefaultProcessContext
-    httpManager <- HTTP.newManager HTTP.defaultManagerSettings
-    run logFunc dbPool processCtx httpManager action
+    withDBPool logFunc connectionInfo 4 \dbPool ->
```
bechurch: 💅 Not sure what `4` is in this context. Could be useful if we move it to a named variable.

app/Fission/Web/Log/Sentry.hs
```diff
@@ -61,9 +62,9 @@ logger minRIOLogLevel sentryService _cs logSource logLevel msg =
...
 loggerName = "Fission.Web.Log.Sentry"
 
-fromRIOLogLevel :: LogLevel -> SentryLevel
+fromRIOLogLevel :: RIO.LogLevel -> SentryLevel
 fromRIOLogLevel = \case
-  LevelWarn      -> Sentry.Warning
-  LevelError     -> Sentry.Error
-  LevelOther msg -> Sentry.Custom <| Text.unpack msg
-  _              -> Sentry.Info
+  RIO.LevelWarn      -> Sentry.Warning
```
bechurch: 👍 onprefixingg `RIO.`

app/Fission/Internal/Development.hs
bechurch: 💅 Not sure what `4` is in this context. Could be helpful if we move it to a named variable. Same with below on line `174`.

library/Fission/Internal/MonadDB.hs
```diff
@@ -0,0 +1,38 @@
...
+import Fission.Internal.MonadDB.Types
+import Fission.Internal.MonadDB.Class
+
+runDBNow :: (MonadTime m, MonadDB m) => (UTCTime -> Transaction m a) -> m a
+runDBNow timeTransaction = do
+  now <- currentTime
+  runDB (timeTransaction now)
+
+ensureOneId :: (Exception err, MonadLogger m, MonadThrow m) => err -> [Entity a] -> m (Key a)
+ensureOneId err entities = pure . entityKey =<< ensureOne err entities
```
bechurch: 💅 It seems like more often than not we use `return` instead of `pure` maybe useful to just commit to one.

library/Fission/Internal/Orphanage/UUID.hs
```diff
@@ -4,12 +4,28 @@
...
+import           Database.Persist.Sql
 
-import Fission.Prelude
+import           Fission.Prelude
 
-instance Enum UUID
+instance PersistField UUID where
+  toPersistValue uuid =
+    uuid
+      |> UUID.toString
```
bechurch: Loving the pipes 😍

library/Fission/Platform/Heroku/Provision/Request/Types.hs
```diff
@@ -0,0 +1,70 @@
...
+import           Data.Swagger hiding (name)
+
+import           Fission.Prelude
+import qualified Fission.Plan.Types                   as Plan
+import qualified Fission.Platform.Heroku.Region.Types as Heroku
+
+data Request = Request
+  { callbackUrl :: Text          -- ^ The URL which should be used to retrieve updated information about the add-on and the app which owns it.
+  , name        :: Text          -- ^ Logical name of the resource being provisioned.
+  -- , oauthGrant :: Maybe Text -- OAuthGrant -- ^ OAuth object details (nullable).
```
bechurch: Whys this commented out?

library/Fission/Random.hs
```diff
@@ -12,13 +12,19 @@ import           Data.Word8
...
+    |> fmap (decodeUtf8Lenient . BS.filter isAsciiAlphaNum)
 
--- | Generate random AlphaNumericSymbol 'Text' with a given length
+-- | Generate random alphanumeric symbol @Text@ with a given length
 alphaNumSymbol :: MonadIO m => Natural -> m Text
-alphaNumSymbol len = decodeUtf8Lenient . BS.filter URL.isValidURLCharacter <$> bsRandomLength len
+alphaNumSymbol len =
+  len
+    |> bsRandomLength
+    |> fmap (decodeUtf8Lenient . BS.filter URL.isValidURLCharacter)
```
bechurch: Much better!

library/Fission/AWS/Route53.hs
dholms: At one point, I thought `MonadRIO` was the single most important type in our codebase. Its gifts were boundless. You won't be forgotten _sheds single tear_

library/Fission/Config/Types.hs
dholms: aw yiss

library/Fission/Internal/MonadDB.hs
dholms: actually asked @expede about this same thing a bit ago:
>Yeah, they're the same
I tend to only use return in do blocks, because it's a pun on imperative programming
And pure is just "bring a pure value into this effectful scope"
Which helps me personally signal which notation I mean to be in
But they're totally interchangeable

library/Fission/User/CID/Mutation.hs
dholms: reads so nice w/ the record syntax & `runDBNow`

app/Fission/Internal/Development.hs
expede: Fair point. This is why newtypes were invented 👍.

library/Fission/Internal/MonadDB.hs
expede: I mean, we _can_ standardize on `pure`. But that needs to be a team decision.

Thoughts?

@dholms @icidasset @bechurch

library/Fission/Platform/Heroku/Provision/Request/Types.hs
expede: Well, we're not doing anything with the grant, it's _technically_ optional. We do want to start using it in the near term, but the blocker has been this PR. It's a whole other data structure, which isn't really that much work to add. The commented out key is mainly a reminder that it lives there. I'll open an Issue for that 👍

expede: For reference, here's an example provision request payload from the Heroku docs:

```
{
  "options": {},
  "oauth_grant": {
    "code": "01234567-89ab-cdef-0123-456789abcdef",
    "type": "authorization_code",
    "expires_at": "2016-03-03T18:01:31-0800"
  },
  "plan": "basic",
  "region": "amazon-web-services::us-east-1",
  "uuid": "01234567-89ab-cdef-0123-456789abcdef"
}
```

library/Fission/User/Security.hs
```diff
@@ -3,18 +3,24 @@ module Fission.User.Security
...
-import           Fission.User.Types
 
--- | Create a 'SecretDigest' from the users ID
---   Barely an obsfucating technique, but enough to hide DB ordering
-hashID :: ID User -> SecretDigest
-hashID = Text.take 20 . digest
+-- | Generate an ID
+genID :: MonadIO m => m SecretDigest
+genID =
+  50
```
dholms: such a small gripe, but I feel like the first line of this would read better as `Random.alphaNum 50`

library/Fission/Web.hs
```diff
@@ -7,108 +7,110 @@ module Fission.Web
...
-     )
-  => RIO cfg (Context '[BasicAuthCheck User, BasicAuthCheck ByteString])
+mkAuth ::
+  ( Has Heroku.ID       cfg
+  , Has Heroku.Password cfg
+  , HasLogFunc          cfg
+  , MonadReader         cfg m
+  , MonadDB        (RIO cfg)
+  )
+  => m (Context Auth.Checks)
```
dholms: `Auth.Checks` is so much better here

library/Fission/Internal/MonadDB.hs
icidasset: I personally prefer `return`, because it makes more sense to me, eg. you `return` something at the end of a `do` block. Also `pure` feels a bit weird to me because it's the inverse of what functions are usually named. That is, with `pure` the function is named after the function argument. Normally, I think, functions are named after the result of a function, for example `sum` or `add`.

That said, this is totally personal, I wouldn't mind either.

bechurch:
Additional note on migrating
image
The alter table commands it outputs will fail. This is because this doesnt take into account the data migration required between renamed columns

library/Fission/Internal/MonadDB.hs
expede: To be clear, I have no strong preference here. Whatever as long as we all agree. However, at least some segment of the community has been slowly moving away from `return`. IMO it makes sense in `do` notation (i.e. where I tend to use it). The argument against is that it means something very different from `return` in any other language. For instance:

```haskell
do
  foo <- runSomething
  bar <- return otherThing
  baz <- return evenMore
  return (foo + bar + baz)
```

You'd normally expect `return` to exit out of a function early.

We can't totally get rid of `pure`, since it's a required method for `Applicative`.

But yeah, sure, we can do `return` everywhere 👍

library/Fission/Web.hs
expede: Right?!

library/Fission/User/Security.hs
expede: Option 1
```haskell
50
  |> Random.alphaNum
  |> fmap hashID
  |> liftIO
```

Option 2
```haskell
liftIO <| hashID <$> Random.alphaNum 50
```

Option 3
```haskell
do
  rando <- Random.alphaNum 50
  liftIO <| return <| hashID rando
```

Option 4
```haskell
do
  rando <- Random.alphaNum 50
  rando
    |> hashID
    |> return
    |> liftIO
```

Option 5
```haskell
liftIO . return . hashID =<< Random.alphaNum 50
```

Thoughts, @dholms ?

dholms: ok yeah I like option 1 & 2 the best out of those.

And the reason for not doing this?
```
  Random.alphaNum 50
    |> fmap hashID
    |> liftIO
```

Is it just against convention to put the first function before the pipes?
My brain interprets pipes as "take this value and apply these transformations to it" and `50` seems so arbitrary that at first glance it doesn't make much sense. `Random.alphaNum 50` gives way more context to what the value being transformed actually is

[ReasonML](https://reasonml.github.io/docs/en/pipe-first) doesn't turn the innermost function into a pipe :stuck_out_tongue_winking_eye:

app/Fission/Storage/PostgreSQL.hs
expede: Yeah, exactly. Not something that we need to day, and we can add if/when we start doing a lot of destructive DB changes.

library/Fission/User/Security.hs
expede: I personally like Options 2, but that's a decade for idiomatic Haskell rearing its head 😅

> Is it just against convention to put the first function before the pipes?
Yeah, this is OCaml/Reason vs Elixir convention 😜

The idea is that since we're relying on a directional metaphor, we should stick to one direction. This is all convention, and Haskell doesn't normally have pipes... so we need to decide what we want to set as a style here.

I used to teach Elixir, so "go with the flow" (direction) is pretty drilled into me, but whatever we can change. I can see an argument especially with nested constructors.

Thoughts @fission-suite/development?

expede: Oops, @icidasset wasn't on that list, so pinging separately

expede:
@bechurch let's just write some SQL for now 👍 My guess is that it'll be messy no matter what 🤷‍♀

expede:
>  in the future if someone hasn't touched web-api in awhile and they have to go tracking down every SQL command 

I'm pretty sure that it's just the 4 of us. I mean, sure toss it in an Issue, but I'll be pleasantly surprised if that turns out to bite someone not on this team 😉

PR #228, opened 2019-12-12
JWT Public Key Auth (via higher order auth)

dholms:
## Problem
We want to switch cli authentication over to private keys. This is more secure than basic auth, keeps us from storing auth details in .fission.yaml, and is a good "testing ground" for our ideas around identity since we own the entire flow

Authentication is currently done through a BasicAuth strategy w/ username and password.

## Solution
Add an additonal JWT auth option that authenticates users through a jwt signed by their private key.

Prerequisite for https://github.com/fission-suite/cli/issues/63
https://github.com/fission-suite/cli/pull/70


library/Fission/Internal/Crypto.hs
```diff
@@ -0,0 +1,37 @@
...
+base64ToEdPubKey = Ed.publicKey . base64ToBytes 
+
+base64ToBytes :: ByteString -> BA.ScrubbedBytes
+base64ToBytes = pack . Base64.decodeLenient
+
+toBase64 :: BA.ByteArrayAccess a => a -> ByteString
+toBase64 = Base64.encode . unpack
+
+edSignature :: BA.ByteArrayAccess ba => ba -> CryptoFailable Ed.Signature
+edSignature = Ed.signature
```
dholms: This is because `signature` is a reserved keyword in Haskell and will screw up syntax highlighting for the rest of the page. Any other suggestions to get around this are appreciated :+1:

library/Fission/Internal/Orphanage/Signature.hs
```diff
@@ -0,0 +1,18 @@
...
+
+import Fission.Prelude
+
+import Servant
+import Servant.Swagger
+import Data.Swagger
+
+import Fission.Web.Auth.Types
+
+-- @@TODO: add an accurate security definition
```
dholms: The only [SecuritySchemeTypes](https://hackage.haskell.org/package/swagger2-2.5/docs/Data-Swagger-Internal.html#t:SecuritySchemeType) :thinking: 

Could use a hand on this^^

library/Fission/Web/Auth.hs
```diff
@@ -54,3 +71,33 @@ checkUser (BasicAuthData username password) = do
...
+  )
+  =>  HasLogFunc cfg
+  => Request
+  -> RIO cfg User
+jwtHandler req =
+  JWT.validateReq req >>= \case
+    Left err -> throwM err
+    Right did -> do
+      logDebug <| "DID: " <> displayShow did
+      -- @@TODO: this part just filler until we update DB:
```
dholms: As the comment says, we'll want to actually look up the user w/ their DID. But we need DB stuff settled before doing this

library/Fission/Internal/Crypto.hs
expede: You can also use the prime symbol (`signature'`), but I actually am not against this naming! It's very clear

library/Fission/Internal/Orphanage/Signature.hs
icidasset: Should be something along the lines of:

```haskell
let
  apiKey = ApiKeyParams "Authorization" ApiKeyHeader
in
...
|> securityDefinitions .~ [( "apiKey", SecurityScheme (SecuritySchemeApiKey apiKey) Nothing )]
```

Is this what you could use help with, or was it something else?

dholms: Yeah our jwt just isn't really an api key. It's more like a transaction signature. But there's no Swagger type for that. ApiKey is probably the closest, so maybe it's best to just use that

dholms:
Checkpointing at 249dc1406f8f3e956f0faf006adfc643b44729c4 before rebase :scream_cat:

app/Fission/Internal/Development.hs
```diff
@@ -31,14 +39,17 @@ import           Fission.App (runApp)
...
 -}
-runOne :: RIO (Config PG) a -> IO a
+runOne :: RIO Config a -> IO a
 runOne action = do
-  logOptions   <- logOptionsHandle stdout True
+  logOptions <- logOptionsHandle stdout True
+  processCtx  <- mkDefaultProcessContext
+  httpManager <- HTTP.newManager HTTP.defaultManagerSettings
+
   withLogFunc (setLogUseTime True logOptions) \logFunc -> do
```
expede: Looks like we can drop the `do` here as well

app/Fission/Internal/Development.hs
```diff
@@ -31,14 +39,17 @@ import           Fission.App (runApp)
      > runOne Network.IPFS.Peer.all
      > -- Right ["/ip4/3.215.160.238/tcp/4001/ipfs/QmVLEz2SxoNiFnuyLpbXsH6SvjPTrHNMU88vCQZyhgBzgw"]
 -}
-runOne :: RIO (Config PG) a -> IO a
+runOne :: RIO Config a -> IO a
 runOne action = do
-  logOptions   <- logOptionsHandle stdout True
+  logOptions <- logOptionsHandle stdout True
+  processCtx  <- mkDefaultProcessContext
+  httpManager <- HTTP.newManager HTTP.defaultManagerSettings
```
expede: Wait, why did these get moved up here?

dholms: don't think I touched these. Just haven't rebased yet

dholms:
quick checkpoint before massive rebase: 4c564d3ba309215c9be60f4f181476c780b352d9

library/Fission/User/Mutation.hs
```diff
@@ -47,29 +49,25 @@ create ::
...
-            { userUsername      = username
-            , userEmail         = email
-            , userRole          = Regular
-            , userActive        = True
-            , userHerokuAddOnId = herokuAddOnId
-            , userSecretDigest  = secretDigest
-            , userInsertedAt    = now
-            , userModifiedAt    = now
-            }
+create username did email herokuAddOnId now = do
```
dholms: hmmm what do we do for heroku accounts? keep basic auth for now? Generate semi-permanent jwts instead? Just thinking out loud here

library/Fission/Internal/Orphanage/ByteString.hs
```diff
@@ -0,0 +1,11 @@
...
+
+module Fission.Internal.Orphanage.ByteString () where
+
+import Fission.Prelude
+
+instance FromJSON ByteString where
+  parseJSON = withText "ByteString" (pure . encodeUtf8)
+
+instance ToJSON ByteString where
+  toJSON = String . decodeUtf8Lenient
```
expede: Where do we want these? There's a good chance that we actually want a `newtype` wrapper for the thing.

library/Fission/Internal/Orphanage/HigherOrder.hs
```diff
@@ -0,0 +1,18 @@
...
+
+import Fission.Prelude
+
+import Servant
+import Servant.Swagger
+import Data.Swagger
+
+import Fission.Web.Auth.Types
+
+-- @@TODO: add an accurate security definition
```
expede: ???

library/Fission/Internal/Orphanage/ByteString.hs
dholms: They're used in `Fission.Web.Auth.JWT.Types` for serializing/deserializing JWTs

library/Fission/Internal/Orphanage/HigherOrder.hs
dholms: Ah right. This is for Swagger. The problem is, none of the [Security Scheme](https://hackage.haskell.org/package/swagger2-2.5/docs/Data-Swagger.html#t:SecuritySchemeType) options match what we're doing. I guess the closest is ApiKey? And then we can describe it more in depth in the description

library/Fission/Internal/Orphanage/RegisterDid.hs
```diff
@@ -0,0 +1,18 @@
...
+import Fission.Prelude
+
+import Servant
+import Servant.Swagger
+import Data.Swagger
+
+import Fission.Web.Auth.Types
+
+-- @@TODO: add an accurate security definition
+instance HasSwagger api => HasSwagger (RegisterDid :> api) where
```
expede: This probably doesn't need to be an orphan instance 🤔 We have `InstanceSigs` turned on in #252 that should probably help with this. Ping me for more info.

library/Fission/Time.hs
```diff
@@ -1,13 +1,19 @@
...
 -- | The Doherty theshold
 doherty :: NominalDiffTime
 doherty = 0.4
 
 dohertyMicroSeconds :: Int
 dohertyMicroSeconds = 400000
+
+getCurrentPOSIXTime :: MonadIO m => m Integer
+getCurrentPOSIXTime =
+  getCurrentTime >>= return . round . utcTimeToPOSIXSeconds
```
expede: Why not use `MonadTime`?

library/Fission/User/Email/Types.hs
```diff
@@ -0,0 +1,31 @@
...
+  parseJSON = withText "Email" \txt -> return (Email txt)
+
+instance MimeRender PlainText Email where
+  mimeRender _ = UTF8.textToLazyBS . unEmail
+
+instance MimeUnrender PlainText Email where
+  mimeUnrender _proxy bs =
+    case bs |> Lazy.toStrict |> decodeUtf8' of
+      Left err  -> Left  <| show err
+      Right txt -> Right <| Email txt
```
expede: Totally fine as-is, but can also be phrased as:

```haskell
instance MimeUnrender PlainText Email where
  mimeUnrender _proxy bs =
    bs
      |> Lazy.toStrict
      |> decodeUtf8'
      |> \case
        Left err  -> Left  <| show err
        Right txt -> Right <| Email txt
```

or

```haskell
instance MimeUnrender PlainText Email where
  mimeUnrender _proxy bs =
    bs
      |> Lazy.toStrict
      |> decodeUtf8'
      |> bimap show Email
```

library/Fission/User/Mutation.hs
expede: Yeah, let's keep basic auth for them as an option for now. We'll switch them to keys in the future, but we need to figure out exactly how we want to do that. There's a few options, such as asking them for a key in the CLI command, but that's less convenient than generating one for them which is less secure. TL;DR leave it as a password for now :P

library/Fission/User/Registration/Types.hs
```diff
@@ -6,44 +6,38 @@ import Fission.Prelude
...
 
     mempty
       |> type_      ?~ SwaggerObject
       |> properties .~
            [ ("username", username')
-           , ("password", password')
            , ("email", email')
            ]
-      |> required .~ ["username", "password"]
+      |> required .~ ["username", "email"]
```
expede: 👍

library/Fission/User/Username/Types.hs
```diff
@@ -0,0 +1,47 @@
...
+  deriving          ( Eq
+                    , Generic
+                    , Show
+                    )
+  deriving anyclass ( ToParamSchema )
+  deriving newtype  ( IsString
+                    , ToHttpApiData
+                    )
+
+  -- deriving anyclass ( ToSchema )
```
expede: 🔪

library/Fission/Web/Auth/Basic.hs
```diff
@@ -0,0 +1,80 @@
...
+
+    attemptMsg :: ByteString
+    attemptMsg = "Unauthorized user! Attempted with username: " <> username
+
+parseBasic :: Auth.BasicToken -> Either Auth.Error BasicAuthData
+parseBasic (Auth.BasicToken token) = do
+  let parts = Ch.split ':' <| Base64.decodeLenient token  
+  case parts of
+    [un,pw] -> Right <| BasicAuthData un pw
+    _ -> Left Auth.BadToken
```
expede: Hmm yeah, this is going to be interesting to merge with the changes #252 😅 I'll look at that

library/Fission/Web/Auth/Error.hs
```diff
@@ -0,0 +1,21 @@
...
+           , Generic
+           , ToJSON
+           )
+
+instance Show Error where
+  show = \case
+    NoToken      -> "No token included on request"
+    BadToken     -> "Token is improperly formatted"
+    NoSuchUser   -> "No such user exists"
+    Unauthorized -> "User not authorized"
```
expede: Why `Show` instead of `Display`?

library/Fission/Web/Auth/JWT.hs
```diff
@@ -0,0 +1,124 @@
...
+validateHeader bytes = 
+  case decodePart bytes of
+    Left err -> Left err
+    Right header -> 
+      if typ header /= "JWT"
+      then Left BadHeader
+      else
+        if alg header /= "Ed25519"
+        then Left UnsupportedAlg
+        else Right ()
```
expede:
```haskell
Right header ->
  case (typ header, alg header) of
    ("JWT", "Ed25519") -> Right ()
    ("JWT", _) -> Left UnsupportedAlg
    _  -> Left BadHeader
```

library/Fission/Web/Auth/JWT.hs
```diff
@@ -0,0 +1,124 @@
...
+
+validateTime :: MonadIO m => JWT.Payload -> m (Either JWT.Error ())
+validateTime pl = do
+  time <- getCurrentPOSIXTime
+  if time > JWT.exp pl
+  then return <| Left JWT.Expired
+  else
+    if time < JWT.nbf pl
+    then return <| Left JWT.TooEarly
+    else return <| Right ()
```
expede: Same as above

library/Fission/Web/Auth/JWT.hs
```diff
@@ -0,0 +1,124 @@
...
+    Just pubkey -> Right pubkey
+
+validateSig :: ByteString -> ByteString -> ByteString -> Either JWT.Error ()
+validateSig content pubkey64 sig64 = 
+  case Crypto.base64ToEdPubKey pubkey64 of
+    CryptoFailed _ -> Left BadPublicKey
+    CryptoPassed pk ->
+      case Crypto.base64ToSignature sig64 of
+        CryptoFailed _ -> Left BadSignature
+        CryptoPassed sig -> do
```
expede: Don't need the `do`

library/Fission/Web/Auth/JWT.hs
```diff
@@ -0,0 +1,124 @@
...
+          case Ed.verify pk (Crypto.pack content) sig of
+            False -> Left IncorrectSignature
+            True -> Right ()
+
+decodePart :: FromJSON a => ByteString -> Either JWT.Error a
+decodePart bytes = bytes 
+  |> Base64.decodeLenient
+  |> BS.Lazy.fromStrict
+  |> decode
+  |> maybe (Left ParseError) Right
```
expede: Really nit-picky and optional, but as we use more and more of pipes, I think that this is the most consistent:

```haskell
decodePart bytes = 
  bytes 
    |> Base64.decodeLenient
    |> BS.Lazy.fromStrict
    |> decode
    |> maybe (Left ParseError) Right
```

library/Fission/Web/Auth/JWT/Types.hs
```diff
@@ -0,0 +1,34 @@
+module Fission.Web.Auth.JWT.Types
+  ( Token (..)
+  , Header (..)
+  , Payload (..)
+  -- , ReqInfo (..)
```
expede: 🔪

library/Fission/Web/Auth/JWT/Types.hs
```diff
@@ -0,0 +1,34 @@
...
+instance ToJSON Header
+
+data Payload = Payload
+  { iss        :: !ByteString
+  , nbf        :: !Integer
+  , exp        :: !Integer
+  } deriving (Show, Generic)
+
+instance FromJSON Payload
+instance ToJSON Payload
```
expede: Standalone instances are fine, but we can also do:

```haskell
data Payload = Payload
  { iss        :: !ByteString
  , nbf        :: !Integer
  , exp        :: !Integer
  } deriving ( Show
             , Generic
             , ToJSON -- This one
             , FromJSON -- and this one
             )
```

library/Fission/Web/Auth/Types.hs
```diff
@@ -1,17 +1,36 @@
...
 
--- | Authorization check for a normal user
-type ExistingUser = BasicAuth "existing user" (Entity User)
-
 -- | Authorization check for the Heroku Addon API
 type HerokuAddOnAPI = BasicAuth "heroku add-on api" ByteString
+
+-- | Authorization check to return encoded did for registering new users
+type RegisterDid = AuthProtect "register-did"
+type instance AuthServerData (AuthProtect "register-did") = ByteString
```
expede: 🤔 Have you considered a `newtype` for the DID auth? It'll likely save us from orphans now or down the road.

library/Fission/Web/Auth/Types.hs
```diff
@@ -1,17 +1,36 @@
...
+type HigherOrder = AuthProtect "higher-order"
+type instance AuthServerData (AuthProtect "higher-order") = Entity User
+
+data Token
+  = Basic BasicToken
+  | Bearer BearerToken
+  | None
+
+newtype BasicToken = BasicToken { unBasic :: ByteString }
+newtype BearerToken = BearerToken { unBearer :: ByteString }
```
expede: 👍

library/Fission/Web/Middleware/AuthBody.hs
```diff
@@ -0,0 +1,45 @@
...
+addBodyHash :: ByteString -> Vault.Key ByteString -> Vault.Vault -> Vault.Vault
+addBodyHash body vaultKey oldVault = 
+    if BS.null body
+    then oldVault
+    else Vault.insert vaultKey hash oldVault
+  where 
+    hash = 
+      body 
+      |> digest
+      |> encodeUtf8
```
expede: Various minor alignment things here. This will not play well with do notation, so probably good to get into the habit:

```haskell
addBodyHash body vaultKey oldVault = 
  if BS.null body
    then oldVault
    else Vault.insert vaultKey hash oldVault
  where 
    hash = 
      body 
        |> digest
        |> encodeUtf8
```

expede:
@dholms our PRs touch some simialr functionality. Totes on me; I'll fix it if we merge my PR first!

library/Fission/Internal/Orphanage/ByteString.hs
dholms: yeah I can see that we might want `newtype`s for some of these. especially a `DID` type

library/Fission/Time.hs
dholms: good call :+1:

library/Fission/Web/Auth/Error.hs
dholms: tbh haven't really figured out the difference. Why `Display` instead of `Show`?

library/Fission/Web/Auth/JWT.hs
dholms: :heart_eyes_cat:

library/Fission/Web/Auth/JWT.hs
dholms: yeah I hadn't really thought to combine multiple checks in the same `case` like this. waay cleaner

library/Fission/Web/Auth/Types.hs
dholms: I think I would just do `DID` because those will be used throughout our system and not always in the context of auth
```haskell
newtype DID  = DID { unwrap :: ByteString }

type instance AuthServerData (AuthProtect "register-did") = DID
```

library/Fission/User/DID/Types.hs
```diff
@@ -0,0 +1,15 @@
...
+import Fission.Prelude
+
+import Data.Swagger
+import Fission.Internal.Orphanage.ByteString ()
+
+newtype DID = DID { unDID :: ByteString }
+  deriving          ( Eq
+                    , Generic
+                    , Show
+                    , ToJSON
```
dholms: Getting this warning:
```
Both DeriveAnyClass and GeneralizedNewtypeDeriving are enabled
      Defaulting to the DeriveAnyClass strategy for instantiating ToJSON
    • In the newtype declaration for ‘DID’
```
Shouldn't cause an issue, but is there a way to suppress this warning?

library/Fission/User/Mutation.hs
```diff
@@ -36,30 +39,31 @@ createWithHeroku herokuUUID herokuRegion username password = runDBNow \now -> do
...
   -> Transaction m (Either Error.Create UserId)
-create username password email herokuAddOnId now =
-  hashPassword' password >>= \case
-    Left err ->
-      return (Left err)
-
-    Right secretDigest -> do
-      let newUserRecord = User
-            { userUsername      = username
+create username didOrPW email herokuAddOnId now =
```
dholms: so after adding a `newtype DID`, I realized that I was improperly calling this function from `createWithHeroku` (I was passing in the password as a did :grimacing:)

Go types! :tada:

Anyway, when creating a user, you always need either a DID or a password. I tried to represent that in the function signature. But I can't tell if it's a smart move or too roundabout/confusing. Thoughts @expede?

library/Fission/User/Creator/Class.hs
```diff
@@ -65,8 +69,18 @@ instance MonadIO m => Creator (Transaction m) where
 
           Nothing ->
             return <| Left <| openUnionLift User.AlreadyExists
+    where
+      -- parsePW :: MonadIO m => Either DID Text -> m (Either Password.FailedDigest (Maybe Text))
```
dholms: was getting a horrible ridiculous no good compiler where when I had the type declaration in
```
    • Occurs check: cannot construct the infinite type:
        m ~ ReaderT SqlBackend m
      Expected type: Transaction m (Either Errors UserId)
        Actual type: m (Either
                          (OpenUnion
                             '[User.AlreadyExists, Heroku.AddOn.Error, Password.FailedDigest])
                          (Key User))
........
 ```
which I wasn't able to figure out :confused:  code seems to work fine so I have no clue what's causing it

File: library/Fission/User/Creator/Class.hs [hunk 28]
```diff
@@ -42,19 +43,22 @@ class Heroku.AddOn.Creator m => Creator m where
     -> m (Either Errors UserId)
 
 instance MonadIO m => Creator (Transaction m) where
-  create username password email herokuAddOnId now =
-    Password.hashPassword password >>= \case
+  create username didOrPW email herokuAddOnId now =
```
dholms: Just cause I think my other comment got loss:
Still not sure if this is smart or a ridiculous :thinking: 
The thinking is, anytime a user is created, they will _always_ have _either_ a did _or_ a password

library/Fission/User/DID/Types.hs
expede: You should specify how you want these derived on the newtype (i.e. use the underlying type’s instance, or treat it as a `data`). So, likely:

```haskell
newtype DID = DID { unDID :: Text }
  deriving anyclass (Eq, Show, Generic)
  deriving newtype (ToJSON)
```

library/Fission/User/Modifier/Class.hs
```diff
@@ -18,8 +20,15 @@ instance MonadIO m => Modifier (Transaction m) where
...
+          [ UserSecretDigest =. Just secretDigest
           , UserModifiedAt   =. now
           ]
 
         return . Right <| Password password
+
+  updateDID userID (DID did) now = do
+    update userID
+      [ UserDid =. Just did
+      , UserModifiedAt   =. now
```
expede: Kinda weird alignment :P

library/Fission/User/Username/Types.hs
```diff
@@ -0,0 +1,45 @@
...
+import Servant
+
+import qualified RIO.ByteString.Lazy as Lazy
+import qualified Network.IPFS.Internal.UTF8 as UTF8
+
+newtype Username = Username { username :: Text }
+  deriving          ( Eq
+                    , Generic
+                    , Show
+                    )
```
expede: You should push these into one of the explicit deriving strategies below. Probably anyclass.

library/Fission/Web/Auth.hs
```diff
@@ -3,31 +3,51 @@ module Fission.Web.Auth
...
 mkAuth ::
-  ( MonadAuth (Entity User) m
-  , MonadAuth Heroku.Auth   m
+  ( MonadAuth (AuthHandler Request DID)           m
+  , MonadAuth (AuthHandler Request (Entity User)) m
+  , MonadAuth (BasicAuthCheck Heroku.Auth)        m
   )
   => m (Context Checks)
 mkAuth = do
+  didAuth    <- verify
```
expede: Nice

expede: Actually, while we’re in here. Something that I’ve found confusing in the test branch is the name `verify`, since it’s not actually running that action; it’s returning a thing that _can_ verify. So why not rename the function to `verifier` or `getVerifier`? Thoughts? Tackle in this PR?

library/Fission/Web/Auth.hs
```diff
@@ -45,3 +66,21 @@ basic unOK pwOK = BasicAuthCheck (return . check)
...
+  , MonadDB        t m
+  , User.Retriever t
+  )
+  => Request
+  -> m (Entity User)
+handler req =
+  case getToken req of
+    Auth.None -> throwM Auth.NoToken
+    Auth.Basic basic' -> Basic.handler basic'
+    Auth.Bearer bearer -> JWT.handler bearer
```
expede: Nice!

library/Fission/Web/Auth/Basic.hs
```diff
@@ -0,0 +1,82 @@
...
+          logWarn attemptMsg
+          return <| Left Auth.Unauthorized
+
+    attemptMsg :: ByteString
+    attemptMsg = "Unauthorized user! Attempted with username: " <> username
+
+parseBasic :: Auth.BasicToken -> Either Auth.Error BasicAuthData
+parseBasic (Auth.BasicToken token) = do
+  let parts = Ch.split ':' <| Base64.decodeLenient token  
+  case parts of
```
expede: Why mind it to `parts` and then case analyze rather than just piping through to `\case` or just putting the expression in a `case ... of`?

library/Fission/Web/Auth/DID.hs
```diff
@@ -0,0 +1,33 @@
...
+  => Request
+  -> m DID
+handler req = 
+  case getToken req of
+    Auth.Bearer token -> 
+      JWT.validateJWT token >>= \case
+        Left err -> do
+          logWarn <| "Failed registration with token " <> Auth.unBearer token
+          throwM err
+        Right did' -> return did'
```
expede: Not required, but can be a bit more declarative with:

```haskell
Auth.Bearer token@(Auth bearer) -> 
  ...
  logWarn <| “Failed registration with token “ <> bearer
  ...
```

library/Fission/Web/Auth/Error.hs
expede: `Display` is way more efficient than `Show`, and works really nicely with the `log*` functions. The problem with `Show` is that it generates a `String` (i.e. `[Char]`), which is a simple data structure, but allocates a bunch of thunks (memory), is O(n) to traverse, and appending is O(n+m). `Display` uses a text builder, which is much more efficient in memory and compute.

You can get both pretty easily:

```haskell
instance Display Error where
  display = \case
    ...

instance Show Error where
  show = show . display
```

library/Fission/Web/Auth/JWT.hs
```diff
@@ -0,0 +1,112 @@
...
+    content = headerRaw <> "." <> payloadRaw
+    did = iss payload
+  pubkey64 <- DID.toPubkey did
+  _ <- validateSig content pubkey64 sig64
+  Right payload
+
+getParts :: ByteString -> Either JWT.Error (ByteString, ByteString, ByteString)
+getParts token = do
+  let parts = Ch.split '.' token
+  case parts of
```
expede: Why the `let`?

library/Fission/Web/Auth/JWT.hs
```diff
@@ -0,0 +1,112 @@
...
+        _                  -> Left BadHeader
+
+validateTime :: MonadTime m => JWT.Payload -> m (Either JWT.Error ())
+validateTime pl = do
+  time <- getCurrentPOSIXTime
+  return <| 
+    case (time > JWT.exp pl, time < JWT.nbf pl) of
+      (True, _) -> Left JWT.Expired
+      (_, True) -> Left JWT.TooEarly
+      _         -> Right ()
```
expede: Awyiss

library/Fission/Web/Auth/Utils.hs
```diff
@@ -0,0 +1,17 @@
...
+
+getToken :: Request -> Auth.Token
+getToken req = 
+  case lookup "Authorization" (requestHeaders req) of
+    Nothing -> Auth.None
+    Just token -> case BS.stripPrefix "Basic " token of 
+      Just basic' -> Auth.Basic <| Auth.BasicToken basic'
+      Nothing -> case BS.stripPrefix "Bearer " token of
+        Just bearer -> Auth.Bearer <| Auth.BearerToken bearer
+        Nothing -> Auth.None
```
expede: 👍

library/Fission/Web/Swagger/Auth.hs
```diff
@@ -0,0 +1,13 @@
...
+import Data.Swagger
+
+fissionSecurity :: SecurityScheme
+fissionSecurity = SecurityScheme (SecuritySchemeApiKey keyParams) describe
+  where
+    keyParams :: ApiKeyParams
+    keyParams = ApiKeyParams "Authorization" ApiKeyHeader
+
+    describe :: Maybe Text
+    describe = Just "A higher-order auth type that accepts both basic auth and fission-formatted jwts as described here: https://whitepaper.fission.codes/identity/authentication"
```
expede: Nitpick: capitialization

library/Fission/Web/User/Email.hs
```diff
@@ -0,0 +1,35 @@
...
+  => ServerT API m
+server (Username username) = do
+  mayUser <- runDB <| User.getByUsername username
+
+  case mayUser of
+    Nothing -> throwM err404
+    Just (Entity _ user) -> do
+      case userEmail user of
+        Nothing -> return <| Email ""
+        Just email -> return <| Email email
```
expede: No need for this `do`.

Also:

```haskell
return . Email <| maybe “” <| userEmail user
```

or 

```haskell
Just (Entity _ User { email = Just email’ }) -> return <| Email email
_ -> return <| Email “”

File: library/Fission/Web/Auth/Error.hs (hunk 15 above)
Comment (dholms) [reply library/Fission/Web/Auth/Error.hs]: oh sweet makes sense :+1:

File: library/Fission/User/Username/Types.hs (hunk 30 above)
Comment (dholms) [reply library/Fission/User/Username/Types.hs]: Not really sure the difference between these :thinking: definitely something I need to look into
The best way to get the compiler to stop complaining seemed to be
```haskell
newtype Username = Username { username :: Text }
  deriving          ( Generic )
  deriving anyclass ( ToParamSchema )
  deriving newtype  ( Show
                    , Eq
                    , IsString
                    , ToHttpApiData
                    )
```

library/Fission/Web/Auth.hs
dholms: yup I like it :+1:

library/Fission/User/Creator/Class.hs
expede: Hmmm yeah. It really doesn’t help us when we what to say, know who has what type of auth (to tell to upgrade), and could potentially lead to confusing edge cases. We want to be able to make solid assumptions about the data in our application that’s passed validation.

How would you feel about changing this class to have 2 methods: creatWithPassword and createWithDID? We can later extend that if we add more types of keys.

expede: Also createWithDID doesn’t need to know about Heroku, which is a plus IMO 😛

library/Fission/Web/Auth/JWT.hs
expede: ACTUALLY now that I think about this again, we could use a multi-way if

 ```haskell
if | time > JWT.exp pl -> ...
   | time < JWT.nbf pl -> ...
   | otherwise -> Right ()
```

But not a huge issue either way

expede:
I'm having trouble with my `env.yaml` since I set up my machine fresh after getting the new logic board / SSD combo. Just waiting to test this locally. Seems to be working, but I'm getting what looks like a misconfigured remote IPFS.

PR #229, opened 2019-12-13
69 alphanumber password

### LINKED ISSUE #69: Explore: wrap server in `transaction`?
Something like what Phoenix does, rather than at each call site? Downside is that it'll apparently roll back even if handled internally.

bechurch:
## Summary
Our heroku register path was broken in two ways
1. The password and username returned has characters that were invalid in an ENV variable as is
2. We returned a nonsense username.

This should fix both of these by:
1. Restricting username and password generation to be alphanumeric
2. Use the heroku UUID as the username
3. Return the appropriate username to heroku

I also took the time to:
1. Remove duplicate url character check code
2. Update random generators to take a fixed length

## Closing issues
closes #69 

## After Merge

### DIFF
--- a/library/Fission/Internal/URL.hs
+++ b/library/Fission/Internal/URL.hs
@@ -8,11 +8,13 @@ import Data.Word8
 import Fission.Internal.Bool (anyX)
 import Fission.Prelude
 
+-- | Check that a byte represents a valid URL character
 isURL :: Word8 -> Bool
 isURL = anyX (isAlpha : isDigit : isSpecial)
   where
     isSpecial = (==) <$> specials
 
+-- | List of URL special character checks
 specials :: [Word8]
 specials =
     [ _asterisk
--- a/library/Fission/Internal/URL.hs
+++ b/library/Fission/Internal/URL.hs
@@ -1,5 +1,5 @@
 module Fission.Internal.URL
-  ( isURL
+  ( isValidURLCharacter
   , specials
   ) where
 
@@ -9,8 +9,8 @@ import Fission.Internal.Bool (anyX)
 import Fission.Prelude
 
 -- | Check that a byte represents a valid URL character
-isURL :: Word8 -> Bool
-isURL = anyX (isAlpha : isDigit : isSpecial)
+isValidURLCharacter :: Word8 -> Bool
+isValidURLCharacter = anyX (isAlpha : isDigit : isSpecial)
   where
     isSpecial = (==) <$> specials
 
--- a/library/Fission/Random.hs
+++ b/library/Fission/Random.hs
@@ -26,6 +26,7 @@ bsRandomLength len =
     len
     |> (* 100)
     |> BS.random
+    |> liftIO
     |> fmap (BS.take toTake)
   where
     toTake = fromIntegral len
--- a/library/Fission/Random.hs
+++ b/library/Fission/Random.hs
@@ -28,5 +28,5 @@ bsRandomLength len = BS.take (naturalToInt len) <$> BS.random (len * 100)
 -- | Check if a given character is AlphaNumeric
 isAsciiAlphaNum :: Word8 -> Bool
 isAsciiAlphaNum w = isAsciiUpper w
-                || isAsciiLower w
-                || isDigit w
+                 || isAsciiLower w
+                 || isDigit w
--- a/library/Fission/Random.hs
+++ b/library/Fission/Random.hs
@@ -1,27 +1,25 @@
 -- | Random values
 module Fission.Random
-  ( alphaNumString
-  , byteString
-  -- , text
+  ( alphaNum
+  , alphaNumSymbol
   ) where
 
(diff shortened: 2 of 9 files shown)


library/Fission/Internal/URL.hs
```diff
@@ -8,11 +8,13 @@ import Data.Word8
 import Fission.Internal.Bool (anyX)
 import Fission.Prelude
 
-isURL :: Word8 -> Bool
-isURL = anyX (isAlpha : isDigit : isSpecial)
+-- | Check that a byte represents a valid URL character
+isValidURLCharacter :: Word8 -> Bool
+isValidURLCharacter = anyX (isAlpha : isDigit : isSpecial)
```
icidasset: Just made the same change in my `tests` branch 😅 👍

library/Fission/Random.hs
```diff
@@ -1,42 +1,32 @@
...
+  , alphaNumSymbol
   ) where
 
 import qualified Data.ByteString.Random as BS
-import           Data.Word8
 import qualified RIO.ByteString         as BS
+import qualified Fission.Internal.URL   as URL
+import           Data.Text              as T
+import           Data.Word8
+import           GHC.Natural
```
icidasset: Maybe use `Numeric.Natural` instead of the internal `GHC.Natural` module (just as the ghc module suggest).

expede: Or just `RIO`, no?

library/Fission/Random.hs
```diff
@@ -1,42 +1,32 @@
...
 
--- | Check that a byte represents a valid URL character
-isURL :: Word8 -> Bool
-isURL w = isAsciiUpper w
-        || isAsciiLower w
-        || isDigit w
-        || anyX urlSpecials w
+-- | Generate random 'ByteString' with a given length
+bsRandomLength :: Natural -> IO ByteString
+bsRandomLength len = BS.take (naturalToInt len) <$> BS.random (len * 100)
```
expede: You may be able to use `fromIntegral`, since `Natural` is an instance of `Num`, and do away with the additional module import

library/Fission/Random.hs
```diff
@@ -1,42 +1,32 @@
...
 
-import Fission.Internal.Bool (anyX)
 import Fission.Prelude
 
--- | Generate random 'Text'
-text :: Natural -> IO Text
-text amount = decodeUtf8Lenient <$> byteString amount
+-- | Generate random AlphaNumeric 'Text' with a given length
+alphaNum :: Natural -> IO Text
+alphaNum len = decodeUtf8Lenient . BS.filter isAsciiAlphaNum <$> bsRandomLength len
```
expede: While we're in here, let's get rid of the `IO` specialization

```haskell
alphaNum :: MonadIO m => Natural -> m Text
alphaNumSymbol :: MoandIO m => Natural -> m Text
```

And then we won't need to wrap this in `liftIO` at every call site

library/Fission/Web/Heroku.hs
```diff
@@ -75,8 +76,8 @@ provision Request {uuid, region} = do
                        logError <| displayShow err
                        return []
 
-  username <- liftIO <| User.genID
-  secret   <- liftIO <| Random.text 200
+  let username = T.pack . UUID.toString <| uuid
```
expede: One character simpler:

```haskell
T.pack <| UUID.toString uuid
```

expede: Now that this is pure, we can probably just stick it inline (line 99). Thoughts?

library/Fission/Random.hs
icidasset: Yeah, it's in `RIO` as well it seems: https://www.stackage.org/haddock/lts-14.4/rio-0.1.12.0/RIO-Prelude-Types.html#t:Natural

library/Fission/Random.hs
expede: Why `* 100`? Why not just pas that up to the caller?

expede: Anyhow, now that we're all pipey:

```haskell
len
  |> (* 100)
  |> BS.random
  |> fmap (BS.take toTake)
where
  toTake = toIntegral len
```

bechurch: Oooh! I prefer the pipes 🚰 😍

library/Fission/Web/Heroku.hs
bechurch: Ohh I like the way that reads! Will change.

> Now that this is pure, we can probably just stick it inline (line 99). Thoughts?

We also use it on line 82.

library/Fission/Random.hs
bechurch: @expede It didnt like this change 
image
Any ideas why?

expede: You need to `liftIO`

bechurch: 🤦‍♂

PR #231, opened 2019-12-14
Improve IPFS Error message for our API consumers

bechurch:
##  Summary
image
In cases where we have a user readable error message from IPFS we should return it to the user so that they can debug appropriately.

_Mostly this was a fun exercise learning more Haskell_

related: https://github.com/fission-suite/ipfs-haskell/pull/9
## After Merge


bechurch:
Just waiting on merging until the new MonadLogger is merged in from https://github.com/fission-suite/web-api/pull/226

PR #233, opened 2019-12-18
Update Register to throw a 409 on unique conflict

bechurch:
## Summary
Currently we throw a 500 if a users attempts to sign up with the same email / username. 
This is incorrect and we should throw a 409.

## After Merge

### DIFF
--- a/library/Fission/User/Mutation.hs
+++ b/library/Fission/User/Mutation.hs
@@ -22,6 +22,7 @@ import qualified Fission.User.Password.Types as User
 createWithHeroku ::
   ( MonadDB     m
   , MonadTime   m
+  , MonadLogger   m
   )
   => UUID
   -> Heroku.Region
@@ -41,20 +42,24 @@ createWithHeroku herokuUUID herokuRegion username password = runDBNow \now -> do
 -- | Create a new, timestamped entry with optional heroku add-on
 create ::
   MonadIO m
+  => MonadLogger m
   => Text
   -> Text
   -> Maybe Text
   -> Maybe HerokuAddOnId
   -> UTCTime
   -> Transaction m (Either Error.Create UserId)
-create username password email herokuAddOnId now =
+create username password email herokuAddOnId now = do
+  logError <| show "ASDASDASD"
+
   hashPassword' password >>= \case
     Left err ->
       return (Left err)
 
     Right secretDigest -> do
-      userID <- insert User
-        { userUsername      = username
+      logError <| show "Here"
+      let newUserRecord = User {
+          userUsername      = username
         , userEmail         = email
         , userRole          = Regular
         , userActive        = True
@@ -64,7 +69,15 @@ create username password email herokuAddOnId now =
         , userModifiedAt    = now
         }
 
-      return (Right userID)
+      insertUnique newUserRecord >>= \case
+        Just userID -> do
+          logError <| show "got"
+
+          return <| Left Error.AlreadyExists
+          -- return (Right userID)
+        Nothing -> do
+          logError <| show "not"
+          return <| Left Error.AlreadyExists
 
 updatePassword ::
   ( MonadDB     m
--- a/library/Fission/User/Mutation.hs
+++ b/library/Fission/User/Mutation.hs
@@ -53,16 +53,16 @@ create username password email herokuAddOnId now =
       return (Left err)
 
     Right secretDigest -> do
-      let newUserRecord = User {
-          userUsername      = username
-        , userEmail         = email
-        , userRole          = Regular
-        , userActive        = True
-        , userHerokuAddOnId = herokuAddOnId
-        , userSecretDigest  = secretDigest
-        , userInsertedAt    = now
-        , userModifiedAt    = now
-        }
+      let newUserRecord = User
+            { userUsername      = username
+            , userEmail         = email
+            , userRole          = Regular
+            , userActive        = True
+            , userHerokuAddOnId = herokuAddOnId
+            , userSecretDigest  = secretDigest
+            , userInsertedAt    = now
+            , userModifiedAt    = now
+            }
 
       insertUnique newUserRecord >>= \case
         Just userID ->
--- a/library/Fission/User/Mutation.hs
+++ b/library/Fission/User/Mutation.hs
@@ -22,7 +22,6 @@ import qualified Fission.User.Password.Types as User
 createWithHeroku ::
   ( MonadDB     m
   , MonadTime   m
-  , MonadLogger   m
   )
   => UUID
   -> Heroku.Region
@@ -42,22 +41,18 @@ createWithHeroku herokuUUID herokuRegion username password = runDBNow \now -> do
 -- | Create a new, timestamped entry with optional heroku add-on
 create ::
   MonadIO m
-  => MonadLogger m
   => Text
   -> Text
   -> Maybe Text
   -> Maybe HerokuAddOnId
   -> UTCTime
   -> Transaction m (Either Error.Create UserId)
-create username password email herokuAddOnId now = do
-  logError <| show "ASDASDASD"
-
+create username password email herokuAddOnId now =
   hashPassword' password >>= \case
     Left err ->
       return (Left err)
 
     Right secretDigest -> do
-      logError <| show "Here"
       let newUserRecord = User {
           userUsername      = username
         , userEmail         = email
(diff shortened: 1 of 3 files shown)


library/Fission/User/Mutation/Error.hs
```diff
@@ -6,6 +6,8 @@ import Fission.Prelude
 import Fission.Web.Error
 
 data Create = FailedDigest
+            | AlreadyExists
+
```
expede: 👍

library/Fission/User/Mutation/Error.hs
```diff
@@ -15,6 +17,8 @@ data Create = FailedDigest
 
 instance Display Create where
   display FailedDigest = "Could not create password digest"
+  display AlreadyExists = "The username or email already exists in our system"
 
 instance ToServerError Create where
   toServerError FailedDigest = err500 { errBody = "Could not create password digest" }
+  toServerError AlreadyExists = err409 { errBody = "The username or email already exists in our system" }
```
expede: Could reuse the `Display` message for consistency:

```haskell
toServerError AlreadyExists = err409 { errBody = display AlreadyExists }
```

bechurch: Oh I am so glad you had this recommendation. It was irking me that we had duplicated messages!

library/Fission/Web/User/Create.hs
```diff
@@ -37,7 +37,8 @@ server
      )
   => ServerT API m
 server (User.Registration username password email) = do
-  void . runDBNow <| User.create username password email Nothing
+  createResponse <- runDBNow <| User.create username password email Nothing
+  void . Web.Err.ensure <| createResponse
```
expede: Can save a character if we treat `<|` as `$`.

```haskell
void <| Web.Err.ensure createResponse
```

Any thoughts? This is pretty un-Elixir in convention, but we need to figure out pipe conventions in Haskell as a team IMO

library/Fission/User/Mutation.hs
```diff
@@ -53,8 +53,8 @@ create username password email herokuAddOnId now =
       return (Left err)
 
     Right secretDigest -> do
-      userID <- insert User
-        { userUsername      = username
+      let newUserRecord = User {
+          userUsername      = username
```
expede: The convention in Haskell is "comma-leading style"

```haskell
let newUserRecord = User
      { userUsername      = username
      , userEmail         = email
      -- ...
      }
```

library/Fission/User/Mutation.hs
```diff
@@ -64,7 +64,12 @@ create username password email herokuAddOnId now =
         , userModifiedAt    = now
         }
 
-      return (Right userID)
+      insertUnique newUserRecord >>= \case
+        Just userID ->
+          return (Right userID)
+
+        Nothing ->
+          return (Left Error.AlreadyExists)
```
expede: I like this phrasing! It's very clear. Just for science, here's another phrasing:

```haskell
insertUnique newUserRecord >>= return . maybe (Left Error.AlreadyExists) Right
```

🤔

Hmm, yeah, I like the `case` 👍

library/Fission/User/Mutation/Error.hs
expede: Happy to help 🙌

Something that may be of interest — because Haskell is pure:
* you can see a function as "some data with holes in it" rather than a machine routine
* function definitions are literally a type of definitional equality

So if:

```haskell
display Foo = "my awesome message"
```

...then...

```haskell
display Foo == "my awesome message"
```

in 100% of cases. It's a simple case, but the same goes for all of your functions, even ones with `throw` and whatnot, because underneath that's all modeled in data (just hidden by a monadic interface to make it feel more imperative)

library/Fission/Web/User/Create.hs
bechurch: I think haskell has a few to many operators so Im all for using a pipe instead since its more widely used.

I think dot syntax if your creating a new function for general use

```haskell
funcZ = funcX . funcY
```

expede: Re-reading your response, I think you may have misunderstood what I wrote:

```haskell
void . Web.Err.ensure <| createResponse
void <| Web.Err.ensure createResponse
```

It's not about eliminating the `.`. It's about not needing to delineate every word on a line 😉

Still going to put my initial response below, because some still applies. It'll be clear that I misread what you meant _here_ about "too many operators" 😅

---

# Pipe Count

Interesting — that's certainly an alchemist's view on pipes 😉 I don't mean that in a disparaging way; just genuinely interesting to see the influence.

Clearly, I'm still reading `<|` as `$`. In idiomatic Haskell, this reads like nonsense...

```haskell
foo $ bar
foo <| bar
```

...because `$` is "put everything on the right in brackets"...

```haskell
foo (bar)
```

...which written that way looks absurd.

`$` can also creates nice lines of delineation of values vs functions, something that we're losing here (or at least trading off).

The pipe has _such a strong metaphor_ that I can see the argument. It's a bit noisier, but seems to be the way that people would think about the pipe.

## More Widely Used

> Im all for using a pipe instead since its more widely used.

Is it, though? Right-pipes are nearer at hand for you, but that's very much the influence of Elixir. They have different conventions in Elm and the OCaml/Reason/F# family, where they're generally used as I presented above.

Given that they're totally foreign to Haskell, we get to decide on how we want to use them, but they're neither common to most programmers nor standard in use across languages.

# Operators

## Too Many

> I think haskell has a few to many operators

Sure, but you also haven't written much idiomatic Haskell. `.` and `$` are _by far_ the most common operators in the language, so the argument about there being too many doesn't totally apply. Remember that JS has about 2 dozen, and they're overloaded to do all kinds of stuff. You just don't see them as operators because you learned with them in place. Yes, Haskell can suffer from operator-itus, but certainly not here where I'm asking if you'd like to remove one.

Our code base has:

* `|>` application
* `.` composition
* `\` lambda
* `=` assignment
* `<$>` map
* `<|>` alternate
* `::` type
* `^.` accessor
* `.~` set
* `?~` set maybe
* Inequalities (`==` and the like)

TS has
* `()` application (usually)
* `.` application
* `.` accessor by string key
* `=>` lambda
* `[]` accessor by non-string key
* `{}` closure (usually)
* `=` assignment
* `:` type 
* Inequalities (`==` and the like)

Yes, in Haskell you can define operators, but mostly they're the same ones reused. Most languages have a lot of different kinds of syntax, but they're all in the C-family, so they feel familiar and seem to "disappear" into the langauge.

## Point Free Style

Writing functions without explicit arguments is called the point free style:

```haskell
funcZ = funcX . funcY
```

Sometimes it's very useful, and it's something that I desperately miss when writing Elixir (they're so close!) There are a lot of cases where this is useful, but I'm slowly moving towards the part of the community that says that explicit arguments are actually better on named functions...

```haskell
funcZ arg = arg |> funcY |> funcX
```

...and anonymous functions can be phrased in a nice way by reducing noise...

```haskell
[1,2,3]
  |> fmap (\x -> upcase (show x))
  |> -- ...

[1,2,3]
  |> fmap (upcase . show)
  |> -- ...
```

# Prime

This is all good stuff adding to thinking about what should go into Prime 🎉

library/Fission/User/Mutation/Error.hs
```diff
@@ -15,6 +18,8 @@ data Create = FailedDigest
 
 instance Display Create where
   display FailedDigest = "Could not create password digest"
+  display AlreadyExists = "The username or email already exists in our system"
 
 instance ToServerError Create where
-  toServerError FailedDigest = err500 { errBody = "Could not create password digest" }
+  toServerError FailedDigest = err500 { errBody = UTF8.showLazyBS <| FailedDigest }
+  toServerError AlreadyExists = err409 { errBody = UTF8.showLazyBS <| AlreadyExists }
```
expede: Yeah, just like in Elixir, let's not over-pipe We're just calling a single function on lines 24 & 25

library/Fission/Web/User/Create.hs
bechurch: > > Im all for using a pipe instead since its more widely used.
> 
> Is it, though? Right-pipes are nearer at hand for you, but that's very much the influence of Elixir. 
Oh I had meant in our project specifically!

bechurch: > but I'm slowly moving towards the part of the community that says that explicit arguments are actually better on named functions

I really agree on this point!

library/Fission/User/Mutation/Error.hs
bechurch: yup good call

PR #236, opened 2019-12-20
Make email required for register

bechurch:
## Summary
Ensure email is required only for users who sign up via the register api. Heroku still does not require an email

## After Merge

### DIFF
--- a/library/Fission/User/Registration/Types.hs
+++ b/library/Fission/User/Registration/Types.hs
@@ -7,21 +7,21 @@ import Fission.Prelude
 data Registration = Registration
   { username :: !Text
   , password :: !Text
-  , email    :: !(Maybe Text)
+  , email    :: !Text
   }
 
 instance ToJSON Registration where
-  toJSON (Registration username' password' maybeEmail) =
+  toJSON (Registration username' password' email') =
     Object [ ("username", String username')
            , ("password", String password')
-           , ("email", maybe Null String maybeEmail)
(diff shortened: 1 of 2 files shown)


expede:
I would love to have Heroku also provide us with email. That must be possible, but I haven't found it yet.

library/Fission/User/Registration/Types.hs
```diff
@@ -7,21 +7,21 @@ import Fission.Prelude
...
   }
 
 instance ToJSON Registration where
-  toJSON (Registration username' password' maybeEmail) =
+  toJSON (Registration username' password' email') =
     Object [ ("username", String username')
            , ("password", String password')
-           , ("email", maybe Null String maybeEmail)
+           , ("email", String email')
            ]
```
expede: 🤔 We could also eliminate these ticks with:

```haskell
instance ToJSON Registration where
  toJSON Registration { username, password, email } =
    Object [ ("username", String username)
           , ("password", String password)
           , ("email",    String email)
           ]
```

library/Fission/User/Registration/Types.hs
```diff
@@ -7,21 +7,21 @@ import Fission.Prelude
 data Registration = Registration
   { username :: !Text
   , password :: !Text
-  , email    :: !(Maybe Text)
+  , email    :: !Text
```
expede: 👍

PR #237, opened 2019-12-20
Request Body Inspection Middleware

expede:
@dholms and I discussed what to do about the fact that reading the request body from the buffer is destructive. It makes sense that it is this way (incoming files may be very large, so not duplicating them can be critical), but is also deeply annoying.

We probably want to additionally create middleware that checks the size of an incoming file, to protect against DoS attacks. The downsides of maintaining a file uploader 😛

### DIFF
--- a/library/Fission/Web/Middleware/AuthBody.hs
+++ b/library/Fission/Web/Middleware/AuthBody.hs
@@ -1,0 +1,39 @@
+{-# OPTIONS_GHC -fno-warn-deprecations #-}
+
+module Fission.Web.Middleware.AuthBody (foo) where
+
+import Fission.Prelude
+
+import Network.Wai (Middleware, strictRequestBody)
+import Network.Wai.Internal
+
+import qualified Data.ByteString.Lazy       as Lazy
+
+import qualified Data.Vault.Lazy as Vault
+
+-- type Middleware = Application -> Application
+-- type Application =
+--    Request -> (Response -> IO ResponseReceived) -> IO ResponseReceived
+
+-- how to generate a unique vault key outside this function:
+-- `vaultKey <- newKey`
+
+foo :: Vault.Key Text -> Middleware
+foo vaultKey app req sendResponse = do
+  lazyBody <- strictRequestBody req
+
+  let
+    oldVault      = vault req
+    body          = Lazy.toStrict lazyBody
+    hash          = undefined -- TODO hash the body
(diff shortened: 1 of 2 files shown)


library/Fission/Web/Middleware/AuthBody.hs
```diff
@@ -0,0 +1,39 @@
...
+import Network.Wai (Middleware, strictRequestBody)
+import Network.Wai.Internal
+
+import qualified Data.ByteString.Lazy       as Lazy
+
+import qualified Data.Vault.Lazy as Vault
+
+-- type Middleware = Application -> Application
+-- type Application =
+--    Request -> (Response -> IO ResponseReceived) -> IO ResponseReceived
```
expede: Just a reminder of the types we're playing with

expede:
@dholms this is mostly done. Just need to know how we need to hash the file to make everything line up. If you want to take this over, or just copy/paste into your branch, that's fine 👍

library/Fission/Web/Middleware/AuthBody.hs
```diff
@@ -0,0 +1,39 @@
+{-# OPTIONS_GHC -fno-warn-deprecations #-}
```
expede: Because they don't want you normally working with `requestBody` directly, they added helpers like `strictRequestBody` and deprecated the `requestBody` accessor. It's fine, because we're using the `Internal` module since we're writing pretty surgical middleware.

library/Fission/Web/Middleware/AuthBody.hs
```diff
@@ -0,0 +1,39 @@
...
+-- type Middleware = Application -> Application
+-- type Application =
+--    Request -> (Response -> IO ResponseReceived) -> IO ResponseReceived
+
+-- how to generate a unique vault key outside this function:
+-- `vaultKey <- newKey`
+
+foo :: Vault.Key Text -> Middleware
+foo vaultKey app req sendResponse = do
+  lazyBody <- strictRequestBody req
```
dholms: are we able to pipe this directly through? or do we need to read the whole thing into memory, hash it, and then reserialize it?

dholms: Also if we did this chunk by chunk, we could keep a counter for the size, and if it gets above a certain threshold (ie the user lied about file size), then we throw an error

PR #238, opened 2019-12-20
Update Heroku provision to throw 409 on unique conflict

bechurch:
## Summary
Update the Heroku Provision endpoint to throw a 409 AlreadyExists when a unique constraint error is thrown.
image

## After Merge

### DIFF
--- a/library/Fission/User/Mutation.hs
+++ b/library/Fission/User/Mutation.hs
@@ -29,14 +29,18 @@ createWithHeroku ::
   -> Text
   -> m (Either Error.Create UserId)
 createWithHeroku herokuUUID herokuRegion username password = runDBNow \now -> do
-  addOnId <- insert HerokuAddOn
-    { herokuAddOnUuid       = herokuUUID
-    , herokuAddOnRegion     = Just herokuRegion
-    , herokuAddOnInsertedAt = now
-    , herokuAddOnModifiedAt = now
-    }
+  let herokuAddOnRecord = HerokuAddOn
+        { herokuAddOnUuid       = herokuUUID
+        , herokuAddOnRegion     = Just herokuRegion
+        , herokuAddOnInsertedAt = now
+        , herokuAddOnModifiedAt = now
+        }
+  insertUnique herokuAddOnRecord >>= \case
+    Just addOnId ->
+      create username password Nothing (Just addOnId) now
 
-  create username password Nothing (Just addOnId) now
+    Nothing ->
+      return (Left Error.AlreadyExists)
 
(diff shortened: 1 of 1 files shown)


expede:
I'm actually not totally sure what Heroku expects from their docs

> Requests MAY be sent multiple times (Heroku follows the pattern of at least once delivery), and endpoints MUST respond idempotently

Idempotency in the DB, or in how we respond, as well?

expede:
There we go:

> The add-on provision request POST /heroku/resources sent from Heroku to the partner is expected to be idempotent. In other words, the same request MAY be sent to the partner multiple times, but only a single resource SHOULD be created per uuid, and the same response SHOULD be returned each time. However, if the resource has been deprovisioned, it SHOULD NOT be created again and a 410 SHOULD be returned. In order to facilitate data expunging by the partner the time frame in which a provision request may be retried is limited to 24 hours.

So, we should respond with a `201 Created` if they send the same request again

library/Fission/User/Mutation.hs
```diff
@@ -29,14 +29,18 @@ createWithHeroku ::
...
+        , herokuAddOnInsertedAt = now
+        , herokuAddOnModifiedAt = now
+        }
+  insertUnique herokuAddOnRecord >>= \case
+    Just addOnId ->
+      create username password Nothing (Just addOnId) now
 
-  create username password Nothing (Just addOnId) now
+    Nothing ->
+      return (Left Error.AlreadyExists)
```
expede: Oh, this is great, though! At the call site, it can capture this is a normal `Left` value and return the `201` 👍

expede: Good job on the internal error modelling 🎉

bechurch: Ok so in this case whats the best way to handle their password 😬 ? 

We can't return it. we can create a new one but then technically it isn't idempotent anymore.

expede: Yeah, good point. The docs aren't clear so I'll email them 🤷‍♀

bechurch: @expede If you do so can you cc me :)

bechurch: @expede any response?

bechurch:
@expede This is update and ready for review.

Note that after our call yesterday we decided to leave this as a 409

library/Fission/User/Mutation.hs
expede: Not yet 🤷‍♀ Whatever let's see if this causes issues

PR #239, opened 2019-12-31
MonadAWS & MonadDNSLink

expede:
Easing back into things, & to help with @icidasset's testing


library/Fission/Prelude.hs
```diff
@@ -64,3 +65,6 @@ import Fission.Internal.UTF8       (putText, putTextLn, textShow)
 
 identity :: a -> a
 identity a = a
+
+bind :: Monad m => (a -> m b) -> m a -> m b
+bind = (=<<)
```
expede: @icidasset fine, I relent!

library/Fission/Web.hs
```diff
@@ -39,23 +38,18 @@ import qualified Fission.Web.User    as User
...
+  , MonadTime           (RIO cfg)
+  , MonadDB             (RIO cfg)
+  , MonadDNSLink        (RIO cfg)
+  , MonadDB                      m
+  , MonadReader              cfg m
+  , Has IPFS.Peer            cfg
+  , Has Web.Host             cfg
+  , Has Heroku.ID            cfg
+  , Has Heroku.Password      cfg
+  , HasLogFunc               cfg
```
expede: Coming along!

app/Fission/Internal/Orphanage/RIO.hs
```diff
@@ -12,19 +12,137 @@ import qualified RIO.ByteString.Lazy as Lazy
...
+
+    let
+      baseURL =
+        case maySubdomain of
+          Nothing ->
+            domain
+
+          Just (Subdomain subdomain) ->
+            subdomain <> "." <> domain
+
```
bechurch: 💅 Could be nice as a named function

app/Fission/Internal/Orphanage/RIO.hs
```diff
@@ -12,19 +12,137 @@ import qualified RIO.ByteString.Lazy as Lazy
...
+          return (Right mockRecordResponse)
+
+      changeRecord' = do
+        logDebug <| "Updating DNS record at: " <> displayShow domain
+
+        req <- createChangeRequest
+
+        within NorthVirginia do
+          res <- send req
+          return <| validate res
```
bechurch: 💅 Having this inside of a where as opposed to their own functions feels like were misusing `where` and they would better fit as their own functions. But who knows maybe thats just the haskell way :)

app/Fission/Internal/Orphanage/RIO.hs
```diff
@@ -12,19 +12,137 @@ import qualified RIO.ByteString.Lazy as Lazy
...
 
+import           Fission.AWS.Route53.Class
+import           Fission.AWS.Types as AWS
+import           Fission.AWS
 import qualified Fission.Config as Config
+import           Fission.Internal.UTF8
+import           Fission.IPFS.DNSLink
+
+instance (Has AWS.AccessKey cfg, Has AWS.SecretKey cfg) => MonadAWS (RIO cfg) where
+  liftAWS awsAction = do
```
bechurch: Question: Where does `liftAWS` get used?

library/Fission/Prelude.hs
icidasset: Hah 😂 Yay!

library/Fission/IPFS/DNSLink/Class.hs
```diff
@@ -0,0 +1,11 @@
...
+
+import           Network.IPFS.CID.Types
+import           Servant
+
+import           Fission.AWS.Route53.Class
+import           Fission.AWS.Types as AWS
+import           Fission.Prelude
+
+class MonadRoute53 m => MonadDNSLink m where
+  set :: Maybe AWS.Subdomain -> CID -> m (Either ServerError AWS.DomainName)
```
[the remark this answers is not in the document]
expede: @bechurch To contain the `class`.

I'm guessing that you're wondering why this file since there's also a `Fission.IPFS.DNSLink`. If the upper-level (which reexports this module) later contains a bunch of helper functions, we may want to import _just_ the `class` to avoid module cycles elsewhere. This is the same reason as the `*.Types` modules.

app/Fission/Internal/Orphanage/RIO.hs
expede: Good instinct! So, it depends. We can break them out, but it probably won't have much benefit. Here's why:

# The Big Reason

The big reason why this is special: it's typeclass instance (Elixir's `defimpl`). We don't expect to be able to reuse the functions separately at all — the point is abstraction. We can break them out, but it won't really help us.

# General Discussion

It common to use `where` to contain larger functions, but that's not a blanket statement. I agree _in general_ that keeping them small is a good idea.

`where`s are inner functions, so they have access to the arguments of the named function. This means less writing out parameters to link things together, and more compiler specialization (not that this is something that we're optimizing for).

In this case we really don't care about the mock function; it's there for local development only. Since we're not exporting it from the module, they're roughly equivalent except for the below.

app/Fission/Internal/Orphanage/RIO.hs
expede: @bechurch 

`MonadAWS` is defined by the AWS package that we're using. It's used internally in all of the calls to AWS that we make. You won't see it used on the surface, since it's being used to hook our code into their library.

A pattern to get a lot of benefit from ad hoc polymorphism is to define some general function, and then get a bunch of functions that depend on that "for free".

For example, in Elixir ad hoc polymorphism is provided by protocols. For `Enumerable`, you define at least<sup>1</sup> `reduce/3` for you data type. Then you get access to almost all of the `Enum` module. The upshot is that for defining one function, you get several dozen that also work on your custom data type.

<sup>1</sup> `Enumerable` has 4 functions to define, but most only actually need `reduce/3`

app/Fission/Internal/Orphanage/RIO.hs
expede: Sure, will do 👍

library/Fission/AWS/Route53.hs
```diff
@@ -1,147 +1,3 @@
-module Fission.AWS.Route53
-  ( createChangeRequest
-  , registerDomain
-  ) where
+module Fission.AWS.Route53 (module Fission.AWS.Route53.Class) where
```
dholms: why bother with this instead of just eliminating this module entirely? readability in imports?

library/Fission/AWS/Types.hs
```diff
@@ -1,11 +1,13 @@
 -- | Fission-speciric AWS types
```
dholms: (sp): specific
(I know this isn't your change but just saw it now)

app/Fission/Internal/Orphanage/RIO.hs
```diff
@@ -3,28 +3,141 @@
...
+
+import           Fission.AWS
+import           Fission.AWS.Types as AWS
+import           Fission.AWS.Route53.Class
 
 import qualified Fission.Config as Config
+import           Fission.Internal.UTF8
+
+import           Fission.IPFS.DNSLink
+import qualified Fission.URL.Sibdomain as Subdomain
```
dholms: `Sibdomain` -> `Subdomain`

library/Fission/URL/Subdomain.hs
```diff
@@ -0,0 +1,18 @@
+-- | URL helpers
+module Fission.URL.Subdomain
+  ( subdomain
```
dholms: think this should be `normalize` instead

library/Fission/AWS/Route53.hs
expede: That, and because in the future a lot of functionality could live in this module. Having to run around and remove the prefix to get access to helper functions could be frustrating. Probably not a big deal in either case, so happy to delete these if it's helpful.

library/Fission/AWS/Types.hs
expede: haha kk will fix

library/Fission/AWS/Route53.hs
dholms: nope just curious :+1:

library/Fission/Web/User.hs
```diff
@@ -27,21 +26,13 @@ type ResetRoute = "reset_password"
...
+server ::
+  ( MonadDB       m
+  , MonadTime     m
+  , MonadDNSLink  m
+  , MonadLogger   m
+  )
   => ServerT API m
 server = Create.server
-    :<|> const Verify.server
+    :<|> (\_ -> Verify.server)
```
dholms: these are the same thing, right? why prefer this?

library/Fission/Web/User.hs
```diff
@@ -27,21 +26,13 @@ type ResetRoute = "reset_password"
...
-     , MonadTime     m
-     , MonadReader                cfg m
-     , Has IPFS.Gateway           cfg
-     , Has AWS.DomainName         cfg
-     , Has AWS.AccessKey          cfg
-     , Has AWS.SecretKey          cfg
-     , Has AWS.ZoneID             cfg
-     , Has AWS.Route53MockEnabled cfg
-     )
+server ::
```
dholms: :heart_eyes_cat:

library/Fission/Web/User/Create.hs
```diff
@@ -4,42 +4,36 @@ module Fission.Web.User.Create
...
+  )
   => ServerT API m
 server (User.Registration username password email) = do
-  createResponse <- runDBNow <| User.create username password (Just email) Nothing
-  void <| Web.Err.ensure <| createResponse
-  void <| Web.Err.ensureM =<< registerDomain username splashCID
+  Nothing
+    |> User.create username password (Just email)
+    |> runDBNow
+    |> bind Web.Err.ensure
```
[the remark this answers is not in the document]
dholms: echoing the initial skepticism and eventual understanding haha

library/Fission/Web/User.hs
expede: Yeah, they're the same. The lambda version is one character longer, and very clear to newcomers. Again, I'm not married to this, but thought I'd see if it made the logic easier to understand.

Thoughts?

library/Fission/URL/Subdomain.hs
```diff
@@ -0,0 +1,21 @@
...
+  , module Fission.URL.Subdomain.Types
+  ) where
+
+import Fission.Prelude
+import Fission.URL.Subdomain.Types
+import Fission.URL.DomainName.Types
+
+-- | Prefix a domain named with an optional subdomain
+normalizePrefix :: DomainName -> Maybe Subdomain -> DomainName
+normalizePrefix domain = \case
```
dholms: Very minor, but: 
when a whole function is wrapped in a `case` statement, my brain parses it better with something like
```
normalizePrefix domain Nothing = domain
normalizePrefix domain (Just subdomain) = prefix domain subdomain
```
Thoughts? Any reason to prefer one over the other?

library/Fission/Web/User.hs
dholms: No real preference either way. `const` did confuse me when first getting into the code base, but now I'm used to it so don't even think twice.

I think it makes it slightly less readable but slightly easier to understand... if that makes any sense haha

library/Fission/URL/Subdomain.hs
expede: Yeah, they're the _exact same_ to the compiler, so this is totally a style thing. There's no community consensus on this. I have a few reasons for preferring the `\case` version.

1. It saves a few keystrokes. Whatever.
2. It's consistent
  * `Maybe` has two branches, but we have places where it's much more
    * For example, the different kinds of IPFS errors
  * We pipe into `\case` a fair bit inside functions
3. It's clear that we're switching on the last argument as logic
4. It's still clear from the type signature how many arguments there are

This is purely style preference. I also realize that I can read point-free code pretty natively, which is not typical of most devs. Maybe we can make a style choice like "if there's 1-3 branches, then use function heads. More than 3 and use `\case`".

Any preference / thoughts?

library/Fission/Web/User.hs
expede: Yup totally makes sense! This is where I kind of get to — there are lots of things that more experiences Haskellers like, because they make sense as you become familiar with the patterns. Lenses are actually surprisingly readable/expressive _once you put the time in_, but there's that initial learning curve. I'm trying to figure out where the balance is 🤷‍♀

library/Fission/URL/Subdomain.hs
dholms: yeah no huge preference here. I'm justed to seeing `\case` within functions so I think it just confused me a bit to see the last argument being cased on like that. 

Didn't know the term "point-free code". Yup that definitely sums up why I had a hard initial time parsing this

I still have a slight preference for using function heads just because I find them more immediately intuitive, but it's slight so really no need to adapt to that.

Curious tho, why would `\case` be better with more branches? Just because you don't have to keep writing the function name & other params?

expede: > Curious tho, why would \case be better with more branches? Just because you don't have to keep writing the function name & other params?

Yes, exactly. Like, this would be annoying to type out:

```haskell
checkIn name colour tasks Monday    = -- ...
checkIn name colour tasks Tuesday   = -- ...
checkIn name colour tasks Wednesday = -- ...
checkIn name colour tasks Thursday  = -- ...
checkIn name colour tasks Friday    = -- ...
checkIn name colour tasks Saturday  = -- ...
checkIn name colour tasks Sunday    = -- ...

-- versus

checkIn name colour tasks = \case
  Monday    -> -- ...
  Tuesday   -> -- ...
  Wednesday -> -- ...
  Thursday  -> -- ...
  Friday    -> -- ...
  Saturday  -> -- ...
  Sunday    -> -- ...
```

app/Fission/Internal/Orphanage/RIO.hs
bechurch: Nice explanation!

app/Fission/Internal/Orphanage/RIO.hs
bechurch: Also a great explantation. Thanks Brooke.

I think my "style" preference is still a separate function. But there's no reason other than I like saving on indentation :P

PR #240, opened 2019-12-31
MonadHerokuAddOn

expede:
More of the same!

### DIFF
--- a/library/Fission/Web.hs
+++ b/library/Fission/Web.hs
@@ -13,12 +13,13 @@ import           Network.IPFS
 import qualified Network.IPFS.Types as IPFS
 
 import           Fission.Prelude
-import qualified Fission.Config     as Config
+import qualified Fission.Config  as Config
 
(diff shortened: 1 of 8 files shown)


expede:
Branched from #239

library/Fission/Web.hs
```diff
@@ -67,20 +64,17 @@ app = do
...
+  , MonadHerokuAddOn m
   )
   => m (Context Auth.Checks)
 mkAuth = do
-  Heroku.ID       hkuID   <- Config.get
-  Heroku.Password hkuPass <- Config.get
-  cfg <- ask
-
+  cfg        <- ask
+  herokuAuth <- authorize
```
dholms: :nail_care: 
I think this might be a little more clear as 
```
herokuAuth <- Heroku.authorize
```
the name `herokuAuth` helps, but it's still not immediately apparent where `authorize` is coming from

PR #241, opened 2019-12-31
MonadLinkedIPFS

expede:
Branched from #240

PR #242, opened 2019-12-31
MonadReflectiveServer

expede:
Continuing the pattern from #241


bechurch:
Question: Why is it called ReflectiveServer?

expede:
> Why is it called ReflectiveServer?

Because it's doing runtime reflection on info about the server

PR #243, opened 2019-12-31
Remove RIO orphans

expede:
Continuing #242


library/Fission/Types.hs
```diff
@@ -36,42 +38,65 @@ import           Fission.Platform.Heroku.AddOn
...
+                   , Monad
+                   , MonadIO
+                   , MonadUnliftIO
+                   , MonadReader Config
+                   , MonadThrow
+                   , MonadCatch
+                   , MonadMask
+                   )
+
+-- TODO move to Fisson.hs
```
bechurch: Todo still here

library/Fission/Types.hs
```diff
@@ -36,42 +38,65 @@ import           Fission.Platform.Heroku.AddOn
...
+runFission :: MonadIO m => Config -> Fission a -> m a
+runFission cfg actions =
+  actions
+    |> unFission
```
bechurch: May be better named as `internalConfig` or something similar

bechurch: cute name though so this isnt an over my dead body :)

library/Fission/Types.hs
```diff
@@ -36,42 +38,65 @@ import           Fission.Platform.Heroku.AddOn
...
+    |> runRIO cfg
+
+instance MonadLogger Fission where
+  monadLoggerLog loc src lvl msg = Fission (monadLoggerLog loc src lvl msg)
+
+instance MonadTime Fission where
+  currentTime = liftIO getCurrentTime
+
+instance MonadReflectiveServer Fission where
+  getHost = asks host
```
bechurch: Question: Where does this asks function come from?

library/Fission/Web.hs
```diff
@@ -37,47 +33,27 @@ import qualified Fission.Web.User    as User
...
+  , MonadLinkedIPFS       m
+  , MonadReflectiveServer m
   )
-  => m Application
-app = do
-  cfg     <- ask
-  auth    <- mkAuth
-  appHost <- getHost
-
+  => (forall a . m a -> Handler a)
```
bechurch: Question: What does this line...um...mean? 😅

library/Fission/Types.hs
```diff
@@ -1,18 +1,18 @@
-{-# OPTIONS_GHC -fno-warn-orphans #-}
-{-# LANGUAGE UndecidableInstances #-}
+module Fission.Types
```
dholms: this file :heart_eyes_cat:  :heart_eyes_cat:

library/Fission/Types.hs
dholms: piggybacking off this, is there any type-checking going on with `asks`? or are we saying "trust me, there's a value in the config with key `host`"?

library/Fission/Types.hs
```diff
@@ -156,51 +172,38 @@ instance
           |> update Txt dnsLinkURL
           |> fmap \_ -> Right (AWS.DomainName baseURL)
 
-instance Has IPFS.Peer cfg => MonadLinkedIPFS (RIO cfg) where
-  getLinkedPeers = pure <$> Config.get
+instance MonadLinkedIPFS Fission where
+  getLinkedPeers =
```
dholms: why not
```haskell
  getLinkedPeers = pure <$> asks ipfsRemotePeer
```
with short data flows like this, I think the one liners are much more readable

library/Fission/Types.hs
expede: Totally get what you're saying, but you're missing that the `newtype` also contains a complete `RIO`. i.e. it's not just the internal config, it's both a monad and the config.

`unFission` is literally "unwrap this from `Fission` to expose the underlying `RIO Config`".

library/Fission/Types.hs
expede: @bechurch 

> Question: Where does this asks function come from?

`asks` comes from `Reader`, which is what `RIO` uses to automagically thread the config through. It takes a function, and plugs the config into it, and gives you the result. Good for things like selectors (as seen here).

Here's some docs: https://www.stackage.org/haddock/lts-14.19/rio-0.1.12.0/RIO-Prelude.html#v:asks

expede: @dholms 

> piggybacking off this, is there any type-checking going on with asks? or are we saying "trust me, there's a value in the config with key host"?

Haskell doesn't really have a "trust me" switch... it's more like "prove it to me!".

Because we have so much abstraction down in the business logic, we can reasonably hardcode the `Config` in `Fission`. It knows that we have `host` in scope because the type of Fission is `RIO Config`, and `Config` has a `host` field.

library/Fission/Web.hs
expede: @bechurch 

Good question! This is not a common thing, so don't worry about it too much. I did this to clean up the constraints. It's not _required_, but is much cleaner this way / a nice to have.

# Short Answer

We're defining a locally scoped type variable (inside the parens).

# Long Answer

Every Haskell function has an implicit `forall` at the beginning of the type signature. For example:

```haskell
myLogInfo :: (Monad m, Show a) => a -> m a

-- is actually
myLogInfo :: forall m a . (Monad m, Show b) => a -> m a
```

This is called "universal quantification", just like in logic. "For all `a`s, the following relationships holds true."

We can also nest these inside the type of passed functions. For example:

```haskell
```

library/Fission/Types.hs
bechurch: Ok, how about `unwrapFission` `internalRIOMonad` `getInternalRIOMonad`? My only concern with `unFission` is someone new coming in will wonder what is it to `unFission` something

library/Fission/Web.hs
bechurch: Ahh gotcha! Thats what was tripping me up I was already reading types like boolean algebra and wondering if we didn't specify it before why were we now.

library/Fission/Types.hs
expede: @dholms it makes me so happy to hear you say that. You certainly have some Haskeller in you 😛 I agree with you, but am trying to not have what I'm used to get in the way of others.

I'm curious to have others weigh in. @bechurch / @icidasset?

library/Fission/Types.hs
expede: I mean, they'll also be confused by `unRIO`, `unReader`, `unWriter`, and so on. But sure, `innerRIO` or something can work.

library/Fission/Types.hs
bechurch: Haskell noob coming in: both feel the same really? 😅 

```haskell
getLinkedPeers = pure <$> asks ipfsRemotePeer
```
Is more haskelly and reads simpler once you know the operator.

```haskell
|> fmap pure
```
I think communicates more of whats going on under the hood. but it's atypical. (Haskell and those proficient in Haskell love their many operators)


My only preference is that we pick one overall, stick with it and reduce the number of operators you have to think about in any given file 😛. Again Haskell noob opinions over here.

library/Fission/Types.hs
expede: Sorry, to clarify (from DMs). That's not meant in frustration; more that they'll run into this. I've renamed it `unwrapFission` 👍

library/Fission/Types.hs
expede: > whats going on under the hood

Yeah, we generally try to get away from under the hood. Subordinate detail to abstraction. I think that they're roughly equivalent in this case.

We did lose the visual guide by dropping `$`. `<$>` is `$` on containers `< >`. It's an extremely common operator, and I'm not sure why operators are scarier than named functions. You need to know what `bind` does, the same way that you need to know what `=<<` does. But I'm also coming from the far other end of the spectrum.

As an aside, I've been thinking about the number of operators since the last time it came up just before the break. It's interesting, since in C-family languages, they _heavily_ overload their operators, which should also be confusing. Is it just familiarity? I dunno. All things that I'm thinking about with Prime.

PR #244, opened 2019-12-31
All the things

expede:
In case anyone wants to see all of these changes at once, here's a diff


library/Fission/Config/Types.hs
```diff
@@ -67,63 +67,3 @@ instance HasProcessContext Config where
 instance HasLogFunc Config where
   logFuncL = lens logFunc \cfg newLogFunc' ->
     cfg { logFunc = newLogFunc' }
-
```
expede: Because all of the `Config` access is now limited to `Fission.Types`, and there's no `Has` constraints throughout the project, `Config.get` is much less useful.

* Delete the `Fission.Config` module
* Remove the now unused `Has` instances

You can still get access to the values with `asks SOME_KEY`. For example:

```haskell
timeout <- asks ipfsTimeout
```

PR #245, opened 2020-01-01
Export logErrorN and logDebugN

bechurch:
A similar PR may already be submitted!
Please search among the [Pull request](../) before creating one.

Thanks for submitting a pull request! Please provide enough information so that others can review your pull request:

For more information, see the `CONTRIBUTING` guide.

## Summary
Simply exports logErrorN and logDebugN from `Prelude`.

This is so in `cli` and `web-api` we can write
```haskell
logErrorN "Hi!"
```
instead of
```haskell
logError ("Hi!" :: Text)
```


expede:
hahaha I read the titles as if we were doing some export that ran in O(log n) 😆 I was like "I dunno, log n is pretty darn good" 😛

PR #246, opened 2020-01-01
Fix 500 on heroku deprovision

bechurch:
## Summary
Our Deprovision endpoint was throwing a 500 for any request.

This was due to that we were trying to due multiple deletes in a single transaction.

This PR does the following:
- deletes associations individually
- moves all mutations to the `Mutations.hs` files
- moves all db queries to the `Query.hs` files

### Test 
- [x] Ensure we successfully delete heroku addons on deprovision
- [x] Ensure we throw a 410 on duplicate requests
- [ ] Test that unpinning works as expected

## After Merge
* [ ] Ensure we run the migration. (you will have to delete a few duplicates)


library/Fission/User/CID/Mutation.hs
```diff
@@ -2,6 +2,8 @@
 module Fission.User.CID.Mutation
   ( create
   , createX
+  , deleteAll
+  , deleteExactUserCid
```
expede: Totally fine to leave this named as such. However, we can also use prefixes to disambiguate when needed.

```haskell
import Fission.User.CID.Mutation as User.CID

foo = User.CID.deleteExact userId cid
```

library/Fission/User/CID/Query.hs
```diff
@@ -0,0 +1,24 @@
...
+  select <| from \userCid -> do
+    where_ (userCid ^. UserCidUserFk ==. val userId)
+    return userCid
+
+-- | Find all CIDs that remain from a list
+getUserCidsByCids :: MonadDB m => [CID] -> Transaction m [Entity UserCid]
+getUserCidsByCids cids =
+  select <| from \userCid -> do
+    where_ (userCid ^. UserCidCid `in_` valList cids)
+    return userCid
```
expede: Same note one naming

library/Fission/User/CID/Query.hs
```diff
@@ -0,0 +1,24 @@
...
+import           Network.IPFS.CID.Types
+import           Database.Esqueleto hiding ((=.), update)
+
+import           Fission.Models
+import           Fission.Prelude
+
+-- | CIDs associated with a user
+getUserCidsByUserId :: MonadDB m => UserId -> Transaction m [Entity UserCid]
+getUserCidsByUserId userId =
+  select <| from \userCid -> do
```
expede: While we're in here, maybe switch this to `selectFirst`. This will also change the return type to `Transaction m (Maybe (Entity UserCid))`

library/Fission/User/CID/Query.hs
expede: Also, while we're in here: we could think about turning these into classes to make mocking out in testing more granular.

```haskell
class MoandDB m => MonadDBQuery m model where
  getBy :: SqlExpr (Value Bool) -> Transaction [Entity model]

-- These can likely be defined in terms of `MonadQuery.get`

getAll
getAllById
-- and so on
```

expede: To hook into existing intuitions about why this would be helpful, we're essentially 12-factor/hexagonalizaing our app logic. Tests now can define their own version of `getBy` to return dummy data to do tests with zero DB setup/cleanup, really detailed tracing in development, and so on.

Correctness composes in this style, so if we test an endpoint in a dummy monad, and then test the production typeclass implementation, we know that the actual business logic works correctly. This is one of the advantages of pairing a strong type system with enforced purity.

library/Fission/Web.hs
```diff
@@ -57,6 +60,7 @@ app handlerNT auth appHost = do
 -- | Web handlers for the 'API'
 server ::
   ( MonadDB               m
+  , MonadDBQuery UserCid m
```
bechurch: @expede Got a version of this working just far enough to see what the type signature of this approach would be.

I gotta say I'm not loving that at the top we'd have to define a type constraint for every query type.

```haskell
MonadDB m
=> MonadDBQuery UserCid m
=> MonadDBQuery User m
=> MonadDBQuery App m
=> MonadDBQuery Domain m
```

I imagine theres a better way or perhaps a technique to avoid this that I dont know. WDYT?

expede: > I'm not loving that at the top we'd have to define a type constraint for every query type.

I mean, being explicit is the entire point here. If we just want unconstrained DB access, then it's pretty straightforward — don't use these and leave it at `MonadDB`. I'm okay with that, as per our call yesterday.

Breaking them out is a tradeoff. It's really not that much boilerplate, and you get a lot more resource control. IMO tracing what a function can touch is more of a thing for mutations than queries, so the choice for these is totally up to you. Breaking queries out will make mocking out tests easier (because now we have granular access)

# Alias

If you want to alias several of these together, that's totally doable though I say discourage it in this case:

```haskell
type MonadQueryAllModels m = (MonadDBQuery User m, MonadDBQUery Domain m, ...)
```

The problem here is that it temps people to use widest constraint out of laziness/convenience, this negating most of the value.

# Default Implementation

To get around having to _define_ these, you can use a default implementation:

```haskell
class MonadDBQuery model m where
  getBy :: SqlExpr something something -> Transaction m [Entity model]
  getBy selector = select <| from \model ->
    where_ (selector model)
    return model
```

Now you get all of the models implemented out of the box with the same version, but you can still overwrite specific cases if you need to (e.g. tests, debugging, &c)

bechurch:
@expede This should be ready to review again.

I broke the Monad change out into a different PR so that our bug fix isn't held up :)

https://github.com/fission-suite/web-api/pull/249

library/Fission/Web/IPFS/Pin.hs
```diff
@@ -63,19 +64,10 @@ unpin ::
...
+    UserCidMutation.destroyExact userId cid
+    UserCidQuery.getByCids [cid]
 
-    select <| from \userCid -> do
-      where_ (selectExact userCid)
-      limit 1
-      return userCid
-
-  when (null remaining) do
+  when (length remaining == 0) do
```
expede: 💅 A quick stylistic note: it's generally more common to phrase things declaratively, rather than inspecting a property

```haskell
when (remaining == []) do
  void <| Web.Err.ensure =<< IPFS.Pin.rm cid
```

library/Fission/Web/IPFS/CID.hs
```diff
@@ -9,13 +9,12 @@ import Servant
...
 allForUser :: MonadDB m => Entity User -> ServerT API m
 allForUser (Entity userId _) = runDB do
-  hashes <- select <| from \userCid -> do
-    where_ (userCid ^. UserCidUserFk ==. val userId)
-    return (userCid ^. UserCidCid)
-
-  return (unValue <$> hashes)
+  userCids <- UserCIDQuery.getByUserId userId
+  let cids = getInner userCidCid <$> userCids
+  return cids
```
expede: These are the same:

```haskell
let cids = getInner userCidCid <$> userCids
return cids
```

```haskell
getInner userCidCid <$> userCids
```

They'll also probably have the same runtime performance (because rewrite rules will optimize it), but you don't actually need to set that variable. We know that it's the return of the function by it just being the last line.

expede: Oh, sorry, I should clarify something about `do` notation.

```haskell
do
  let foo = bar
  -- more stuff

do
  foo <- return bar
  -- more stuff

return bar >>= \foo -> -- more stuff
```

I wish they had called it something like `const` to disambiguate it from `let ... in`, but here we are

library/Fission/Web/Heroku/Deprovision.hs
```diff
@@ -50,73 +55,39 @@ deleteAssociatedWith ::
...
   )
   => UUID
   -> Transaction m HerokuAddOnId
-herokuAddOnByUUID uuid' = ensureOneId err410 =<< query
-  where
-    query = select <| from \herokuAddOn -> do
-      where_ (herokuAddOn ^. HerokuAddOnUuid ==. val uuid')
-      limit 1
-      return herokuAddOn
+herokuAddOnByUUID uuid' = ensureOneId err410 =<< (UserQuery.getHerkouAddonByUUID uuid')
```
expede: 💅 No need for the parens:

```haskell
herokuAddOnByUUID uuid' = ensureOneId err410 =<< UserQuery.getHerkouAddonByUUID uuid'
```

Or if you want pipes (though I personally find this less clear)

```haskell
herokuAddOnByUUID uuid' =
  uuid'
    |> UserQuery.getHerkouAddonByUUID
    |> bind (ensureOneId err410)
```

library/Fission/Web/Heroku/Deprovision.hs
```diff
@@ -50,73 +55,39 @@ deleteAssociatedWith ::
...
-userIdForHerokuAddOn addOnId = ensureOneId err410 =<< query
-  where
-    query = select <| from \user -> do
-      where_ <| user ^. UserHerokuAddOnId ==. val (Just addOnId)
-            &&. user ^. UserActive        ==. val True
-      limit 1
-      return user
-
--- | Heroku add-on with a specific UUID
+userIdForHerokuAddOn addOnId = ensureOneId err410 =<< (UserQuery.getHerkouAddonByUserId addOnId)
```
expede: Same as line 93

library/Fission/User/Query.hs
```diff
@@ -0,0 +1,23 @@
...
+getHerkouAddonByUserId addOnId = select <| from \user -> do
+      where_ <| user ^. UserHerokuAddOnId ==. val (Just addOnId)
+            &&. user ^. UserActive        ==. val True
+      limit 1
+      return user
+
+getHerkouAddonByUUID :: MonadDB m => UUID -> Transaction m [Entity HerokuAddOn]
+getHerkouAddonByUUID uuid = select <| from \herokuAddOn -> do
+      where_ (herokuAddOn ^. HerokuAddOnUuid ==. val uuid)
+      limit 1
```
expede: There's also a `selectFirst` function: https://www.stackage.org/haddock/lts-14.20/persistent-2.9.2/Database-Persist-Class.html#v:selectFirst

```haskell
selectFirst :: (MonadIO m, PersistRecordBackend record backend) => [Filter record] -> [SelectOpt record] -> Transaction m (Maybe (Entity record)) 
```

It's a nice fit because:
1. It returns a `Maybe`, which is more likely what you want rather than a `[]` that you later have to inspect)
2. Has the `limit` built-in

library/Fission/Web/IPFS/CID.hs
bechurch: So looking at this. I dont think your first comment is correct?
Changing this:
```haskell
let cids = getInner userCidCid <$> userCids
return cids
```

To this:
```haskell
getInner userCidCid <$> userCids
```
resulted in a compilation error.

However this compiled just fine
```haskell
  return (getInner userCidCid <$> userCids)
```

Is that right? or did I miss something?

expede: Oh, yeah, indeed you're right 😅 the `let` creates a `return` but you don't have one otherwise 🤦‍♀ Juuust ignore me

bechurch:
@expede Ive got this back up using selectFirst! 🎉 Ill merge at the end of the day if you have no comments

expede:
👀

library/Fission/User/CID/Mutation.hs
```diff
@@ -55,3 +57,16 @@ createX userId hashes = runDBNow \now -> do
...
+  delete <| from \userCid -> where_ (userCid ^. UserCidId `in_` valList userCidIds)
+
+destroyExact :: MonadDB m => UserId -> CID -> Transaction m ()
+destroyExact userId cid = do
+    delete <| from \userCid ->
+      where_ (selectExact userCid)
+  where
+    selectExact userCid =
+          userCid ^. UserCidCid    ==. val cid
+      &&. userCid ^. UserCidUserFk ==. val userId
```
expede: We've been following 2-space indentation, but this is another popular style.

```haskell
destroyExact :: MonadDB m => UserId -> CID -> Transaction m ()
destroyExact userId cid = do
  delete <| from \userCid ->
    where_ (selectExact userCid)
  where
    selectExact userCid =
          userCid ^. UserCidCid    ==. val cid
      &&. userCid ^. UserCidUserFk ==. val userId
```

I find that syntax highlighting and whitespace make it clear enough for me. How does it feel to you? Mostly thinking in terms of a longer-term company style, and totally open to opinions.

library/Fission/User/Mutation.hs
```diff
@@ -97,3 +99,11 @@ hashPassword' password = do
...
     Nothing           -> Left Error.FailedDigest
     Just secretDigest -> Right <| decodeUtf8Lenient secretDigest
+
+destroy :: MonadDB m => UserId -> Transaction m ()
+destroy userId = delete <| from \user ->
+    where_ (user ^. UserId ==. val userId)
+
+destroyHerokuAddon :: MonadDB m => UUID -> Transaction m ()
+destroyHerokuAddon uuid =
+  delete <| from \herokuAddOn -> where_ (herokuAddOn ^. HerokuAddOnUuid ==. val uuid)
```
expede: 2 vs 4 space indentation ☝️

library/Fission/User/Query.hs
```diff
@@ -0,0 +1,17 @@
...
+import qualified Database.Persist as P
+
+import           Fission.Models
+import           Fission.Prelude
+
+getHerkouAddonByUserId :: MonadDB m => HerokuAddOnId -> Transaction m (Maybe (Entity User))
+getHerkouAddonByUserId addOnId = selectFirst [UserHerokuAddOnId P.==. (Just addOnId), UserActive P.==. True] []
+
+getHerkouAddonByUUID :: MonadDB m => UUID -> Transaction m (Maybe (Entity HerokuAddOn))
+getHerkouAddonByUUID uuid = selectFirst [HerokuAddOnUuid P.==. uuid] []
```
expede: Does naming a function get us much over just using `selectFirst`?

Other than that:

library/Fission/Web/Heroku/Deprovision.hs
```diff
@@ -48,75 +53,41 @@ deleteAssociatedWith ::
...
-    where_ (userCid ^. UserCidUserFk ==. val userId)
-    return userCid
-
--- | Users associated with those Heroku add-ons
-userIdForHerokuAddOn ::
+deleteAssociatedRecords :: MonadDB m => UserId -> UUID -> [Entity UserCid] -> Transaction m ()
+deleteAssociatedRecords userId uuid userCids = do
+  UserCIDMutation.destroyAll      (entityKey <$> userCids)
+  UserMutation.destroy        userId
+  UserMutation.destroyHerokuAddon uuid -- TODO would be nice if this was cascading....
```
expede: 1. 

💅 SUPER nitpicky, but kinda funky alignment

2. 

> would be nice if this was cascading

Indeed

library/Fission/Web/Heroku/Deprovision.hs
```diff
@@ -48,75 +53,41 @@ deleteAssociatedWith ::
...
   => UUID
-  -> Transaction m HerokuAddOnId
-herokuAddOnByUUID uuid' = ensureOneId err410 =<< query
-  where
-    query = select <| from \herokuAddOn -> do
-      where_ (herokuAddOn ^. HerokuAddOnUuid ==. val uuid')
-      limit 1
-      return herokuAddOn
+  -> Transaction m (Entity HerokuAddOn)
+herokuAddOnByUUID uuid' = ensureEntity err410 =<< UserQuery.getHerkouAddonByUUID uuid'
```
expede: Random thought out of scope of this PR

```haskell
ensureEntityM httpErr action = ensureEntity httpErr =<< action

uuid'
  |> UserQuery.getHerkouAddonByUUID
  |> ensureEntityM err410
```

library/Fission/Web/IPFS/CID.hs
```diff
@@ -9,13 +9,12 @@ import Servant
...
 
 type API = Get '[JSON, PlainText] [CID]
 
 allForUser :: MonadDB m => Entity User -> ServerT API m
 allForUser (Entity userId _) = runDB do
-  hashes <- select <| from \userCid -> do
-    where_ (userCid ^. UserCidUserFk ==. val userId)
-    return (userCid ^. UserCidCid)
+  userCids <- UserCIDQuery.getByUserId userId
+  return (getInner userCidCid <$> userCids)
```
expede: 👍

library/Fission/Web/Heroku/Deprovision.hs
```diff
@@ -48,75 +53,41 @@ deleteAssociatedWith ::
...
-  addOnId  <- herokuAddOnByUUID uuid'
-  userId   <- userIdForHerokuAddOn addOnId
-  userCids <- cidsForUserId userId
-
-  userCids
-    |> associatedRecords uuid'
-    |> delete
+  addOn  <- herokuAddOnByUUID uuid'
+  addOnUser   <- userForHerokuAddOn (entityKey addOn)
+  userCids <- UserCIDQuery.getByUserId (entityKey addOnUser)
```
expede: Also kinda funky aligment. Whatevs.

library/Fission/Web/Heroku/Deprovision.hs
```diff
@@ -17,6 +17,11 @@ import           Servant
 import           Fission.Prelude
 import           Fission.Models
 
+import qualified Fission.User.Query        as UserQuery
+import qualified Fission.User.Mutation     as UserMutation
```
expede: We can also re-export these from `Fission.User`, &c. Then it would read like:


```haskell
userCids <- UserCID.getByUserId (entityKey addOnUser)
-- and
User.destroy userId
```

library/Fission/User/Query.hs
bechurch: So this is one convention ive picked up where I like to isolate anything that uses sql like conventions to a file under named functions to start building up our DSL.

So while its not much I think it improves both code readability and isolation of the database logic.

library/Fission/Web/Heroku/Deprovision.hs
bechurch: Hmm i like that

library/Fission/User/Query.hs
expede: >  isolation of the database logic.

We get a lot of that from the types and classes. But yes, in general I like this pattern. We used it pretty extensively in Lonely Planet for example, IIRC. Part of that was a consequence of Ecto letting you do _whatever_, but yeah, can't hurt to add some abstraction 👍

library/Fission/User.hs
```diff
@@ -0,0 +1,7 @@
+module Fission.User
```
bechurch: @expede wdyt, Did I do the import export correctly?

PR #247, opened 2020-01-02
Bump ipfs-haskell to 1.0.1

expede:
Switch to using the Hackage version of `ipfs-haskell`, and bump to the most recent version


package.yaml
```diff
@@ -60,6 +60,7 @@ default-extensions:
   - NamedFieldPuns
   - NoImplicitPrelude
   - NoMonomorphismRestriction
+  - NumericUnderscores
```
expede: This allows us to write numbers with underscores, like `123_456_789`

PR #248, opened 2020-01-07
Add tests

icidasset:
Adds some doctests (unit and property tests) and adds some basic setup for unit tests using `hspec`. Had written some instances as well, but I just removed them all because of all the refactoring going on the past week. Will continue working on that after this PR.

@dholms Is there any other testing stuff you need set up?

### DIFF
--- a/library/Fission/Internal/UTF8.hs
+++ b/library/Fission/Internal/UTF8.hs
@@ -18,6 +18,16 @@ import qualified RIO.ByteString      as Strict
 import qualified RIO.ByteString.Lazy as Lazy
 import qualified RIO.Text            as Text
 
+
+-- Property testing
+--
+-- $setup
+-- >>> :set -XOverloadedStrings
+-- >>> import Test.QuickCheck
+-- >>> import Test.QuickCheck.Instances ()
+-- >>> import qualified Data.Text as Text
+-- >>> import qualified RIO.ByteString.Lazy as Lazy
+
 class Textable a where
   encode :: a -> Either UnicodeException Text
 
@@ -33,14 +43,41 @@ showLazyBS = textToLazyBS . textDisplay . displayShow
 textToLazyBS :: Text -> Lazy.ByteString
 textToLazyBS = Lazy.fromStrict . Text.encodeUtf8
 
+{-| Strip one newline character from the end of a lazy `ByteString`.
+
+    >>> stripNewline ";)\n"
+    ";)"
+
+    >>> stripNewline "<>\n\n"
+    "<>\n"
+
+    prop> stripNewline (Lazy.append bs "\n") == bs
+    prop> stripNewline (Lazy.append bs "\n\n") == bs <> "\n"
+
+-}
 stripNewline :: Lazy.ByteString -> Lazy.ByteString
 stripNewline bs = bs
                |> Lazy.stripSuffix "\n"
                |> fromMaybe bs
 
+{-| Show text.
+
+    >>> textShow 1
+    "1"
+
+-}
 textShow :: Show a => a -> Text
 textShow = textDisplay . displayShow
 
+{-| Remove a number of characters from the beginning and the end of a lazy `ByteString`.
+
+    >>> stripNBS 3 "aaabccc"
+    "b"
+
+    >>> stripNBS 0 "b"
+    "b"
+
+-}
 stripNBS :: Natural -> Lazy.ByteString -> Lazy.ByteString
 stripNBS n bs = bs
              |> Lazy.take ((Lazy.length bs) - i)
@@ -49,6 +86,17 @@ stripNBS n bs = bs
     i :: Int64
     i = fromIntegral n
 
+{-| Remove a number of characters from the beginning and the end of some text.
+
+    >>> stripN 3 "aaabccc"
+    "b"
+
+    >>> stripN 0 "b"
+    "b"
+
+    prop> stripN n (Text.center (3 + fromIntegral n * 2) '_' "o.O") == "o.O"
+
+-}
 stripN :: Natural -> Text -> Text
 stripN n = Text.dropEnd i . Text.drop i
   where
@@ -63,5 +111,12 @@ putText = Strict.putStr . encodeUtf8
 putTextLn :: MonadIO m => Text -> m ()
 putTextLn txt = putText <| txt <> "\n"
 
+{-| Wrap text with some other piece of text.
+
+    prop> Text.head (wrapIn "|" s) == '|'
+    prop> Text.last (wrapIn "|" s) == '|'
+    prop> Text.length (wrapIn "|" s) == (Text.length s) + 2
+
+-}
 wrapIn :: Text -> Text -> Text
-wrapIn txt wrapper = wrapper <> txt <> wrapper
+wrapIn wrapper txt = wrapper <> txt <> wrapper
--- a/library/Fission/Platform/Heroku/Provision/Types.hs
+++ b/library/Fission/Platform/Heroku/Provision/Types.hs
@@ -47,6 +47,16 @@ data Provision = Provision
              , Show
              )
 
+instance FromJSON Provision where
+  parseJSON = withObject "Provision" \obj -> do
+    id       <- obj .: "id"
+    config   <- obj .: "config"
+    message  <- obj .: "message"
+
+    let peers = []
+
+    return Provision {..}
+
 instance ToJSON Provision where
   toJSON Provision {..} = object
     [ "id"      .= id
--- a/library/Fission/URL/Subdomain.hs
+++ b/library/Fission/URL/Subdomain.hs
@@ -9,11 +9,20 @@ import Fission.Prelude
 import Fission.URL.Subdomain.Types
 import Fission.URL.DomainName.Types
 
--- | Prefix a domain named with an optional subdomain
+{-| Prefix a domain named with an optional subdomain.
+-}
 normalizePrefix :: DomainName -> Maybe Subdomain -> DomainName
 normalizePrefix domain (Just subdomain) = prefix domain subdomain
 normalizePrefix domain Nothing          = domain
 
--- | Prefix a domain with a subdomain
+{-| Prefix a domain with a subdomain.
+
+    $setup
+    >>> :set -XOverloadedStrings
+
+    >>> prefix (DomainName "domain") (Subdomain "sub")
+    DomainName {getDomainName = "sub.domain"}
+
+-}
 prefix :: DomainName -> Subdomain -> DomainName
 prefix (DomainName domain) (Subdomain subdomain) = DomainName (subdomain <> "." <> domain)
--- a/library/Fission/User/Password.hs
+++ b/library/Fission/User/Password.hs
@@ -4,7 +4,9 @@ import           Fission.Prelude
 import           Fission.Random              as Random
 import qualified Fission.User.Password.Types as User
 
+{-| Generate a password for a `User`.
+-}
 random :: MonadIO m => m User.Password
 random = do
   pass <- Random.alphaNum 50
-  return <| User.Password pass
+  return (User.Password pass)
--- a/test/testsuite/Main.hs
+++ b/test/testsuite/Main.hs
@@ -1,87 +1,21 @@
-{-# LANGUAGE NoImplicitPrelude #-}
-
--- {-# LANGUAGE QuasiQuotes #-}
-
 module Main (main) where
 
-import RIO
+import Test.Tasty
 
-import Test.Tasty                   (TestTree, defaultMainWithIngredients, testGroup)
-import Test.Tasty.HUnit             (Assertion, testCase, (@?=))
-import Test.Tasty.Ingredients.Rerun (rerunningTests)
-import Test.Tasty.Runners           (consoleTestReporter, listingTests)
-import Test.Tasty.SmallCheck        (testProperty)
+-- ⚛
 
--- import Test.Hspec
--- import Test.Hspec.Wai
--- import Test.Hspec.Wai.JSON
+import Fission.Prelude
 
--- import Network.Wai.Handler.Warp (run)
+-- Specs
 
--- main :: IO ()
--- main = return ()
+import qualified Test.Random
 
--- main = do
---   -- port <- lookupSetting "PORT" 8081
---   -- pool <- makePool Test
-
---   -- runSqlPool doMigrate pool
-
---   -- let logger = setLogger Test
---   --     config = Config { getPool = pool
---   --                     , getEnv  = Test
---   --                     }
-
---   putStrLn "Testing..."
---   hspec . spec . run port . logger $ app config
-
--- spec :: IO () -> Spec
--- spec server = with server do
---   describe "GET /ping" $ do
---     it "responds with 200" do
---       1 `shouldBe` 1
---       -- get "/users" `shouldRespondWith` 200
-
---         -- it "responds with 'Simple'" do
---         --     get "/" `shouldRespondWith` "Simple"
-
--- config = Config { getPool = , getEnv = Test }
-
--- main :: IO ()
--- main = return ()
 
 main :: IO ()
-main =
-  defaultMainWithIngredients
-    [ rerunningTests [listingTests, consoleTestReporter] ]
-    (testGroup "all-tests" tests)
+main = defaultMain tests
 
-tests :: [TestTree]
-tests =
-  [ testGroup "SmallCheck" scTests
-  -- , testGroup "Unit tests" huTests
+
+tests :: TestTree
+tests = testGroup "Tests"
+  [ Test.Random.tests
   ]
-
-scTests :: [TestTree]
-scTests =
-  [ testProperty "inc == succ" prop_succ
-  , testProperty "inc . negate == negate . pred" prop_pred
-  ]
-
--- huTests :: [TestTree]
--- huTests =
---   [ testCase "Increment below TheAnswer" case_inc_below
---   , testCase "Decrement above TheAnswer" case_dec_above
---   ]
-
-prop_succ :: Int -> Bool
-prop_succ n = 1 + n == 1 + n
-
-prop_pred :: Int -> Bool
-prop_pred n = 1 + (negate n) == negate (n - 1)
-
--- case_inc_below :: Assertion
--- case_inc_below = inc 41 @?= (42 :: Int)
-
--- case_dec_above :: Assertion
--- case_dec_above = negate (inc (negate 43)) @?= (42 :: Int)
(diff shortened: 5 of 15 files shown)


icidasset:
Not sure what's causing the doc tests to fail, they worked some time ago 🤔  It's no longer recognizing `"string"` as something with `IsString`, but just `[Char]` instead even though I imported `RIO`.

icidasset:
One small issue left, which is that `stack test :fission-test --fast` complains about a `fission-test.tix` file which is automatically generated. If you remove the file it runs the tests properly, but then it generates that file again. Not sure why it does that.

expede:
IIRC, `.tix` is a code coverage file. We can set the tests to delete it on every run, but yeah weird that it's giving you grief. I'll give it a look.

library/Fission/Internal/UTF8.hs
```diff
@@ -18,6 +18,21 @@ import qualified RIO.ByteString      as Strict
...
+-- $setup
+-- >>> :set -XOverloadedStrings
+-- >>> import Test.QuickCheck
+-- >>> import Test.QuickCheck.Instances ()
+-- >>> import qualified Data.Text as Text
+-- >>> import qualified RIO.ByteString.Lazy as Lazy
+
+
+
+-- Instances
```
expede: Soooo two things:

1. `class` is a class, not an instance
2. Do we need to announce every section? To me this needlessly wordy to me, akin to:

```haskell
do
  emailUser user -- emails user
  updateDB emailed True -- update DB that user was emailed
```

Given that we have syntax to announce that these are classes, instances, and functions, why not just:

```haskell
-- Property testing
--
-- $setup
-- >>> :set -XOverloadedStrings
-- >>> import Test.QuickCheck
-- >>> import Test.QuickCheck.Instances ()
-- >>> import qualified Data.Text as Text
-- >>> import qualified RIO.ByteString.Lazy as Lazy

class Textable a where
  encode :: a -> Either UnicodeException Text
  encode :: a -> Either UnicodeException Text

instance Textable ByteString where
  encode = decodeUtf8'

instance Textable Lazy.ByteString where
  encode = encode . Lazy.toStrict

showLazyBS :: Show a => a -> Lazy.ByteString
showLazyBS = textToLazyBS . textDisplay . displayShow
```

...or break the the class & instances into their own module?

Thoughts?

library/Fission/Internal/UTF8.hs
```diff
@@ -27,20 +42,51 @@ instance Textable ByteString where
...
 
+{-| Strip one newline character from the end of a lazy `ByteString`.
+
+    >>> stripNewline ";)\n"
+    ";)"
+
+    >>> stripNewline "<>\n\n"
+    "<>\n"
+
+    prop> stripNewline (Lazy.append bs "\n") == bs
```
expede: Maybe another prop test that it doesn't do multiple newlines, too:

```haskell
prop> stripNewline (Lazy.append bs "\n\n") /= bs
```

library/Fission/Internal/UTF8.hs
```diff
@@ -49,6 +95,15 @@ stripNBS n bs = bs
...
     i = fromIntegral n
 
+{-| Remove a number of characters from the beginning and the end of some text.
+
+    >>> stripN 3 "aaabccc"
+    "b"
+
+    >>> stripN 0 "b"
+    "b"
+
```
expede: Could add a prop that the length changes by `n` (or goes to 0)

library/Fission/Internal/UTF8.hs
```diff
@@ -63,5 +118,12 @@ putText = Strict.putStr . encodeUtf8
 putTextLn :: MonadIO m => Text -> m ()
 putTextLn txt = putText <| txt <> "\n"
 
+{-| Wrap text with some other piece of text.
+
+    prop> Text.head (wrapIn "|" s) == '|'
+    prop> Text.last (wrapIn "|" s) == '|'
+    prop> Text.length (wrapIn "|" s) == (Text.length s) + 2
```
expede: 👍

library/Fission/Platform/Heroku/Provision/Types.hs
```diff
@@ -47,6 +47,16 @@ data Provision = Provision
...
 
+instance FromJSON Provision where
+  parseJSON = withObject "Provision" \obj -> do
+    id       <- obj .: "id"
+    config   <- obj .: "config"
+    message  <- obj .: "message"
+
+    let peers = []
+
+    return Provision {..}
```
expede: Nice 👍

expede:
> stack test :fission-test --fast

I like it! We should add this to the Makefile, too

library/Fission/Internal/UTF8.hs
expede: 🤔 Where doctests really shine is in story telling:

```haskell
{-| Strip one newline character from the end of a lazy `ByteString`
      
    It removes a newline from the end of strings:  
  
    >>> stripNewline ";)\n"
    ";)"

    prop> stripNewline (Lazy.append bs "\n") == bs

    But only does a single newline:

    >>> stripNewline "<>\n\n"
    "<>\n"

    prop> stripNewline (Lazy.append bs "\n\n") == (bs <> "\n")

    It also doesn't touch other types of whitespace:

    prop> stripNewline (Lazy.append bs "\r") == (bs <> "\r")

    Or newlines at the beginning of the string:

    prop> stripNewline (Lazy.append "\n" bs) == ("\n" <> bs)
-}
```

expede:
```shell
    The import of ‘Fission.Prelude’ is redundant
      except perhaps to import instances from ‘Fission.Prelude’
    To import instances alone, use: import Fission.Prelude()
  |
8 | import           Fission.Prelude
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

library/Fission/Internal/UTF8.hs
expede: Actually, you know what I take that back. These don't really add to the _docs_, but are good test cases for the actual test suite. Leaving here instead of deleting because... sure why not.

expede:
> complains about a fission-test.tix 

🤔 I'm not getting this warning

expede:
> Finished in 0.0005 seconds

🤯

library/Fission/Internal/UTF8.hs
icidasset: So, I like to divvy things up that feel separate, that's why I leave so much space between those things. I like how I have a better visual overview when I do this.

That said, it doesn't make sense to me to create new files for stuff that's so small. I only like to separate things into multiple files if one file is over a 1000 lines long or more.

It's a style I adopted from Elm.
I can use your style if this bothers you?

library/Fission/Internal/UTF8.hs
icidasset: Will add the property test 👍 
This type of story telling could be great for more complicated functions though?

icidasset:
Ok, somehow I don't have the issue anymore with the `.tix` file, weird 😅

icidasset:
Decided to remove the `-fhpc` ghc flag which generated the `tix` file. Is that ok @expede ? 

By the way, to reproduce the issue I had, I had to cause a spec to fail. Issue doesn't occur when all tests are passing.

Also added a few more small tests.
Any thoughts on what I should test next?
I'm not sure where to go from here.

expede:
@icidasset 

> Decided to remove the -fhpc ghc flag which generated the tix file. Is that ok @expede ?

Without that flag we can't do code coverage reports. Why not change the Makefile:

```shell
test:
	$(stack) build --test --fast $(package)

test-unit: # <- this one
    rm *.tix && $(stack) test :fission-test --fast
```

icidasset:
@expede I thought we could set the flag (ghc-option) per "package" under `tests:` in `package.yml`. So we could set it for `fission-coverage-code` which is commented out right now, but remove it from `fission-test`. Or am I forgetting something?

library/Fission/Internal/UTF8.hs
expede: >  it doesn't make sense to me to create new files for stuff that's so small

It's small now, but can grow over time.

> It's a style I adopted from Elm.

Haskell and Elm have different features. Your right that in Elm you absolutely need comments to break apart the sections of your code, but that's a consequence of Elm having fewer options to structure your application. You have a bunch of `map` and `andThen` pseudo-instances, for example, and probably want those all labelled as such. Haskell has more options for abstraction, and so also more keywords like `instance`, so we don't need another label announcing that it's an instance.

> I only like to separate things into multiple files if one file is over a 1000 lines long or more.

Again, very different trade offs between languages. Purely on style 1000 LOC seems pretty steep to me in any language — a file over _200 lines_ is acceptable but should be rare.

For broken down modules, it depends on if you want consistency or not. If we have all of our stuff broken into `*.Types`, it makes more sense than sometimes having them and sometimes not. Think of this as being cut the other way like in Phoenix: you have a directory called "controllers" where all of your short controllers live, and another called "views" where your short views live. We're slicing the stack the other way, from domain down.

Splitting out `.Types` is important because it helps prevent circular dependencies. Splitting out `.Class` for the `class` and instances for pre-existing types helps get rid of a bunch of noise in the module.

library/Fission/Internal/UTF8.hs
expede: > could be great for more complicated functions though?

Yeah, I think that it's great in general, and is how I tend to write doctests. I also realize that it can make the docs pretty noisy, so it's a balance.

expede:
@icidasset it looks like it gathers that information from the test runs themselves

https://wiki.haskell.org/Haskell_program_coverage

But I also don't know for sure if we won't get that in the coverage run. Sure, let's more it there for now and change if/when we run into issues 👍

library/Fission/Internal/UTF8.hs
icidasset: > Yes please 

You got it, will change 👍 

> It's small now, but can grow over time.

True, that said, I'm more of the "split it up when needed" kind of guy.

> Purely on style 1000 LOC seems pretty steep to me in any language

Yeah, I get that a lot. Doesn't bother me much really 🤷‍♂ 
My reasoning is more like:
- Say you would have a `Text` module and everything related to text is in that module
- I would just use my editor to navigate between all the functions and types. I can easily browse through all the functions and types using autocomplete.
- If things are in separate files, I have to do more effort to look for things.

But yeah, totally personal preferences.
Will change 😉

expede: @icidasset thanks! And extra thanks for explaining your reasoning and engaging in the conversation in a strong-opinions-loosely-held way 🎉🎉🎉

icidasset:
@expede Updated. Let me know if this works for you. The only place I left those dividing comments in, is in the `Spec.Common` module, where it feels appropriate.

test/testsuite/Spec/Common.hs
```diff
@@ -0,0 +1,48 @@
...
+  --
+  , userInsertedAt = agesAgo
+  , userModifiedAt = agesAgo
+  }
+
+testUserEntity :: Database.Entity User
+testUserEntity = Database.Entity
+  { entityKey = Database.toSqlKey 0
+  , entityVal = testUser
+  }
```
expede: @icidasset yup these fixtures work for now 👌 

In general, fixtures are pretty brittle / only handle the one case. This can lead to sub-optimal tests that rely on these specific cases. This is equally true for Elixir, Ruby, JS, and so on. In those languages I generally recommend starting with a using a faker of some kind. Haskell has one: https://www.stackage.org/lts-14.20/package/fakedata-0.2.2.

However, since we're also going to be prop testing, we probably want to have them as instances of QuickCheck's `Arbitrary` typeclass. Here's an example:

```haskell
newtype Prime a = Prime a deriving Show

instance (Integral a, Arbitrary a) => Arbitrary (Prime a) where
  arbitrary = do
    x <- frequency [ (10, choose (0, 1000))
                   , (5, choose (1001, 10000))
                   , (1, choose (10001, 50000))
                   ]
    return $ Prime (primes !! x)

-- Random value outside of QuickCheck:

randomPrime :: Prime <- generate
```

Let's merge this for now, though.

test/testsuite/Spec/Random.hs
```diff
@@ -0,0 +1,18 @@
+module Spec.Random (spec) where
+
+import qualified RIO.ByteString as BS
+import           Test.Hspec as Hspec
+
+-- ⚛
```
expede: What does this symbol mean?

test/testsuite/Spec/Random.hs
```diff
@@ -0,0 +1,18 @@
...
+
+import           Fission.Prelude
+import qualified Fission.Random as Random
+
+-- 👩‍🔬
```
expede: Same here 🤔

test/testsuite/Spec/Common.hs
```diff
@@ -0,0 +1,48 @@
...
+import           Fission.Models
+import qualified Fission.User.Role.Types as User.Roles
+import           Fission.Prelude
+
+
+-- TIME
+
+
+agesAgo :: UTCTime
+agesAgo = UTCTime
```
expede: 😆 cute naming

test/testsuite/Main.hs
```diff
@@ -1,87 +1,24 @@
...
-prop_pred :: Int -> Bool
-prop_pred n = 1 + (negate n) == negate (n - 1)
-
--- case_inc_below :: Assertion
--- case_inc_below = inc 41 @?= (42 :: Int)
-
--- case_dec_above :: Assertion
--- case_dec_above = negate (inc (negate 43)) @?= (42 :: Int)
+main = Hspec.hspec do
+  -- Specs
```
expede: I mean... same comment as last time. Yeah, these are specs. Do we need to announce the fact, give that they're literally prefixed with `Spec.` and end in `.spec`?

test/testsuite/Main.hs
```diff
@@ -1,87 +1,24 @@
...
-import Test.Tasty.Runners           (consoleTestReporter, listingTests)
-import Test.Tasty.SmallCheck        (testProperty)
-
--- import Test.Hspec
--- import Test.Hspec.Wai
--- import Test.Hspec.Wai.JSON
+-- Hspec
 
--- import Network.Wai.Handler.Warp (run)
+import qualified Test.Hspec as Hspec
```
expede: Why not `Tasty.Hspec`?

expede: https://documentup.com/feuerbach/tasty

> It lets you combine your unit tests, golden tests, QuickCheck/SmallCheck
properties, and any other types of tests into a single test suite.

library/Fission/Web/DNS.hs
```diff
@@ -18,7 +19,10 @@ type API = Capture "cid" CID
         :> PutAccepted '[PlainText, OctetStream] URL.DomainName
 
 server :: MonadDNSLink m => Entity User -> ServerT API m
-server (Entity _id User { userUsername }) cid =
+server = updateLink
+
+updateLink :: MonadDNSLink m => Entity User -> CID -> m URL.DomainName
+updateLink (Entity _id User { userUsername }) cid =
```
expede: Why?

library/Fission/User/Password.hs
```diff
@@ -4,7 +4,9 @@ import           Fission.Prelude
 import           Fission.Random              as Random
 import qualified Fission.User.Password.Types as User
 
+{-| Generate a password for a `User`.
+-}
```
expede: Not a fan of `-- |`?

library/Fission/URL/Subdomain.hs
```diff
@@ -9,11 +9,20 @@ import Fission.Prelude
...
 normalizePrefix domain Nothing          = domain
 
--- | Prefix a domain with a subdomain
+{-| Prefix a domain with a subdomain.
+
+    $setup
+    >>> :set -XOverloadedStrings
+
+    >>> prefix (DomainName "domain") (Subdomain "sub")
+    DomainName {getDomainName = "sub.domain"}
```
expede: 👍

test/testsuite/Spec/Random.hs
icidasset: It's the atom symbol, under which I place all the Fission stuff.

test/testsuite/Main.hs
icidasset: Ah yeah for sure, forgot to remove this one. Old version had a bunch more stuff in this function. Will remove.

expede:
I can't comment on them because of how GH works, but why did the Postgres stuff get moved into the library? That change going imposes a Postgres dependency on the CLI.

test/testsuite/Main.hs
icidasset: Ah right forgot about that. I started with following the Servant guide on testing and went from there. Eventually we removed the Servant stuff, so I was left with this `Test.Hspec`, forgot all about that `tasty` library.

library/Fission/Web/DNS.hs
icidasset: I assumed we wanted the specs to be independent of Servant, hence this code. Didn't go through with the specs after all the refactoring.

library/Fission/User/Password.hs
icidasset: Doesn't bother me. I started writing a doc test for this and ended up removing it, so that's why it looks like this.

library/Fission/Web/DNS.hs
expede: I see; gotcha.

`ServerT api` is just a fancy type alias. No need to break them out into separate functions 👍

test/testsuite/Spec/Random.hs
expede: Ooooooooh haha I see. Fission... atomic... yeah cool cool 👍😆

I like your puns

icidasset:
@expede Which postgres stuff? 🤔 I'm not seeing anything related to Postgres.

test/testsuite/Spec/Random.hs
expede: Oooh, okay, so this is the label for "tests", because it's an experimenalist running tests 😛 

My comment about needless naming still applies, but this is just _so cute_ that sure let's leave it for now

expede:
> @expede Which postgres stuff? 🤔 I'm not seeing anything related to Postgres.

Screen Shot 2020-01-08 at 1 53 39 PM

test/testsuite/Spec/Random.hs
icidasset: Just removed it ;p

library/Fission/Web/DNS.hs
icidasset: Just put it back as before.

icidasset:
@expede Oh riiight, need to put that back 👀

test/testsuite/Main.hs
icidasset: Will replace with Tasty tomorrow.

icidasset:
@expede Added tasty 👍

PR #249, opened 2020-01-08
Convert DB Queries and Mutations to use a monad

bechurch:
Broken out from #246 
See comments starting here: https://github.com/fission-suite/web-api/pull/246#discussion_r363557432

## Summary
Updates Queries and Mutations to use a Monad that will allow for easier stubbing during testing.

## Closing issues

Fixes #

## After Merge

### DIFF
--- a/library/Fission/Storage/Mutation/Class.hs
+++ b/library/Fission/Storage/Mutation/Class.hs
@@ -15,6 +15,8 @@ deleteWhereGeneric ::
   => (SqlExpr (Entity model) -> SqlExpr (Value Bool))
   -> Transaction m ()
 deleteWhereGeneric whereClause = delete <| from \model -> where_ (whereClause model)
+-- Could also be
+-- deleteWhereGeneric = Esqueleto.deleteWhere
 
 class (MonadDB m, PersistEntity model, PersistRecordBackend model SqlBackend) => MonadDBMutation model m where
   deleteWhere :: (SqlExpr (Entity model) -> SqlExpr (Value Bool)) -> Transaction m ()
@@ -22,5 +24,7 @@ class (MonadDB m, PersistEntity model, PersistRecordBackend model SqlBackend) =>
 instance MonadDBMutation User Fission where
   deleteWhere = deleteWhereGeneric
 
+instance MonadDBMutation User Fission where
+  deleteWhere = deleteWhereGeneric
 
 
--- a/library/Fission/Storage/Mutation/Class.hs
+++ b/library/Fission/Storage/Mutation/Class.hs
@@ -1,0 +1,19 @@
+module Fission.Storage.Mutation.Class (MonadDBMutation (..)) where
+
+import           Network.IPFS.CID.Types
+import           Database.Esqueleto hiding ((=.), update)
+
+import           Fission.Models
+import           Fission.Prelude hiding (Value)
+import           Fission.Types
+
+deleteWhereGeneric whereClause = delete <| from \model -> where_ (whereClause model)
+
+class MonadDB m => MonadDBMutation model m where -- Note that this is Monad not MonadDB
+  deleteWhere :: (SqlExpr (Entity model) -> SqlExpr (Value Bool)) -> Transaction m ()
+
+instance MonadDBMutation User Fission where
+  deleteWhere = deleteWhereGeneric
+
+
+
--- a/library/Fission/Storage/Mutation/Class.hs
+++ b/library/Fission/Storage/Mutation/Class.hs
@@ -7,9 +7,16 @@ import           Fission.Models
 import           Fission.Prelude hiding (Value)
 import           Fission.Types
 
+deleteWhereGeneric ::
+  ( PersistEntity model
+  , PersistRecordBackend model SqlBackend
+  , MonadDB m
+  )
+  => (SqlExpr (Entity model) -> SqlExpr (Value Bool))
+  -> Transaction m ()
 deleteWhereGeneric whereClause = delete <| from \model -> where_ (whereClause model)
 
-class MonadDB m => MonadDBMutation model m where -- Note that this is Monad not MonadDB
+class (MonadDB m, PersistEntity model, PersistRecordBackend model SqlBackend) => MonadDBMutation model m where
   deleteWhere :: (SqlExpr (Entity model) -> SqlExpr (Value Bool)) -> Transaction m ()
 
 instance MonadDBMutation User Fission where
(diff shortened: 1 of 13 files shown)


library/Fission/Storage/Query/Class.hs
```diff
@@ -0,0 +1,34 @@
+module Fission.Storage.Query.Class (MonadDBQuery (..)) where
+
+import           Network.IPFS.CID.Types
+import           Database.Esqueleto hiding ((=.), update)
+
+import           Fission.Models
+import           Fission.Prelude hiding (Value)
+import           Fission.Types
+
+-- getByGeneric :: (SqlExpr (Entity model) -> SqlExpr (Value Bool)) -> Transaction m [Entity model]
```
bechurch: I tried defining a type for these functions however it seemed to rely on having access to the internals of esquelito. @expede Any ideas here?

This was the type my editor kept suggesting
```haskell
getByGeneric :: (esqueleto-3.0.0:Database.Esqueleto.Internal.Internal.SqlSelect a r, MonadIO m, BackendCompatible SqlBackend backend, PersistQueryRead backend, PersistUniqueRead backend, From a) => (a -> SqlExpr (Value Bool)) -> ReaderT backend m [r]
```

expede: Yeah, we can almost certainly scope this to our `Transaction` type (which only works on Postgres, but that satisfies the constraints). I'll pull it up in emacs, one sec...

library/Fission/Storage/Mutation/Class.hs
```diff
@@ -0,0 +1,19 @@
...
+import           Fission.Prelude hiding (Value)
+import           Fission.Types
+
+deleteWhereGeneric whereClause = delete <| from \model -> where_ (whereClause model)
+
+class MonadDB m => MonadDBMutation model m where -- Note that this is Monad not MonadDB
+  deleteWhere :: (SqlExpr (Entity model) -> SqlExpr (Value Bool)) -> Transaction m ()
+
+instance MonadDBMutation User Fission where
+  deleteWhere = deleteWhereGeneric
```
expede: You can stick this right in the default implementation, too. Then you don't need to write special instances 👍

library/Fission/Storage/Query/Class.hs
expede:
```haskell
getByGeneric :: 
  ( PersistEntity model
  , BackendCompatible SqlBackend (PersistEntityBackend model)
  , MonadIO m
  ) 
  => (SqlExpr (Entity model) -> SqlExpr (Value Bool))
  -> Transaction m [Entity model]
getByGeneric whereClause = 
  select <| from \model -> do
    where_ (whereClause model)
    return model
```

library/Fission/Storage/Mutation/Class.hs
expede: Okay, so these need some tweaks to allow that to happen, but this is all still good:

```haskell
class 
  ( MonadIO m
  , PersistEntity model
  , BackendCompatible SqlBackend (PersistEntityBackend model)
  ) 
  => MonadDBMutation model m where
    deleteWhere :: (SqlExpr (Entity model) -> SqlExpr (Value Bool)) -> Transaction m 
    deleteWhere whereClause = delete <| from \model -> where_ (whereClause model)

class 
  ( MonadIO m
  , PersistEntity model
  , BackendCompatible SqlBackend (PersistEntityBackend model)
  ) 
  => MonadDBQuery model m where
    getBy :: (SqlExpr (Entity model) -> SqlExpr (Value Bool)) -> Transaction m 
    getBy whereClause =
      select <| from \model -> do
        where_ (whereClause model)
        return model
```

expede: This now says "any model (that has an `instance PersistModel model`, which they all do) can be used with this interface _automatically_ with this default function. If you want to overwrite it (e.g. in tests) then write an `instance`, but otherwise don't worry about it".

library/Fission/Storage/Query/Class.hs
expede: See the other comment below for fancy protip ❇️

expede:
Also, aside from the protips above, this is looking really good 👍

library/Fission/User/Query.hs
```diff
@@ -8,16 +8,15 @@ import           Database.Esqueleto hiding ((=.), update)
...
-getHerkouAddonByUserId addOnId = select <| from \user -> do
-      where_ <| user ^. UserHerokuAddOnId ==. val (Just addOnId)
-            &&. user ^. UserActive        ==. val True
-      limit 1
-      return user
+getHerkouAddonByUserId :: MonadDBQuery User m => HerokuAddOnId -> Transaction m [Entity User]
+getHerkouAddonByUserId addOnId =
+      getOneBy (\user ->
+                  user ^. UserHerokuAddOnId ==. val (Just addOnId)
+                  &&. user ^. UserActive        ==. val True)
```
expede: Protip: no need for the parens

```haskell
getOneBy \user ->
      user ^. UserHerokuAddOnId ==. val (Just addOnId)
  &&. user ^. UserActive        ==. val True
```

library/Fission/User/Mutation.hs
```diff
@@ -100,9 +101,8 @@ hashPassword' password = do
     Nothing           -> Left Error.FailedDigest
     Just secretDigest -> Right <| decodeUtf8Lenient secretDigest
 
-destroy :: MonadDB m => UserId -> Transaction m ()
-destroy userId = delete <| from \user ->
-    where_ (user ^. UserId ==. val userId)
+destroy :: MonadDBMutation User m => UserId -> Transaction m ()
+destroy userId = deleteWhere (\user -> user ^. UserId ==. val userId)
```
expede: Also don't need the parens here:

```haskell
deleteWhere \user -> user ^. UserId ==. val userId
```

---

You know, if you wanted the simpler-style syntax, we could write something like:

```haskell
deleteWhere' matchOne = deleteWhere \model -> model ^. matchOne

-- Then:
destroy userId = deleteWhere (UserId ==. val userId)
```

Or even:

```haskell
deleteWhere'' andList = deleteWhere \model ->
  andList
    |> fmap (model ^.)
    |> foldr1 (&&.)

-- Use:

deleteWhere'' [UserId ==. val userId, UserEmail ==. val email. UserAge ==. val 42]

-- Same as:

deleteWhere \model ->
      model ^. UserId    ==. val UserId
  &&. model ^. UserEmail ==. val email
  &&. model ^. UserAge   ==. val 42
```

bechurch:
Closed due to starting fresh in a new PR after the first approach didnt pan out
https://github.com/fission-suite/web-api/pull/251

PR #251, opened 2020-01-10
Convert DB Queries and Mutations to use a monad

bechurch:
Broken out from #246 and #249 
See comments starting here: https://github.com/fission-suite/web-api/pull/246#discussion_r363557432

## Summary
Updates Queries and Mutations to use a Monad that will allow for easier stubbing during testing.

## Closing issues

Fixes #

## After Merge

### DIFF
--- a/library/Fission/Types.hs
+++ b/library/Fission/Types.hs
@@ -45,6 +45,12 @@ import qualified Fission.User.CID as User.CID
 
 -- | The top-level app type
 newtype Fission a = Fission { unwrapFission :: RIO Config a }
+  deriving anyclass ( User.MonadDBQuery
+                    , User.MonadDBMutation
+                    , User.CID.MonadDBQuery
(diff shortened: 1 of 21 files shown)


library/Fission/Types.hs
```diff
@@ -191,3 +198,32 @@ instance MonadRemoteIPFS Fission where
...
+
+instance User.CID.MonadDBQuery Fission where
+  getByUserId = User.CID.getByUserId
+  getByCids = User.CID.getByCids
+
+instance User.CID.MonadDBMutation Fission where
+  create = User.CID.create
+  createX = User.CID.createX
+  destroyAll = User.CID.destroyAll
+  destroyExact = User.CID.destroyExact
```
bechurch: @expede I wasn't able to had `Fission` default to using these base implementations without specifying the instance.

Heres the error I was receiving below.
image

Any ideas?

PR #252, opened 2020-01-13
DB Access Classes & Errors

expede:
A big thanks to Ben for getting this most of the way done. I needed to unwind this weekend, so I refactored it a bit more, too.

# Summary 

Break out all DB logic into type classes. This helps us test the DB logic, and give us really nice composability of these effects. We also decouple from it _having_ to come from the DB, so we can (for instance) get data from network calls, microservices, hardcoded values, and whatnot, without changing the biz logic of the code (e.g. no need to change the web handlers).

# Naming

We don't actually need to prefix everything with `Monad` :P In fact, none of the classes in this PR actually _have_ to be monads, but we use them monadically so frequently that it's annoying to have to specify in every function, so they do all inherit the `Monad` constraint.

Why not name them after CRUD actions? They kind of are, but a different convention from "create, read, update, destroy". The main problem is that `Read` is an existing, widely used typeclass from `base`. So they're all following the pattern "create, retrieve, modify, destroy", or `Foo.Creator`, `Foo.Retriever`, `Foo.Modifier`, and `Foo.Destroyer`.

# `OpenUnion`

I found a library that implements the same stuff as that `oops` repo! It's not the fastest of these implementations, but fast enough, used in other packages (like `servant-checked-exceptions`), and pretty well documented (examples, explanations, &c).

Okay, sure, cool whatever. Why is this useful to us? Well, we can now very easily merge errors from various parts of the codebase. Additionally, if each error has a `ToServerError err` instance, then you can just pass the whole thing to `Web.Err.ensure` or `Web.Err.throw` and it'll handle extracting and throwing the error itself. I've pointed to where it gets used in the code in a PR comment, but in the meantime, some background:

This uses some type-level magic underneath, but essentially it's an `Either` with a dynamic number of cases. These express the same idea:

```haskell
Either a b
OpenUnion '[a, b]
```

But with `OpenUnion`, we can also have this:

```haskell
OpenUnion '[a, b, c, d, e, f]

-- same as

data CustomErr a b c d e f
  | FooErr a
  | BarErr b
  | BazErr c
  | QuuxErr d
  | QuuxierErr e 
  | QuuxiestErr f
```

It's like an "anonymous union" of types. TypeScript has these, too:

```typescript
number | string | MyObj
```

```haskell
OpenUnion '[Int, Text, MyObj]
```

How do you get values into an `OpenUnion`?

```haskell
foo :: OpenUnion '[String, Int]
foo = openUnionLift "hello"

bar :: OpenUnion '[String, Int]
bar = openUnionLift 42
```

How do you get values out of an `OpenUnion`? Fairly straightforward, but you need to provide handlers for all of the cases:

```haskell
-- Handler style

extractor :: OpenUnion '[String, Int] -> Int
extractor strOrInt = fromOpenUnion (\str -> length str) strOrInt

-- Catch style

catcher :: OpenUnion '[String, Int] -> String
catcher strOrInt = catchesOpenUnion (strHandler, intHandler) strOrInt
  where
    strHandler :: String -> String
    strHandler = \str -> "got a String: " ++ str

    intHandler :: Int -> String
    intHandler = \int -> "got an Int: " ++ show int
```

You can now also talk about certain kinds of subtypes:

```haskell
-- errs contains *at least* FooErr and BarErr
foo ::
  ( IsMember FooErr errs
  , IsMember BarErr errs
  )
  => Either errs Int
```

Seems pretty great, right? Why not use this everywhere?

1. It has a small performance hit because this is harder to optimize
2. `Either` is powerful enough for the 90% use case, and super well supported

Also some edges to be aware of:

```haskell
data FooErr = FooErr

OpenUnion '[FooErr] -- adds exactly nothing to the bare type other than having to put stuff into / pull out of the OpenUnion. There's no reason to do this unless you expect to add more stuff to it.

OpenUnion '[] -- Can't actually make a value of this type... because it accepts no types. You'll see this in a few classes as a base case, but that's all type level programming so don't worry about it too much.
```


app/Fission/Storage/PostgreSQL.hs
```diff
@@ -23,7 +23,7 @@ import           Fission.Models
 --
 --   > runOne (runDB updateDBToLatest)
 updateDBToLatest :: MonadIO m => Transaction m ()
-updateDBToLatest = runMigration migrateAll
+updateDBToLatest = Transaction <| runMigration migrateAll
```
expede: Probs going to put this back to a `type`

library/Fission/Platform/Heroku/AddOn/Destroyer/Class.hs
```diff
@@ -0,0 +1,15 @@
...
+
+import           Data.UUID (UUID)
+import           Database.Esqueleto
+
+import           Fission.Models
+import           Fission.Prelude
+
+-- | Destroy @Heroku.AddOn@s
+class Monad m => Destroyer m where
+  destroyByUUID :: UUID   -> m ()
```
expede: 💅

library/Fission/Types.hs
```diff
@@ -35,35 +35,57 @@ import qualified Fission.URL as URL
...
+import           Fission.Platform.Heroku.Authorizer     as Heroku
 
 import qualified Fission.Web.Auth              as Auth
 import           Fission.Web.Server.Reflective
 
+import Fission.User.Retriever as User
+import Fission.User.Authorizer.Class as User
+import           Servant
+import Fission.Models
+-- import Database.Esqueleto (Entity)
```
expede: 🔪

library/Fission/Types.hs
```diff
@@ -35,35 +35,57 @@ import qualified Fission.URL as URL
...
+  deriving newtype  ( Functor
+                    , Applicative
+                    , Monad
+                    , MonadIO
+                    , MonadUnliftIO
+                    , MonadReader Config
+                    , MonadThrow
+                    , MonadCatch
+                    , MonadMask
+                    )
```
expede: 💅

library/Fission/Types.hs
```diff
@@ -191,3 +213,46 @@ instance MonadRemoteIPFS Fission where
...
+      attemptMsg :: ByteString -> ByteString
+      attemptMsg username = "Unauthorized user! Attempted with username: " <> username
+
+instance AsHandler Fission where
+  asHandler action = Servant.Handler <| ExceptT do
+    cfg <- ask
+    action
+      |> unwrapFission
+      |> runRIO cfg
+      |> Fission.Prelude.try
```
expede: 🍴

library/Fission/User/Creator/Class.hs
```diff
@@ -0,0 +1,56 @@
...
+        Transaction do
+          insertUnique newUserRecord >>= \case
+            Just userID ->
+              return (Right userID)
+
+            Nothing ->
+              return (Left AlreadyExists)
+
+  createWithHeroku herokuUUID herokuRegion username password now = do
+    addOnId <- Transaction <| insert HerokuAddOn -- TODO EXTRACT INTO A TYPECLASS
```
expede: ☝️

library/Fission/User/Modifier/Class.hs
```diff
@@ -0,0 +1,27 @@
...
+import           Database.Persist
+
+import           Fission.Models
+import           Fission.Prelude
+
+import qualified Fission.User.Creator.Error as Create
+import           Fission.User.Password      as Password
+
+class Monad m => Modifier m where
+  updatePassword :: UserId -> Password -> UTCTime -> m (Either Create.Error Password) -- TODO wait, which error?
```
expede: ☝️

library/Fission/User/Password.hs
```diff
@@ -1,12 +1,30 @@
...
+-- | Generate a password for a `User`.
 random :: MonadIO m => m User.Password
 random = do
   pass <- Random.alphaNum 50
   return (User.Password pass)
+
+hashPassword :: MonadIO m => Text -> m (Either User.Creator.Error Text)
+hashPassword password = do
+  hashed <- liftIO <| hashPasswordUsingPolicy slowerBcryptHashingPolicy <| encodeUtf8 password
+  return <| case hashed of
```
expede: convert

library/Fission/User/Authorizer/Class.hs
```diff
@@ -0,0 +1,12 @@
...
+import           Servant
+import           Fission.Prelude
+
+import Fission.Models
+import Database.Esqueleto
+
+-- | Authorization for the Heroku Partner API
+class Monad m => Authorizer m where
+  -- | Verify that the sender is the Heroku Partner web service
+  verify :: m (BasicAuthCheck (Entity User))
```
expede: Maybe

```haskell
class Monad m => Authorizer m a where
  verify :: m (BasicAuthCheck a)
```

library/Fission/Web/Heroku/Provision.hs
```diff
@@ -42,27 +43,31 @@ create Request {uuid, region} = do
...
-        , message = "Successfully provisioned Interplanetary Fission!"
-        , config  = User.Provision
-           { username = username
-           , password = Secret secret
-           , url      = url'
-           }
-        }
+    Left err -> do
+      logError <| textShow err
+      return []
```
expede: Just easier to read this way. It was a bit harrier than it needed to be before.

library/Fission/Web/Heroku/Provision.hs
```diff
@@ -42,27 +43,31 @@ create Request {uuid, region} = do
...
+        return Provision
+          { id      = userID
+          , peers   = ipfsPeers
+          , message = "Successfully provisioned Interplanetary Fission!"
+          , config  = User.Provision
+            { username = username
+            , password = Secret secret
+            , url      = url'
+            }
+          }
```
expede: I don't love seeing...

```haskell
|> bind foo
|> bind bar
```

...but the good news about being in a pure lazy language is that the compiler _will_ rewrite this as:

```haskell
|> bind \x -> 
  foo (bar x)
```

Having a huge string of `bind...bind...bind...bind` would be annoying, so maybe let's limit it to like... 2 or 3 max?

library/Fission/Prelude.hs
```diff
@@ -6,6 +6,7 @@ module Fission.Prelude
   , module Data.Aeson
   , module Data.Has
   , module Data.Maybe
+  , module Data.WorldPeace
```
expede: Really badly named package, but this is the one that gives us `OpenUnion`

library/Fission/User/Creator/Class.hs
```diff
@@ -0,0 +1,72 @@
...
+
+import qualified Fission.User.Creator.Error as User
+import qualified Fission.User.Password      as Password
+import           Fission.User.Role.Types
+
+type Errors = OpenUnion
+  '[ User.AlreadyExists
+   , Heroku.AddOn.Error
+   , Password.FailedDigest
+   ]
```
expede: This is an example of `OpenUnion` in the wild

library/Fission/User/Creator/Class.hs
```diff
@@ -0,0 +1,72 @@
...
+
+class Heroku.AddOn.Creator m => Creator m where
+  -- | Create a new, timestamped entry with optional heroku add-on
+  create ::
+       Text
+    -> Text
+    -> Maybe Text
+    -> Maybe HerokuAddOnId
+    -> UTCTime
+    -> m (Either Errors UserId)
```
expede: The `OpenUnion` is `Errors` here

library/Fission/User/Creator/Class.hs
```diff
@@ -0,0 +1,72 @@
...
+              , userInsertedAt    = now
+              , userModifiedAt    = now
+              }
+
+        insertUnique newUserRecord >>= \case
+          Just userID ->
+            return (Right userID)
+
+          Nothing ->
+            return <| Left <| openUnionLift User.AlreadyExists
```
expede: Example of how to get something into the `OpenUnion`

library/Fission/User/Creator/Class.hs
```diff
@@ -0,0 +1,72 @@
...
+
+  createWithHeroku herokuUUID herokuRegion username password now = do
+    Heroku.AddOn.create herokuUUID herokuRegion now >>= \case
+      Right addOnId -> create username password Nothing (Just addOnId) now
+      Left  err     -> return <| Left <| openUnionLift err
```
expede: Same here

library/Fission/User/Creator.hs
```diff
@@ -0,0 +1,7 @@
+module Fission.User.Creator
+  ( module Fission.User.Creator.Class
+  , module Fission.User.Creator.Error
+  ) where
+
+import Fission.User.Creator.Class
+import Fission.User.Creator.Error
```
expede: Save us multiple imports

library/Fission/User/CID.hs
```diff
@@ -1,7 +1,9 @@
...
+  ( module Fission.User.CID.Retriever
+  , module Fission.User.CID.Creator
+  , module Fission.User.CID.Destroyer
   ) where
 
-import Fission.User.CID.Query
-import Fission.User.CID.Mutation
+import Fission.User.CID.Retriever
+import Fission.User.CID.Creator
+import Fission.User.CID.Destroyer
```
expede: Saves us from needing so many explicit imports when doing a lot of stuff with these classes.

```haskell
import qualified Fission.User.CID as User.CID

-- ...

foo :: (User.CID.Retriever m, User.CID.Destroyer m) => m Foo
```

icidasset:
Just wanted to come in here and say, totally agree with the naming 👍 and `OpenUnion` looks great! Should we put this info about OpenUnion in a documentation file in the repo, or do you think the OpenUnion docs are good enough?

I'll give this PR a proper review tomorrow, if you still need it then 👀

library/Fission/Prelude.hs
dholms: lmao Haskell community never fails to impress

library/Fission/User/Creator/Class.hs
dholms: oh yeah I dig this quite a bit

library/Fission/Web.hs
```diff
@@ -46,11 +46,11 @@ app ::
   -> Context Auth.Checks
   -> Web.Host
   -> Application
-app handlerNT auth appHost = do
+app handlerNT authChecks appHost = do
   appHost
     |> server
-    |> Auth.server      api handlerNT
-    |> serveWithContext api auth
+    |> Auth.authWithContext api handlerNT
```
dholms: :+1: to both of these naming changes

library/Fission/Web/Auth.hs
```diff
@@ -1,96 +1,47 @@
...
 
 import           Fission.Models
-import           Fission.Platform.Heroku.AddOn
+import           Fission.Platform.Heroku.Auth.Types as Heroku
 import           Fission.Prelude
+import           Fission.Web.Auth.Class
+import           Fission.Web.Auth.Types
 
-type Checks = '[BasicAuthCheck (Entity User), BasicAuthCheck ByteString]
+type Checks = '[BasicAuthCheck (Entity User), BasicAuthCheck Heroku.Auth]
```
dholms: :+1:

library/Fission/Web/Auth.hs
```diff
@@ -1,96 +1,47 @@
...
 
 -- | Construct an authorization context
 mkAuth ::
-  ( MonadHerokuAddOn m
-  , MonadDB          inner
-  , MonadLogger      inner
+  ( MonadAuth (Entity User) m
```
dholms: What's the reason for making these their own Monad? They're only used here & we weren't threading through `cfg` or anything

library/Fission/Web/Heroku/Provision.hs
dholms: makes sense to me, & I think it reads fine here

library/Fission/Web/IPFS/DAG.hs
```diff
@@ -43,5 +43,9 @@ put (Entity userId _) (Serialized rawData) =
           Web.Err.throw err
 
         Right pinnedCID -> do
-          _ <- User.CID.createX userId [newCID]
+          newCID
+            |> return
+            |> User.CID.createX userId
```
dholms: It seems like `User.CID.Creator` should have to be in the constraints in order to call this?

It seems like some places, you include them in the constraints (`Fission.Web.Heroku.Deprovision`) and other cases you don't (here)

dholms: Oh wait I think I get the difference in where they're called. This one, the Monad is `MonadDB` for running the txs. in `Heroku.Deprovision`, the constraints are used to construct the txs and then they're actually run in `MonadDB` in `destroy`.

Still seems weird that you can call `createX` without being instead a `User.CID.Creator` monad... Or are we inside one, and it's just implicit?

library/Fission/Web/IPFS/Upload/Simple.hs
```diff
@@ -15,26 +15,30 @@ import           Network.IPFS.File.Types as File
...
+  , MonadTime        m
   )
   => Entity User
   -> ServerT API m
 add (Entity userId _) (Serialized rawData) = IPFS.addRaw rawData >>= \case
   Right newCID -> IPFS.Pin.add newCID >>= \case
     Right pinnedCID -> do
-      _ <- User.CID.createX userId [pinnedCID]
+      pinnedCID
+        |> return
```
dholms: this seems like a case where `pure` would make more sense than `return`

library/Fission/Web/IPFS/DAG.hs
expede: @dholms Ah, yes, sorry I had meant to add this to the write up. This is called "contsraint plucking".

# Plucking

Just like how `runRIO` in `app/Main.hs` takes some `RIO cfg a` action, and turns it into `IO ()`, here we're taking some `Transaction m a` and turning it into `m a`. If you look at the definition of `runDB`...

```haskell
class MonadDB m where
  runDB :: Transaction m a -> m a
```

...you see that it takes something more specific (a `Transaction`) and turns it into something less specific. Now look at one of the model typeclasses:

```haskell
class User.Retriever m where
  getById :: m (Maybe User)

instance User.Retriever (Transaction m) where
  getById uid = selectFirst [UserId ==. uid]
```

We're saying that this will work on _any_ `m` (i.e. helpful for testing functions without `runDB`). But the one that we use in production is specifically a `Transaction`, which conveniently matches the input of `runDB`.

The original plan was to have each handler specify which resources it had access to. This put us in one of two places:

1. Use `runDB` in each `instance`, so we unwrap the `Transaction` per call. This loses us the database transaction and query optimizer. Not great, but lets us specify a constraint on any function in a really straightforward way. This PR does not do this.

2. Move `runDB` outside of `instances` to... somewhere else, ideally as high up in the stack as possible. But that means that everything is running inside `Transaction m`... which is actually totally fine, but means that all of our instances now need to know how to run inside a `Transaction`. This is automatic for many, but not all of them. So instead of writing a bunch of custom instances and orphans, this PR "plucks" the constraint out by giving each handler DB access and running the transaction.

We can still write helpers that constrain helpers, but at the end of the day the application has access to the DB. It's a trade-off, but IMO feels like this is the balance to strike: we can break out sensitive code (via helpers), mock everything out in tests, get access to the automatic query optimizer, and everything runs inside DB transactions, but place it in a more general / looser context to avoid a bunch of boilerplate.

# Not a Monad

> Still seems weird that you can call createX without being instead a User.CID.Creator monad... Or are we inside one, and it's just implicit?

A quick note on terminology: not all constraints are monadic, but "the monad class" is a constraint. We can run these in a concrete monad (like "List" or "Maybe") that _also_ implements the constraint `User.CID.Creator`.

> Or are we inside one, and it's just implicit?

With `User.CID.Creator`, we're inside a monad (data type) that _knows how to make `UserCid` creation DB transactions_. To actually execute those, we need to hand them to `runDB`. There's actually nothing implicitly monadic about creating a DB transaction — my first draft didn't even have `Monad m =>` on it, but we use them so frequently in monadic contexts that it was annoying to have to add an additional constraint to every function.

So, in production code, `createX` makes a `Transaction (RIO cfg)` function that will later be interpreted and executed by `runDB`. In tests, we're going to have something like:

```haskell
instance MonadDB (TestMock a) where
  runDB :: Transaction (TestMock a)
  runDB (Transaction toRun) = toRun

instance User.CID.Retriever (TestMock a) where
  getById _ = return (Just fakeUser1)
```

library/Fission/Web/Auth.hs
expede: It lets us fully mock them out in tests (and potentially local dev if we want to e.g. disable that with a flag)

library/Fission/Web/IPFS/DAG.hs
expede: Nope! Just ignore me, I think that I have a way of constraining `MonadDB` to specify the various access constraints! Whuich is great since as previously stated I'm not super thrilled with the tradeoffs as they are but it seemed like a balance. I think we can have our cake and eat it too!

This is why PRs are helpful 😅

library/Fission/Internal/MonadDB/Class.hs
```diff
@@ -2,7 +2,5 @@ module Fission.Internal.MonadDB.Class (MonadDB (..)) where
 
 import RIO
 
-import Fission.Internal.MonadDB.Types
-
-class MonadIO m => MonadDB m where
-  runDB :: Transaction m a -> m a
+class MonadIO m => MonadDB transaction m | m -> transaction where
+  runDB :: transaction a -> m a
```
expede: What the heck is that `| m -> transaction`? It says that there's only one implementation of this type class for `m`. So `Fission` will _always_ point at `Transaction`, and `TestMock` will _always_ point at `TestTransaction` (or whatever)

library/Fission/Internal/MonadDB.hs
```diff
@@ -29,16 +28,8 @@ ensureEntity err = \case
...
-ensureOneId :: (Exception err, MonadLogger m, MonadThrow m) => err -> [Entity a] -> m (Key a)
-ensureOneId err entities = pure . entityKey =<< ensureOne err entities
-
-ensureOne :: (Exception err, MonadLogger m, MonadThrow m) => err -> [a] -> m a
-ensureOne err = \case
-  []          -> throwM err
-  [x]         -> return x
-  (x : _ : _) -> do
-    logErrorN ("Too many database returns" :: Text)
-    return x
```
expede: No longer used without Selda

library/Fission/Web.hs
```diff
@@ -33,37 +37,59 @@ import qualified Fission.Web.User    as User
...
+  , MonadDB                t m
+  , Heroku.AddOn.Retriever t
+  , Heroku.AddOn.Destroyer t
+  , User.Creator           t
+  , User.Retriever         t
+  , User.Modifier          t
+  , User.Destroyer         t
+  , User.CID.Creator       t
+  , User.CID.Retriever     t
+  , User.CID.Destroyer     t
```
expede: Check it out! We never modify CIDs or Heroku Add Ons (which totally makes sense)

library/Fission/Web/Heroku.hs
```diff
@@ -8,20 +8,33 @@ import           Servant
...
-  , MonadLocalIPFS        m
-  , MonadRemoteIPFS       m
-  , MonadReflectiveServer m
+  ( MonadReflectiveServer    m
+  , MonadRemoteIPFS          m
+  , MonadLocalIPFS           m
+  , MonadLogger              m
+  , MonadLogger            t
+  , MonadThrow               m
+  , MonadThrow             t
```
expede: Keeping same constraints together. Lemme know if y'all would prefer to have the `t`s grouped instead.

library/Fission/Web/IPFS/DAG.hs
expede: Aaaaand done

library/Fission/Internal/Orphanage/OpenUnion.hs
```diff
@@ -0,0 +1,13 @@
...
+module Fission.Internal.Orphanage.OpenUnion () where
+
+import RIO
+import Data.WorldPeace
+
+instance Display (OpenUnion '[]) where
+  -- As a base case for the type-level recursion only. Essentially @Void@.
+  display = absurdUnion
+
+instance (Display a, Display (OpenUnion as)) => Display (OpenUnion (a ': as)) where
```
icidasset: Is this to easily render OpenUnion errors to text?

library/Fission/Web/Error/Class.hs
```diff
@@ -0,0 +1,91 @@
...
+    504 -> err504
+    505 -> err505
+
+    n -> error (show n <> " is not an error status code")
+
+instance ToServerError Error where
+  toServerError = \case
+    AddErr           addErr -> toServerError addErr
+    GetErr           getErr -> toServerError getErr
+    LinearizationErr linErr -> toServerError linErr
```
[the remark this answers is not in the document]
expede: The definition of these live in the `ipfs` package. But sure, can add some cross-documentation

library/Fission/User/Retriever/Class.hs
```diff
@@ -0,0 +1,22 @@
...
+
+instance MonadIO m => Retriever (Transaction m) where
+  getByUsername username = selectFirst
+    [ UserUsername P.==. username
+    , UserActive   P.==. True
+    ] []
+
+  getByHerkouAddOnId addOnId = selectFirst
+    [ UserHerokuAddOnId P.==. Just addOnId
+    , UserActive        P.==. True
```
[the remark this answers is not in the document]
expede: Because both Esqueleto and Persistent have `==.` operators, but they work on different types. To use this style, we need the Persistent `==.`, but want Esqueleto elsewhere.

library/Fission/Internal/Orphanage/OpenUnion.hs
expede: Yup! The power of type classes 💪

library/Fission/Web/Heroku/Deprovision.hs
```diff
@@ -64,28 +77,16 @@ deleteAssociatedWith uuid' = do
   return (deletedUserCids \\ remainingCIDs)
 
 -- | All records associated with the UUID, across the user, user CID, and Heroku add-on tables
-deleteAssociatedRecords :: MonadDB m => UserId -> UUID -> [Entity UserCid] -> Transaction m ()
+deleteAssociatedRecords ::
+  ( User.Destroyer         t
```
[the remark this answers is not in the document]
expede: Yup! It says that the function has access to the effect of destroying users, whether in a database, on a server, or in memory. This function can do that thing, and we'll specify _how_ it does it elsewhere.

library/Fission/Web/Heroku/Deprovision.hs
```diff
@@ -44,15 +52,20 @@ destroy uuid' = do
...
+  , MonadThrow             t
   )
   => UUID
-  -> Transaction m [CID]
+  -> t [CID]
 deleteAssociatedWith uuid' = do
-  addOn     <- herokuAddOnByUUID uuid'
-  addOnUser <- userForHerokuAddOn (entityKey addOn)
+  addOn     <- ensureEntityM err410 <| Heroku.AddOn.getByUUID uuid'
+  addOnUser <- ensureEntityM err410 <| User.getByHerkouAddOnId <| entityKey addOn
```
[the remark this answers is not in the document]
expede: @icidasset Sorry, I think you typed the same thing twice. What should it be?

`getByHerkouAddOnId`
`getByHerokuAddOnId`

library/Fission/Web/Heroku/Deprovision.hs
```diff
@@ -17,23 +17,31 @@ import           Servant
...
+  , User.Destroyer         t
+  , User.CID.Retriever     t
+  , User.CID.Destroyer     t
+  , Heroku.AddOn.Retriever t
+  , Heroku.AddOn.Destroyer t
   )
   => ServerT API m
 destroy uuid' = do
-  toUnpin <- runDB (deleteAssociatedWith uuid')
+  toUnpin <- runDB <| deleteAssociatedWith uuid'
```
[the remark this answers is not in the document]
expede: I don't have a super strong preference, but yeah parens are nice. Ben had a stronger preference for pipes, and this is based on his previous PR... so may have been either of us, not sure. I'll put it back, though 👍

Out of curiosity, which do you prefer?

library/Fission/User/CID/Destroyer/Class.hs
```diff
@@ -0,0 +1,29 @@
...
+import Fission.Prelude
+import Fission.Models
+
+-- | Actions for destroying @UserCid@s
+class Monad m => Destroyer m where
+  -- | Destroy a specific @UserCid@
+  destroy :: UserId -> CID -> m ()
+
+  -- | Destroy several @UserCid@s by they primary keys
+  destroyX :: [Key UserCid] -> m ()
```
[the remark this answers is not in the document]
expede: Sure, `destroyMany` works 👍

library/Fission/Web/Heroku/Deprovision.hs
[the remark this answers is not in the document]
icidasset: The second, the first has `Herkou` instead of `Heroku`.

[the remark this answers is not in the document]
expede: Ooooh, yeah, I see it now 😅 Thanks!!

library/Fission/Web/Heroku/Deprovision.hs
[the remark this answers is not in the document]
icidasset: Parentheses for small statements as this one (cuz maths).

library/Fission/User/CID/Destroyer/Class.hs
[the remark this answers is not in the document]
expede: Also changed `createX` to `createMany`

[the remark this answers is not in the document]
expede: I actually like `destroySeveral`, but Esqueleto has a couple `*Many` functions, so this is more consistent 👍

library/Fission/Web/Heroku.hs
dholms: I do think it makes more sense to group the `t`s. but don't have a strong opinion on this at all

library/Fission/Web/IPFS/DAG.hs
dholms: ok sweet yeah this makes sense. Love how much information we get about functions from the new way of doing it

library/Fission/Web.hs
```diff
@@ -33,37 +37,59 @@ import qualified Fission.Web.User    as User
...
+  , MonadDNSLink             m
+  , MonadLogger              m
+  , MonadLogger            t
+  , MonadThrow             t
+  , MonadTime                m
+  , MonadDB                t m
+  , Heroku.AddOn.Retriever t
+  , Heroku.AddOn.Destroyer t
+  , User.Creator           t
+  , User.Retriever         t
```
dholms: These constraints are starting to get... long. Which isn't necessarily bad. But what would you think of making something like 
```haskell
type User.DB t = (User.Creator t, User.Retriever t, User.Destroyer t, User.Modifier t)
``` 
and then just using `User.DB t` in the constraint list?

The downside is we lose some specificity from the constraints

expede: Yeah, the downside is that we can easily get lazy and just slap `User.DB` on everything, negating many of the benefits. As long as we keep it as small as possible, I'm okay with this 👍

expede: Called it `User.CRUD` and `User.CID.CRUD`

PR #253, opened 2020-01-20
Effectful HSpecs

expede:
Example output:

Screen Shot 2020-01-20 at 11 42 05 PM

Also, all those ones with a time next to them are running 10k auto-generated test cases each 💪 Actually did find a some edge cases this way. But yeah, at time of writing, the suite is running 40,008 tests.

The effect mocking is all set up. We don't have any complex setups here, so these are pretty straightforward. It's nice enough now, but this same strategy also applies to hanlders that mix external API calls with DB and IPFS lookups, &c, &c. Essentially we're turning impure production code into predicable pure code.


test/Test/Fission/Web/Ping.hs
```diff
@@ -0,0 +1,34 @@
...
+server :: MonadIO m => m Pong
+server = hoistServer api fromHandler (pure pong)
+
+tests :: IO TestTree
+tests =
+  testSpec "Fission.Web.Ping" do
+    describe "GET /" do
+      it "follows best practices" do
+        withServantServer api (pure server) \baseURL ->
+          serverSatisfies api baseURL defaultArgs do
```
icidasset: Few questions here:
- Are `withServantServer` and `serverSatisfies` both functions from the servant libraries?
- What does this test exactly? That it doesn't return a 5xx status?

test/Test/Fission/Mock/Types.hs
```diff
@@ -0,0 +1,100 @@
...
+
+{- | Fission's mock type
+
+     Notes:
+     * We will likely want @State@ in here at some stage
+     * @RIO@ because lots of constraints want @MonadIO@
+       * Avoid actual @IO@, or we're going to have to rework this 😉
+
+-}
+newtype Mock effs ctx a = Mock
```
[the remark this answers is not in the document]
expede: Testing pure code is much easier than all the things with side effects. This patterns turns side-effecting code into pure code.

We can then log what would have happened, and return fake values from e.g. a fake network call. So, for instance, in a complex handler that does a bunch of DB and AWS stuff: instead of actually trying to hit Route53 and DB setup/teardown it'll both return the correct response that the server would have responded with, and push `UpdateRoute53` and `CreatedUser` into the list of which effects were run. We then write tests for both the return values, and that it would have actually hit AWS and the database.

Because of how types work, we can use these mocks, and then separately test the actual DB access or whatever, and know that the test correctness composes (in a way that it just doesn't in languages with weaker type systems).

test/Test/Fission/Web/Ping.hs
expede: > Are withServantServer and serverSatisfies both functions from the servant libraries?

Yes, these are from libraries: https://hackage.haskell.org/package/servant-quickcheck-0.0.7.4/docs/Servant-QuickCheck.html

> What does this test exactly? That it doesn't return a 5xx status?

So far it's testing that:

* All valid inputs for the type defined in the `API` actually succeed (prop test)
* Has correct headers when the user is unauthorized

We can add more assertions later, but this is mainly for getting it off the ground

test/Test/Fission/Mock/Types.hs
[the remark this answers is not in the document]
icidasset: Amazing 🤯 👏

expede:
Wow, so good to have my laptop back 🙌 Pulling what I had from the temp DO server, cleaning up, and finishing this off 🚀

library/Fission/Internal/Mock/Types.hs
```diff
@@ -0,0 +1,251 @@
...
+    runner <- asks setDNSLink
+    return <| runner s c
+
+instance IsMember RunLocalIPFS effs => MonadLocalIPFS (Mock effs) where
+  runLocal _ _ = do
+    Effect.log RunLocalIPFS
+    asks localIPFSCall
+
+instance IsMember RunRemoteIPFS effs => MonadRemoteIPFS (Mock effs) where
+  runRemote = undefined
```
expede: Right now this class has `runRemote` as required, when the other options should be to define the other functions. Needs to be fixed in `ipfs-haskell`

PR #259, opened 2020-02-14
Apps, domains, subdomains, & GHC 8.8.3

expede:
* [x] Models
  * [x] `App`
  * [x] `Domain`
  * [x] `AppDomain` join table w/ optional subdomain
  * [x] Add `User.dataRoot` for FFS
* [x] Endpoints
  * [x] `/user`
    * [x] UPDATE
      * [x] Data root
  * [x] `/app`
    * [x] INDEX
    * [x] CREATE
    * [x] UPDATE
      * [x] CID
    * [x] DESTROY

# "While I Was In There"

* [x] Update form GHC 8.6.5 to 8.8.3 (even-numbered minor version = stable)
* [x] Remove unused dependencies

# PUSHED TO NEXT PR

* Endpoints
  * `/dns/domain/:domain` (?)
    * CREATE (register)
    * READ (status)
    * UPDATE (move ownership)
    * DESTROY (unregister)


library/Fission/App/Domain/Associate/Class.hs
```diff
@@ -0,0 +1,56 @@
...
+class Monad m => Associate m where
+  associate ::
+       AppId
+    -> DomainName
+    -> Maybe Subdomain
+    -> UTCTime
+    -> m (Either Errors ())
+
+instance MonadIO m => Associate (Transaction m) where
+  associate appId domainName maySubdomain now = do
```
expede: Needs auth!

library/Fission/App/Domain/Dissociate/Class.hs
```diff
@@ -0,0 +1,66 @@
...
+class Monad m => Dissociate m where
+  dissociate    :: AppId -> DomainName -> Maybe Subdomain -> UTCTime -> m (Either Errors ())
+  -- dissociateAll :: AppId -> UTCTime -> m (Either Errors ())
+
+  -- FIXME: move to Domain/Delete
+  -- dissociateAllByDomainName :: AppId -> DomainName -> UTCTime -> m (Either Errors ())
+
+instance MonadIO m => Dissociate (Transaction m) where
+  dissociate appId domainName maySubdomain now = do
+    insert_ DissociateAppDomainEvent
```
expede: Needs auth!

library/Fission/App/Retriever/Class.hs
```diff
@@ -0,0 +1,19 @@
...
+import           Fission.Models
+
+import           Fission.Models.Error
+
+class Monad m => Retriever m where
+  byId :: AppId -> m (Either (NotFound App) (Entity App))
+
+instance MonadIO m => Retriever (Transaction m) where
+  byId appId =
+    appId
```
expede: Needs auth

library/Fission/App/Modifier/Class.hs
```diff
@@ -0,0 +1,36 @@
...
+
+type Errors = OpenUnion
+  '[ NotFound App
+   ]
+
+class Monad m => Modifier m where
+  updateCID :: AppId -> CID -> UTCTime -> m (Either Errors ())
+
+instance MonadIO m => Modifier (Transaction m) where
+  updateCID appId newCID now = do
```
expede: AUTH!

library/Fission/LoosePin/Destroyer/Class.hs
```diff
@@ -0,0 +1,29 @@
...
+class Monad m => Destroyer m where
+  -- | Destroy a specific @LoosePin@
+  destroy :: UserId -> CID -> m ()
+
+  -- | Destroy several @LoosePin@s by they primary keys
+  destroyMany :: [LoosePinId] -> m ()
+
+instance MonadIO m => Destroyer (Transaction m) where
+  destroyMany userCidIds =
+    delete <| from \userCid ->
```
expede: Auth!

library/Fission/App/Retriever/Class.hs
expede: Fixed on line 27

library/Fission/Models.hs
```diff
@@ -60,26 +75,160 @@ User
...
+Domain
+  ownerId    UserId
+  domainName DomainName
+
   insertedAt UTCTime
   modifiedAt UTCTime
 
-  UniqueUserCid userFk cid
+  Primary domainName
+  UniqueDomainNamePerOwner domainName ownerId
```
expede: TODO: remove the uniqueness constraint because it's already the primary key

library/Fission/Models.hs
```diff
@@ -60,26 +75,160 @@ User
...
+------------
+-- Events --
+------------
+
+CreateAppEvent
+  ownerId     UserId
+  cid         CID
+
+  insertedAt  UTCTime
+  deriving Show Eq
```
expede: Don't need this either

library/Fission/Models.hs
```diff
@@ -60,26 +75,160 @@ User
...
+
+AppDomain
+  appId      AppId
+
+  domainName DomainName
+  subdomain  Subdomain Maybe
+
+  insertedAt UTCTime
+
+  UniqueSubdomainPerDomain domainName subdomain !force
```
expede: Check for nulls, too

package.yaml
```diff
@@ -92,21 +92,18 @@ dependencies:
...
   - flow
-  - fsnotify
   - Glob
-  - haskeline
   - http-client
   - http-client-tls
   - http-media
   - http-types
-  - ip
+  # - ip
```
expede: Actually delete this line

library/Fission/App/Creator/Class.hs
```diff
@@ -0,0 +1,46 @@
...
+instance (AppDomain.Initializer m, MonadDNSLink m) => Creator (Transaction m) where
+  create ownerId cid now = do
+    appId <- insert App
+      { appOwnerId    = ownerId
+      , appCid        = cid
+      , appInsertedAt = now
+      , appModifiedAt = now
+      }
+
+    AppDomain.associateDefault ownerId appId now >>= \case
```
dholms: why don't you need an `Associator` monad as a constraint on this instance? It looks like it's a constraint to call `associateDefault`

dholms: Actually I see this is the same thing for others as well (such as `AppDomain.allSiblingsByDomain` used in the `App.Destroyer` instance looking like it should have a `Retriever` constraint)

library/Fission/Models.hs
```diff
@@ -60,26 +75,160 @@ User
 
   deriving Show Eq
 
-UserCid
-  userFk     UserId
+------------
+-- Events --
```
dholms: Events are just normal SQL records right? But they're used as an append-only log of all of the transformations?

library/Fission/App/Domain/Initializer/Class.hs
```diff
@@ -0,0 +1,11 @@
+module Fission.App.Domain.Initializer.Class (Initializer (..)) where
+
+import           Fission.Prelude
+import           Fission.URL.DomainName.Types (DomainName)
+
+class Monad m => Initializer m where
+  -- | The default domain that's managed by Fission
```
dholms: I'm confused how this goes about getting the default value such that it can be used 
`defaultDomainName <- AppDomain.initial`

library/Fission/Error.hs
```diff
@@ -0,0 +1,29 @@
+module Fission.Error
+  ( openLeft
+  , relaxedLeft
+  , fromMaybe
+  , fromMaybe'
+  , module Fission.Error.Types
+  ) where
+
+import Data.WorldPeace
```
dholms: why import these separately instead of `Fission.Prelude`?

library/Fission/Models.hs
expede: @dholms yep!

* Regular SQL table
* Append-only = no delete or update operations

library/Fission/Platform/Heroku/AddOn/Creator/Class.hs
```diff
@@ -3,14 +3,14 @@ module Fission.Platform.Heroku.AddOn.Creator.Class (Creator (..)) where
...
 import           Fission.Prelude
+import           Fission.Error.Types
+import           Fission.Models
 
 import           Fission.Platform.Heroku.Region.Types
-import           Fission.Platform.Heroku.AddOn.Creator.Error
 
 class Monad m => Creator m where
-  create :: UUID -> Region -> UTCTime -> m (Either Error HerokuAddOnId)
+  create :: UUID -> Region -> UTCTime -> m (Either (AlreadyExists HerokuAddOn) HerokuAddOnId)
```
dholms: so much better

library/Fission/Web/App/Destroy.hs
```diff
@@ -0,0 +1,66 @@
...
+import           Fission.URL.Types
+
+import qualified Fission.App.Destroyer.Class as App
+import           Fission.Web.Error.Class
+
+type API = ByURLAPI :<|> ByIdAPI
+
+type ByURLAPI
+  =  "associated"
+  :> Capture "url" URL
```
dholms: how does servant tell the difference between a URL & an AppId if they're both strings?

library/Fission/Web/Error/Class.hs
```diff
@@ -1,5 +1,8 @@
 module Fission.Web.Error.Class (ToServerError (..)) where
 
+import           RIO
```
dholms: `Fission.Prelude`?

library/Fission/App/Domain/Initializer/Class.hs
expede: No problem! This is a fairly common pattern, so good to go over:

```haskell
initial :: m DomainName
```

This says that we gen get a domain name inside some monadic context. This may or may not be a constant, it depends on `m`.

```haskell
instance App.DomainName.Initializer TestContext where
  initial = return (DomainName "fake.com") -- i.e. Hardcoded constant still possible

example = do
  domain <- App.DomainName.initial
  return (domain == DomanName "fake.com") -- True
```

The `Fission` monad contains environment variables it reads on startup, which is why this

1. isn't a hardcoded constant
2. needs to be in a monad (the "effect" is reading from the app's config)

```haskell
data Config = Config
  {  baseAppDomainName :: !URL.DomainName
  -- ...plus a bunch more fields
  }

-- | The top-level app type
newtype Fission a = Fission { unwrapFission :: RIO Config a }
                                                -- ^^^^^^
                                                --  THIS

instance App.Domain.Initializer Fission where
  initial = asks baseAppDomainName -- i.e. extract the value from Config
```

---

To expand your mind on what you _can_ do with this:

Either is a monad for things that may fail.

```haskell
instance App.Domain.Initializer (Either MyError) where
  initial = Right (DomainName "example.com")
```

Lists are monads (for when you have multiple possible values)

```haskell
instance App.Domain.Initializer [a] where
  initial = [ DomainName "example.com"
            , DomainName "omgubuntu.co.uk"
            , DomainName "cbc.ca"
            ]
```

Random values:

```haskell
instance App.Domain.Initializer Random where
  initial = do
    txt <- randomText
    return (DomainName (txt <> ".com"))
```

library/Fission/Error.hs
expede: Ah, good point! I had a module cycle at one point, and removed `Fission.Prelude`. Will put that back as it was 👍

library/Fission/Web/App/Destroy.hs
expede: It'll check for `"associated"` first

```
/app/associated/foo.com
/app/12345
```

library/Fission/App/Creator/Class.hs
expede: > why don't you need an Associator monad as a constraint on this instance? It looks like it's a constraint to call associateDefault

Yeah, you'd think, right? The compiler tells you that it's not required here. Because they both have `MonadIO m =>` constraints, and it can resolve that they both work on _any_ `MonadIO` (think of it now as a property of MonadIO).

A simpler example is the concrete `Fission` type. It has instances for all of these classes, but they don't need special constraints, because Fission _is a member of_ each typeclass (it is an Associator, a Modifier, &c). The specific `Associator` constraint is thus redundant.

Typeclasses work the opposite direction from inheritance, more like a mixin —- it's superclassing rather than subclassing.

dholms: ah got it. ok very cool :+1:

library/Fission/App/Domain/Initializer/Class.hs
dholms: ahh okay this makes sense. I didn't see the Initializer instances in `Fission.Types` so was hopelessly confused how it was finding them :laughing:

library/Fission/Web/App/Destroy.hs
dholms: ah got it got it :+1:

PR #260, opened 2020-02-20
Support Multiple DID Methods in HOAN

expede:
Since the WebCrypto API doesn't officially include EdDSA (yet?), Chrome has support for it but FF does not 😭 Since this is all user-controlled AuthN, we should probably support multiple methods. We already have Higher Order AuthN (HOAN), so this should not be prohibitive to implement in the near term.

* [x] Figure out which we will support first
  * ECDSA
  * ~~SECP~~
  * ~~P-256 / P-512~~
  * RSA (2048/~~4096~~)
* [ ] Implement in DB

### DIFF
--- a/library/Fission/Models.hs
+++ b/library/Fission/Models.hs
@@ -44,7 +44,9 @@ HerokuAddOn
   deriving Show Eq
 
 User
-  did           DID           Maybe
+  didAlgorithm  DID.Algorithm Maybe
+  didPK         DID.PublicKey Maybe
+
   username      Username
   email         Email         Maybe
   role          Role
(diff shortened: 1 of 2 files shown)


expede:
Notes from Discord conversation:

* Today only Chrome supports EdDSA in both the WebCrypto API _and_ IndexedDB
* Same issue with NIST P-256/512
* Everyone supports RSA

Logistics:
1. Implement 2048-bit RSA for all browsers
2. Add EdDSA for those that support it
3. Upgrade users to EdDSA as it becomes available

expede:
Superseded by #299

PR #262, opened 2020-03-16
Custom dns

dholms:
:construction: :construction:  WIP :construction: :construction: 

Creating PR to easily see/annotate changes

This adds functions/routes to:
- create a hosted zone (still need to add code for reusable delegation sets, see my comment)
- request a certificate
- describe a certificate (partially implemented)

The process for creating a custom domain:
- create hosted zone (with reusable delegation set from our whitelabel name servers)
- request a certificate
- update route53 to match CertManager's dns challenge
- add certificate to gateway loadbalancer
- have user point name servers at aws (do we need some sort of challenge to ensure that they own that domain? Or is being able to point nameservers is enough?)
- update domain/subdomains through route53


library/Fission/AWS/Route53.hs
```diff
@@ -1,3 +1,145 @@
...
+  ( MonadAWS    m
+  , MonadLogger m
+  )
+  => URL.DomainName
+  -> m (Either ServerError CreateHostedZoneResponse)
+createHostedZone' (URL.DomainName domain) = do
+  logDebug <| "Updating DNS record at: " <> displayShow domain
+
+  let
+    req = createHostedZone domain "CallerReference"
```
dholms: Need to use lens setters here to `chzDelegationSetId`

I have that id locally (you can also get it by running `aws route53 list-reusable-delegation-sets` from the aws-cli). I don't think it needs to be kept private, but I'll hold off posting it for now

library/Fission/Web/Cert.hs
```diff
@@ -0,0 +1,32 @@
...
+
+import           Fission.Web.Error          as Web.Err
+import qualified Fission.URL.Types          as URL
+
+type API = Capture "domain" URL.DomainName
+        :> PostAccepted '[PlainText, OctetStream] URL.DomainName
+
+server :: (MonadLogger m, MonadCertManager m) => Entity User -> ServerT API m
+-- server (Entity _id User { userUsername }) domain =
+server _user domain = do
```
dholms: This is all still very WIP, just trying to get a hang of the API

library/Fission/AWS/CertManager.hs
```diff
@@ -0,0 +1,110 @@
...
+  ( MonadAWS    m
+  , MonadLogger m
+  )
+  => CertARN
+  -> m (Either ServerError CertificateDetail)
+describeCertMock arn = do
+  logDebug <| "MOCK: Describing Cert for arn: " <> show arn
+  return <| Right certificateDetail
+
+parseValidation :: MonadLogger m => CertificateDetail -> m ()
```
dholms: Also still very WIP, getting a hang of the API for issuing certs

library/Fission/AWS/Route53.hs
dholms: http://hackage.haskell.org/package/amazonka-route53-1.6.1/docs/Network-AWS-Route53-CreateHostedZone.html

PR #266, opened 2020-03-19
Hotfix 2.0.1

expede:
Post mortem:
* Upgrade script was restarting the wrong port [FIXED IN SCRIPT]
* DB was enforcing unique emails [WAITING ON DOMAINS TO DE-DUP USERS]
* Route53 API is now pickier about double quotes (maybe a newer API when we upgraded libs?). Switched quotes in code.
* A staging server wouldn't be the worst idea. We've probably hit that point

### DIFF
--- a/library/Fission/Models.hs
+++ b/library/Fission/Models.hs
@@ -61,7 +61,7 @@ User
   role          Role
   active        Bool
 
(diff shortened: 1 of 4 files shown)


expede:
Working on prod, backported these changes locally

PR #268, opened 2020-03-21
Username validity criterea & checks

expede:
Some usernames are... bad since they're also subdomains. Phishing and whatnot could ensure if we allowed, for example `support.fission.name`


library/Fission/User/Creator/Error.hs
```diff
@@ -18,3 +21,16 @@ instance Display AlreadyExists where
 instance ToServerError AlreadyExists where
   toServerError AlreadyExists =
     err409 { errBody = UTF8.showLazyBS <| textDisplay AlreadyExists }
+
+data InvalidUsername = InvalidUsername
```
dholms: Do we still need this type if we have `Username.Invalid`? Doesn't seem to be used anywhere

expede: Good catch! 🔪

PR #269, opened 2020-03-23
HOTFIX: backwards compatability with current CLI

expede:

### DIFF
--- a/library/Fission/User/Creator/Class.hs
+++ b/library/Fission/User/Creator/Class.hs
@@ -105,6 +105,7 @@ instance
                     Left err             -> Error.relaxedLeft err
                     Right (_, subdomain) -> Right (userId, subdomain)
 
+  -- FIXME: remove
   createWithPassword username password email now =
     Password.hashPassword password >>= \case
       Left err ->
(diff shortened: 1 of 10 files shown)


library/Fission/User/Creator/Class.hs
```diff
@@ -98,6 +105,37 @@ instance
                     Left err             -> Error.relaxedLeft err
                     Right (_, subdomain) -> Right (userId, subdomain)
 
+  -- FIXME: remove
```
expede: Just a reminder that this will get ✂️ in the future

library/Fission/User/Registration/Types.hs
```diff
@@ -26,7 +29,8 @@ instance ToJSON Registration where
 
 instance FromJSON Registration where
   parseJSON = withObject "Registration" \obj -> do
-    username <- obj .:  "username"
+    username <- obj .: "username"
+    password <- obj .: "password"
```
dholms: Does this need to be `.:?` since password isn't always included?

expede: Good question! It's only `.:?` if you want to provide a default value. We're saying here that it could be `Nothing`, and that's a valid value!

dholms: Oh cool cool :+1:

expede: (Context: this is for a JSON parser, which may have a different representation than the underlying data structure)

PR #272, opened 2020-03-24
Re-integrate CLI

expede:


app/Fission/Web/Log/Sentry.hs
```diff
@@ -1,70 +0,0 @@
--- | Error reporting via @sentry.io@
```
expede: Actual version is now in `/web`, but somehow GH can't figure that out, soooo 🤷‍♀

PR #299, opened 2020-03-31
Self Signed `did:key` JWT — Ed25519 & RSA2048

expede:
# This PR

* [x] Make our JWTs work with the `did:key` spec (Ed25519 and RSA2048)
* [x] Add field to users table to account for which key type they're using
* [x] Update higher order auth in server
* [x] Update higher order auth in client
* [x] Rename `JWT.Token` to `JWT`
* [x] Consolidate some low level crypto stuff
* [x] Move parsing to `ToJSON`/`FromJSON` instances
* [x] Tests
* [x] Brought back `$` 
  * It was driving me pretty wild
  * We have multiple conventions 🤷‍♀ 

# For Later PRs

* [ ] PublicKey chain logging -- Record full credential in middleware logger
* [ ] Enforce & check `aud` in recursive JWTs
* [ ] Change `CommandM` to `CommandT` (see note in code comment)


expede:
Okay, basically done except for the comments that are like "oh no write tests I have no structure gaaaaaah!"

expede:
To be honest, I found the `did:key` multiformats stuff SUPER confusing. To avoid as much of that as possible in the future, we're using some magic numbers for the two kinds of keys that we actually have. Please feel free to double check my math 😉

expede:
Branching to `UCAN` for macaroons

library/Fission/User/DID/Types.hs
```diff
@@ -1,27 +1,117 @@
...
+      header = "did:" <> textDisplay method <> ":" <> "z"
+
+      multicodec64 :: ByteString
+      multicodec64 = BS.pack magicBytes <> encodeUtf8 pk
+
+      magicBytes :: [Word8]
+      magicBytes =
+        case algo of
+          Ed25519 -> [0xed, 0x01]
+          RSA2048 -> [0x00, 0x75, 0x01]
```
dholms, line 87: got a different magic number for RSA (373): [0xf5, 0x03] instead of [0x75, 0x01]
0xf5 is the same as 0x75 except for the leading bit, which needs to be a 1 since there is another byte that follows
the second number needs to equal 256. from their examples 128 == 00000001 == 0x01 so 256 == 00000010 == 0x03
lmk if I'm thinking about it wrong

library/Fission/User/DID/Types.hs
```diff
@@ -1,27 +1,117 @@
...
+
+instance ToJSON DID where
+  toJSON (DID (Key.Public pk) algo method) =
+    String (header <> UTF8.toBase58Text multicodec64)
+    where
+      header :: Text
+      header = "did:" <> textDisplay method <> ":" <> "z"
+
+      multicodec64 :: ByteString
+      multicodec64 = BS.pack magicBytes <> encodeUtf8 pk
```
dholms, line 81: These aren't base64 encoded so we probably want to change the name of this variable

I'm not sure what `BS.pack` actually encodes to, but it's what the base58 package does under the hood so that seems right

`encodeUtf8 pk` encodes to UTF-8 so we're sending `base58(utf8(pk))`. where it looks like pk is already base58 encoded text. so it's actually `base58(utf8(base58(pk)))`. Wouldn't it be better to just send `base58(pk)`? I know we talked about something similar on our call, but I can't quite remember the details, so I might be forgetting the reason we're doing this

expede: @dholms 

> These aren't base64 encoded so we probably want to change the name of this variable

I'm very confused then 😛  Here's my thinking:

* `magicBytes` are binary
  * Entered in hex (base16)
  * Rendered down to `[Word8]`s (lists of base8)

* The type of is `pk :: Text`, which is UTF8, which can technically be in any base
  * What you do with it from there is up to you

My understanding is that you can't just append arbitrary base58 together. I've been double checking my work against a base58 converter.

```
| Base16       | Base64 | Base58   |
|--------------|--------|----------|
| 616263       | abc    | ZiCa     |
|       646566 |    def | aisf     |
| 616263646566 | abcdef | qVgfxYy3 | <- Lines 1 & 2 of base64 appended
| 41aff07b7296 | Aﾯ￰{rﾖ  | ZiCaaisf | <- Lines 1 & 2 of base58 appended
```

Since the actual multicodec flags are hex, as I assume is the key, my understanding is that we can't just append text together and call it a day. 

We need to _first render the entire string down to binary_, and then encode in base58. Here's the format:

```
<MULTIBASE(base58-btc, MULTICODEC(public-key-type, raw-public-key-bytes))>
                                                   ^^^^^^^^^^^^^^^^^^^^
                                                    MAKE THE KEY BINARY
```

So in pseudo-code:

```haskell
base58key
  |> toBinary
  |> append(magicNumberBinary)
  |> encodeBase58
  |> append("z")
```

But I can also be TOTALLY WRONG. This stuff is confusing.

library/Fission/User/DID/Types.hs
expede: Working out loud:

```
From example

11111111 00000001 = 255 
c1111111 c0000001
 |--+--|        ^
    |           |
   127   +     128 = 255 ✅
```

```
  0x75     0x02
01110101 00000010
c1110101 c0000010
 |--+--|       ^
    |          |
   117   +    256 = 373 ✅
```

> 0xf5 is the same as 0x75 except for the leading bit, which needs to be a 1 since there is another byte that follows

Oh god, I did this _again_ in the above 😆

```
  0xF5     0x02
11110101 00000010
```

So we have a discrepancy

`[0xF5, 0x02]` vs `[0xF5, 0x02]`

Can you double check our math again please? 😅

dholms: oh yeah it's totally 0x02 idk why my brain interpreted that as 0x03.

Yeah so just so we're in agreement:
:guardsman::guardsman:  `[0xF5, 0x02]` :guardsman::guardsman:

library/Fission/User/DID/Types.hs
dholms: yeah sorry if I was unclear 

3 points:

1) `multicode64`: I was mostly remarking on the variable name here. None of these values are base64, `BS.pack magicBytes` is raw binary (packed as a bytestring) and `encodeUtf8 pk` is a utf-8 encoded base58 key. a better variable name might just be `multicodec`

2) `BS.pack magicBytes` is raw binary packed into a bytestring. This seems right :+1: 

3) `encodeUtf8 pk` does not go to the raw public key bytes. `pk` is already a base-58 representation of the key. So `encodeUtf8 pk = utf8(base58(raw-public-key-bytes))`

In other words, the pseudo code is right, but the key is not being properly translated to binary
```haskell
base58key
  |> toBinary -- we're not actually going to the raw binary of the key
  |> append(magicNumberBinary)
  |> encodeBase58
  |> append("z")
```

unless I'm misunderstanding the code, right now the `toBinary` step is `encodeUtf8 pk`

it should be something like:
```haskell
toBinary :: Public -> ByteString
toBinary = BS58.BTC.toBytes . BS58.BTC.fromText . publicKey
```

also agreed on base-58 concatenation not working :+1: I don't think arbitrary concatenation works unless the base is an exponent of 2

expede: > pk is already a base-58 representation of the key.

Ah, yeah, good point! This is already in base58 because it's coming off the wire with the right _characters_ but in base64. So yes — reinterpret as base58 with `fromText`!

```
toBinary :: Public -> ByteString
toBinary = BS58.BTC.toBytes . BS58.BTC.fromText . publicKey
```

Right, exactly like what I have in the deserializer :P


Thanks!

expede: Oh wait, actually no.

The `pk` in `ToJSON` is a `Text` value that we previously stored. _It's stored in base64._

dholms: I thought `pk` was a Base-58 string from looking at its `Arbitrary` instance. Even tho it's encoded base2 by `Text`. So like base2 encoded base-58 string (this stuff is so confusing to talk about 😛)

> multicodecW8 is a better name
yeah I like that

expede: Oh, yeah, confusing for sure 💯 `Arbitrary` gives us our consistent _internal_ representation, which is base2 `Text`

expede: Oh, you mean the `Arbitrary Key.Public` instance. Yeah... that is literally just random  bytes. The base58 stuff there is just me being confused about base vs text encoding 😅

dholms: But it's a base2 `Text` representation of a base58 string right?
```haskell
instance Arbitrary Public where
  arbitrary = do
    nonempty :: NonEmpty Word8 <- arbitrary
    let pk = BS58.BTC.toText . BS58.BTC.fromBinary $ NonEmpty.toList nonempty
    return (Public pk)
```

expede: Comment at the same time! 👏

It is yeah. But we don't care about that for our internal representation. I've changed it to this just now:

```
instance Arbitrary Public where
  arbitrary = Public <$> arbitrary
```

dholms: ah okay cool 👍 so what's the expected encoding of a key then? Base64? (that's what I use in `keystore-idb` for instance)

expede: The key itself is binary — so base64 is a pretty natural way to turn that into text 👍

Inside the Haskell application, it's raw binary again. When we read in a key, the default decoder will treat the incoming text as base64. When sending us request authed by a UCAN, it'll need to include this `did:key:z<BASE58>` stuff, but you can store the key in whatever format on the client 👍

PR #302, opened 2020-04-14
UCAN

expede:
Implement UCAN auth in server and CLI


library/Fission/Authorization/ServerDID/Class.hs
```diff
@@ -0,0 +1,11 @@
...
+
+import Servant.Server
+
+import Fission.Prelude
+
+import Fission.User.DID.Types
+
+class Monad m => ServerDID m where
+  getServerDID :: m DID
+  publicize    :: m (Either ServerError ())
```
expede, line 11: Break out into own TC

Issue #314, opened 2020-04-24
User creation endpoint should return error instead of a 500 error


icidasset:
## Problem

The `PUT /user` endpoint gives me a 500 status instead of a 4xx with a proper error message.

## Solution

Return a `422` error or something with some json along the lines of `{ "error": "..." }`.

expede:
@icidasset it should be a LOT better now, though I'm investigating even better, nicely formatted JSON errors

PR #317, opened 2020-04-27
Fix RSA DB encoding & simplify PK types

expede:
This is a good one I think:

* Radically simplify PK encoding / decoding
* Drop `user.algorithm` DB column
  * We can infer the key type from length
* Much more detailed error logging
* Ignore the ELB health check ping in logs
* Fix Fission DID (oops!)


expede:
Closes #314

expede:
Deploy checklist:

* [ ] Run SQL to drop column
* [ ] Change the Fission DID 😬 _OOPS_

PR #319, opened 2020-04-30
Fix RSA DID decoding format

expede:


expede:
Okay, I guess PEM keys are meant to be base64 encoded 😅

PR #323, opened 2020-05-02
🛂🔗 CLI User ID Linking 

expede:
Waiting on app changes. Also need to figure out what to do about pubsub.


expede:
Restarting this one

PR #326, opened 2020-05-12
Fix link to releases page in README

ellttBen:
## Summary

* Fixes a wrong link to releases in the README which pointed to the fission org. instead of fission-suite.

## Test plan (required)

N/A

## Closing issues

N/A

## After Merge
N/A


bmann:
Thanks @ellttBen -- merged.

Also, looks like there's a bug in our new contributor greeting, so I'll file an issue about that ;)

PR #335, opened 2020-06-08
Only check the Fission DID for the outmost wrapper

expede:


library/Fission/Web/Auth/Token/JWT/Proof.hs
```diff
@@ -9,18 +9,19 @@ module Fission.Web.Auth.Token.JWT.Proof
   , module Fission.Web.Auth.Token.JWT.Proof.Error
   ) where
 
-import qualified RIO.Text as Text
+import qualified RIO.Text                               as Text
 
 import           Fission.Prelude
 
-import           Fission.Web.Auth.Token.JWT as JWT
+import           Fission.Web.Auth.Token.JWT             as JWT
```
expede, line 16: ☝️ Stylish Haskell

library/Fission/Internal/Fixture/Bearer/Nested.hs
```diff
@@ -0,0 +1,57 @@
...
+instance Proof.Resolver InTimeBounds where
+  resolve = error "shouldn't hit this in our specific case"
+
+instance ServerDID InTimeBounds where
+  getServerDID = pure did
+    where
+      Right did = eitherDecode "\"did:key:zStEZpzSMtTt9k2vszgvCwF4fLQQSyA15W5AQ4z3AR6Bx4eFJ5crJFbuGxKmbma4\""
+
+instance MonadTime InTimeBounds where
+  currentTime = pure $ Time.fromSeconds 1591627760
```
expede, line 43: In case anyone is interested in how we test side effects in a totally pure way 😉

PR #338, opened 2020-06-16
CLI — Retry 100 times on `up` or `watch`

expede:
# Problem

When the IPFS gateway fails with a 502 or 504, it usually works after a manual retry. It's annoying to have to do that.

# Solution

Automatically retry up to 100(?) times on specifically a 502 or 504

# Tasks

* [x] `fission up` auto-retries
* [x] `fission watch` auto-retries
  * _But probably don't if there's a FSNotify event that there's new content!!_
* [ ] Docs
* [x] Automated tests

### DIFF
--- a/library/Fission/CLI/Command/Up.hs
+++ b/library/Fission/CLI/Command/Up.hs
@@ -1,13 +1,21 @@
 -- | File sync, IPFS-style
-module Fission.CLI.Command.Up (cmd, up) where
+module Fission.CLI.Command.Up
+  ( cmd
+  , up
+  , requestWithRetry
+  ) where
 
 import qualified Crypto.PubKey.Ed25519 as Ed25519
 
 import           Options.Applicative
 import           RIO.Directory
 
+import           Network.HTTP.Types.Status
+
 import           Network.IPFS
 import qualified Network.IPFS.Add as IPFS
+
+import           Servant.Client
 
 import           Fission.Prelude
 import           Fission.Models
@@ -84,7 +92,7 @@ up Up.Options {..} = do
       logDebug $ "Starting single IPFS add locally of " <> displayShow absPath
      
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid -> do
-        withRetryM 100 (sendRequestM (updateApp url cid copyFiles)) >>= \case
+        retrier (updateApp url cid copyFiles) >>= \case
           Left err ->
             CLI.Error.put err "Server unable to sync data"
            
@@ -114,3 +122,27 @@ parseOptions = do
 
   return Up.Options {..}
 
+requestWithRetry ::
+  ( MonadWebClient m
+  , MonadLogger m
+  )
+  => m (ClientM a)
+  -> m (Either ClientError a)
+requestWithRetry req =
+  sendRequestM req >>= \case
+    Right val ->
+      return $ Right val
+
+    Left err@(FailureResponse _req res) -> do
+      let code = statusCode $ responseStatusCode res
+
+      if code >= 502 && code <= 504
+        then do
+          logWarn $ "Got a " <> textDisplay code <> "; retrying..."
+          retrier req
+
+        else
+          return $ Left err
+
+    Left err ->
+      return $ Left err
--- a/library/Fission/CLI/Command/Up.hs
+++ b/library/Fission/CLI/Command/Up.hs
@@ -11,9 +11,11 @@ import qualified Network.IPFS.Add as IPFS
 
 import           Fission.Prelude
 import           Fission.Models
+
+import           Fission.Error
+
 import           Fission.Authorization.ServerDID
-import           Fission.Error.NotFound.Types
- 
+
 import           Fission.Web.Auth.Token
  
 import           Fission.Web.Client     as Client
@@ -82,7 +84,7 @@ up Up.Options {..} = do
       logDebug $ "Starting single IPFS add locally of " <> displayShow absPath
      
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid -> do
-        sendRequestM (updateApp url cid copyFiles) >>= \case
+        withRetryM 100 (sendRequestM (updateApp url cid copyFiles)) >>= \case
           Left err ->
             CLI.Error.put err "Server unable to sync data"
            
@@ -111,3 +113,4 @@ parseOptions = do
     ]
 
   return Up.Options {..}
+
--- a/library/Fission/CLI/Command/Up.hs
+++ b/library/Fission/CLI/Command/Up.hs
@@ -28,6 +28,7 @@ import           Fission.Web.Auth.Token
  
 import           Fission.Web.Client     as Client
 import           Fission.Web.Client.App as App
+import           Fission.Web.Client.Error
 
 import           Fission.CLI.Environment as Environment
 
--- a/library/Fission/CLI/Command/Up.hs
+++ b/library/Fission/CLI/Command/Up.hs
@@ -90,7 +90,7 @@ up Up.Options {..} = do
       logDebug $ "Starting single IPFS add locally of " <> displayShow absPath
      
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid -> do
-          req <- App.update url cid copyFiles
+          req <- App.mkUpdateReq url cid copyFiles
           retryOnStatus [status502, status504] 100 req >>= \case
             Left err ->
               CLI.Error.put err "Server unable to sync data"
--- a/library/Fission/CLI/Command/Up.hs
+++ b/library/Fission/CLI/Command/Up.hs
@@ -92,13 +92,15 @@ up Up.Options {..} = do
       logDebug $ "Starting single IPFS add locally of " <> displayShow absPath
      
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid -> do
-        retrier (updateApp url cid copyFiles) >>= \case
-          Left err ->
-            CLI.Error.put err "Server unable to sync data"
-           
-          Right _  -> do
-            CLI.Success.live cid
-            CLI.Success.dnsUpdated url
+          updateApp url cid copyFiles
+            |> retryOnErr [status502, status504] 100 
+          >>= \case
+            Left err ->
+              CLI.Error.put err "Server unable to sync data"
+            
+            Right _  -> do
+              CLI.Success.live cid
+              CLI.Success.dnsUpdated url
 
   where
     updateApp url cid copyFiles =
@@ -139,7 +141,8 @@ requestWithRetry req =
       if code >= 502 && code <= 504
         then do
           logWarn $ "Got a " <> textDisplay code <> "; retrying..."
-          retrier req
+          req |> 
+            retryOnErr [status502, status504] 100
 
         else
           return $ Left err
--- a/library/Fission/CLI/Command/Up.hs
+++ b/library/Fission/CLI/Command/Up.hs
@@ -91,7 +91,7 @@ up Up.Options {..} = do
      
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid -> do
           updateApp url cid copyFiles
-            >>= retryOnErr [status502, status504] 100 
+            >>= retryOnStatus [status502, status504] 100 
           >>= \case
             Left err ->
               CLI.Error.put err "Server unable to sync data"
--- a/library/Fission/CLI/Command/Up.hs
+++ b/library/Fission/CLI/Command/Up.hs
@@ -90,22 +90,14 @@ up Up.Options {..} = do
       logDebug $ "Starting single IPFS add locally of " <> displayShow absPath
      
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid -> do
-          updateApp url cid copyFiles
-            >>= retryOnStatus [status502, status504] 100 
-          >>= \case
+          req <- App.update url cid copyFiles
+          retryOnStatus [status502, status504] 100 req >>= \case
             Left err ->
               CLI.Error.put err "Server unable to sync data"
             
             Right _  -> do
               CLI.Success.live cid
               CLI.Success.dnsUpdated url
-
-  where
-    updateApp url cid copyFiles =
-      authClient (Proxy @App.Update)
-        `withPayload` url
-        `withPayload` cid
-        `withPayload` Just copyFiles
 
 parseOptions :: Parser Up.Options
 parseOptions = do
--- a/library/Fission/CLI/Command/Up.hs
+++ b/library/Fission/CLI/Command/Up.hs
@@ -2,7 +2,6 @@
 module Fission.CLI.Command.Up
   ( cmd
   , up
-  , requestWithRetry
   ) where
 
 import qualified Crypto.PubKey.Ed25519 as Ed25519
@@ -14,8 +13,6 @@ import           Network.HTTP.Types.Status
 
 import           Network.IPFS
 import qualified Network.IPFS.Add as IPFS
-
-import           Servant.Client
 
 import           Fission.Prelude
 import           Fission.Models
@@ -94,7 +91,7 @@ up Up.Options {..} = do
      
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid -> do
           updateApp url cid copyFiles
-            |> retryOnErr [status502, status504] 100 
+            >>= retryOnErr [status502, status504] 100 
           >>= \case
             Left err ->
               CLI.Error.put err "Server unable to sync data"
@@ -124,29 +121,3 @@ parseOptions = do
     ]
 
   return Up.Options {..}
-
-requestWithRetry ::
-  ( MonadWebClient m
-  , MonadLogger m
-  )
-  => m (ClientM a)
-  -> m (Either ClientError a)
-requestWithRetry req =
-  sendRequestM req >>= \case
-    Right val ->
-      return $ Right val
-
-    Left err@(FailureResponse _req res) -> do
-      let code = statusCode $ responseStatusCode res
-
-      if code >= 502 && code <= 504
-        then do
-          logWarn $ "Got a " <> textDisplay code <> "; retrying..."
-          req |> 
-            retryOnErr [status502, status504] 100
-
-        else
-          return $ Left err
-
-    Left err ->
-      return $ Left err
--- a/library/Fission/CLI/Command/Watch.hs
+++ b/library/Fission/CLI/Command/Watch.hs
@@ -43,7 +43,9 @@ import qualified Fission.CLI.Command.App.Init as App.Init
 import           Fission.CLI.Command.Types
 import           Fission.CLI.Command.Watch.Types as Watch
 
+import qualified Fission.CLI.Display.Error   as CLI.Error
 import qualified Fission.CLI.Display.Success as CLI.Success
+
 import qualified Fission.CLI.Prompt.BuildDir as Prompt
 
 
@@ -103,12 +105,37 @@ watcher runner Watch.Options {..} = do
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid@(CID hash) -> do
         UTF8.putText $ "👀 Watching " <> Text.pack absPath <> " for changes...\n"
 
-        withRetryM 100 (sendRequestM (updateApp url cid copyFiles)) >>= putErrOr \_ -> do
+        retrier (updateApp url cid copyFiles) >>= putErrOr \_ -> do
           liftIO $ FS.withManager \watchMgr -> do
             hashCache <- newMVar hash
             timeCache <- newMVar =<< getCurrentTime
             void $ handleTreeChanges runner url copyFiles timeCache hashCache watchMgr absPath
             forever . liftIO $ threadDelay 1000000 -- Sleep main thread
+
+retrier ::
+  ( MonadWebClient m
+  , MonadLogger m
+  )
+  => m (ClientM a)
+  -> m (Either ClientError a)
+retrier req =
+  sendRequestM req >>= \case
+    Right val ->
+      return $ Right val
+
+    Left err@(FailureResponse _req res) -> do
+      let code = statusCode $ responseStatusCode res
+ 
+      if code >= 502 && code <= 504
+        then do
+          logWarn $ "Got a " <> textDisplay code <> "; retrying..."
+          retrier req
+
+        else
+          return $ Left err
+
+    Left err ->
+      return $ Left err
 
 updateApp ::
   ( MonadIO      m
--- a/library/Fission/CLI/Command/Watch.hs
+++ b/library/Fission/CLI/Command/Watch.hs
@@ -22,8 +22,8 @@ import           Options.Applicative.Simple hiding (command)
 import           System.FSNotify as FS
 
 import           Fission.Prelude hiding (handle)
+import           Fission.Error
 
-import           Fission.Error.NotFound.Types
 import           Fission.URL
 import           Fission.Models
 import qualified Fission.Time                 as Time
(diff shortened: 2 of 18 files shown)


expede:
@dholms this is the WIP I was talking about

library/Fission/CLI/Command/Up.hs
```diff
@@ -82,7 +92,7 @@ up Up.Options {..} = do
       logDebug $ "Starting single IPFS add locally of " <> displayShow absPath
      
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid -> do
-        sendRequestM (updateApp url cid copyFiles) >>= \case
+        retrier (updateApp url cid copyFiles) >>= \case
```
expede, line 95: `retrier` is really not a great name 😆

library/Fission/CLI/Command/Up.hs
```diff
@@ -111,3 +121,28 @@ parseOptions = do
...
+  -> m (Either ClientError a)
+requestWithRetry req =
+  sendRequestM req >>= \case
+    Right val ->
+      return $ Right val
+
+    Left err@(FailureResponse _req res) -> do
+      let code = statusCode $ responseStatusCode res
+
+      if code >= 502 && code <= 504
```
expede, line 139: I mean... this should be `if code == 502 || code == 504`

library/Fission/CLI/Command/Up.hs
```diff
@@ -111,3 +121,28 @@ parseOptions = do
...
+  )
+  => m (ClientM a)
+  -> m (Either ClientError a)
+requestWithRetry req =
+  sendRequestM req >>= \case
+    Right val ->
+      return $ Right val
+
+    Left err@(FailureResponse _req res) -> do
+      let code = statusCode $ responseStatusCode res
```
expede, line 137: Could probably just destructure

library/Fission/CLI/Command/Watch.hs
```diff
@@ -103,13 +105,38 @@ watcher runner Watch.Options {..} = do
...
+      if code >= 502 && code <= 504
+        then do
+          logWarn $ "Got a " <> textDisplay code <> "; retrying..."
+          retrier req
+
+        else
+          return $ Left err
+
+    Left err ->
+      return $ Left err
```
expede, line 138: I believe that this is duplicated — could probably abstract it out, with an extra parameter for the predicate. I can see us retrying other things based on some criteria (in this case the status code).

But also maybe doesn't have to happen _right now_

library/Fission/Error.hs
```diff
@@ -25,3 +27,20 @@ fromMaybe err okHandler = maybe (openLeft err) (Right . okHandler)
...
+retryForever :: Monad m => (err -> m _ignored) -> m (Either err a) -> m a
+retryForever onErr action = go
+  where
+    go = action >>= \case
+      Right val ->
+        return val
+
+      Left err -> do
+        _ <- onErr err
+        go
```
expede, line 46: I'm not sure that these are actually used anymore 😅 (from previous commits on this branch)

library/Fission/CLI/Command/Up.hs
```diff
@@ -82,13 +92,15 @@ up Up.Options {..} = do
...
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid -> do
-        sendRequestM (updateApp url cid copyFiles) >>= \case
-          Left err ->
-            CLI.Error.put err "Server unable to sync data"
-           
-          Right _  -> do
-            CLI.Success.live cid
-            CLI.Success.dnsUpdated url
+          updateApp url cid copyFiles
+            |> retryOnErr [status502, status504] 100
```
dholms, line 96: It feels like this indentation is kinda wonky but it's the only way that it would parse correctly

library/Fission/URL/DomainName/Types.hs
```diff
@@ -5,7 +5,7 @@ import qualified RIO.Char            as Char
 import qualified RIO.Text            as Text
 
 import           Database.Persist.Postgresql hiding (get)
-import           Data.Swagger                hiding (get)
+import           Data.Swagger                hiding (get, host)
```
dholms, line 8: was getting a compiler warning from this. not technically part of this PR but I wanted it to go away so :man_shrugging:

library/Fission/Web/Client/Error.hs
```diff
@@ -0,0 +1,36 @@
...
+import Network.HTTP.Types.Status
+
+retryOnErr ::
+  ( MonadWebClient m
+  , MonadLogger m
+  )
+  => [Status]
+  -> Integer
+  -> ClientM a
+  -> m (Either ClientError a)
```
expede, line 16:
```haskell
retryOnErr ::
  Monad m
  => [a -> Bool]
  -> Natural
  -> m a
  -> m a
```

expede:
```
retryOnErr ::
  Monad m
  => (a -> Bool)
  -> Natural
  -> m a
  -> m a
```

expede:
```haskell
        
retryOnErrBrooke doesMatch 100 (sendRequestM (updateApp url cid copyFiles))


doesMatch = undefined -- [status502, status504] 
```

expede:
```ts
async retryOnErrBrooke(doesMatch, tries, action) {
  // ...
  action()
}
```

expede:
```haskell
retryOnErrBrooke doesMatch 100 do
  sendRequestM (updateApp url cid copyFiles)
```

expede:
```haskell
retryOnErr predicate tries action = do
  result <- action
  if pred result
    then return result
    else retryOnErr predicate (tries - 1) action

File: library/Fission/CLI/Command/Up.hs (hunk 6 above)
Comment (expede) [reply library/Fission/CLI/Command/Up.hs]: @dholms This comment is outdated, but the way to make the notation consistent here is:

```haskell
copyFiles
  |> updateApp url cid
  |> retryOnErr [status502, status504] 100 
```

The 2-space indentation is _required_ in `do` notation, since every line is sugar for `>>=`. It is consistent this way.

The other way to write this is a `let` or `where`:

```haskell
let req = updateApp url cid copyFiles
retryOnErr [status502, status504] 100 req

--

foo = retryOnErr [status502, status504] 100 req
  where
    req = updateApp url cid copyFiles
```

I actually think that this reads better than pipes, since we're not "just" transforming data.

library/Fission/URL/DomainName/Types.hs
expede: Weird

library/Fission/CLI/Command/Watch.hs
```diff
@@ -102,32 +103,14 @@ watcher runner Watch.Options {..} = do
...
-  , ServerDID    m
-  , MonadWebAuth m Token
-  , MonadWebAuth m Ed25519.SecretKey
-  )
-  => URL
-  -> CID
-  -> Bool
-  -> m (ClientM NoContent)
-updateApp url cid copyFiles =
-  authClient (Proxy @App.Update)
```
dholms, line 126: I remember client functions being split out in their own modules. Is there a reason we have them sitting closer to where they're used?

I split this one out again to increase reusability (used in both `Watch` & `Up`)

package.yaml
```diff
@@ -94,6 +94,7 @@ dependencies:
   - binary
   - bcrypt
   - bytestring
+  - containers
```
dholms, line 97: added for `Data.Sequence` for web request Fixtures

test/Test/Fission/Error.hs
```diff
@@ -0,0 +1,49 @@
...
+
+      describe "retries on an invalid result" do 
+        it "retries 100 times" do
+          L.length failEffectLog `shouldBe` 101
+
+        it "should return the invalid result after retrying" do
+          failResult `shouldBe` 6
+
+
+equal5 ::
```
dholms, line 42: Just chose a very simple arbitrary check function

library/Fission/Internal/Fixture/Web.hs
```diff
@@ -0,0 +1,36 @@
...
+fissionURL = BaseUrl (Https) "example.com" 443 "/"
+
+failureResp :: Status -> ClientError
+failureResp status = mkFailureResponse fissionURL defaultRequest $
+  Response 
+    { responseStatusCode = status
+    , responseHeaders = Seq.empty
+    , responseHttpVersion = http11
+    , responseBody = ""
+    }
```
expede, line 28: Instead of doing this by hand, you could use a helper function like https://www.stackage.org/haddock/lts-16.2/wai-3.2.2.1/Network-Wai.html#v:responseLBS

This would also remove the need for `containers`

package.yaml
expede: See note on that file

test/Test/Fission/Error.hs
```diff
@@ -0,0 +1,49 @@
+module Test.Fission.Error (tests) where
+
+import Test.Fission.Prelude as Mock
+import Fission.Error
+import Data.List as L
```
expede, line 5: A couple minor things:

Ideally use more than one character for legibility. `List.foo` is better than `L.foo` in almost all cases.

They're the same underneath, but for consistency, it probably best to keep to the RIO versions of everything

```haskell
import qualified RIO.List as List

List.length successEffectLog `shouldBe` 1
```

or in this case, the general [`length`](https://www.stackage.org/haddock/lts-16.2/base-4.13.0.0/Prelude.html#v:length) function from `base` (should be imported by default)

```haskell
length successEffectLog `shouldBe` 1
```

library/Fission/Internal/Fixture/Web.hs
dholms: yeah the servant `Reponse` type (https://hackage.haskell.org/package/servant-client-core-0.17/docs/Servant-Client-Core.html#t:Response) is different from the Network.Wai `Response` type :unamused: 
I've been looking for a better way to do this

expede: Ah, okay I see. Yeah, this is probably intended by them to be internal.

`ServerError`s (e.g. `err404`) can be converted very easily with `responseServerError :: ServerError -> Response`.

Success cases lemme look 👀

expede: Okay, yeah you do need to build that manually.

On the plus side, we can get rid of the dependency by leaning into `RIO`:

https://www.stackage.org/haddock/lts-15.5/rio-0.1.14.1/RIO-Seq.html

expede: In general I'd say take a look at RIO first. It has a bunch of really useful stuff in there.

dholms: ah good looks! that's perfect

library/Fission/CLI/Command/Watch.hs
```diff
@@ -102,32 +103,14 @@ watcher runner Watch.Options {..} = do
 
       IPFS.addDir ignoredFiles absPath >>= putErrOr \cid@(CID hash) -> do
         UTF8.putText $ "👀 Watching " <> Text.pack absPath <> " for changes...\n"
-
-        sendRequestM (updateApp url cid copyFiles) >>= putErrOr \_ -> do
+        req <- App.update url cid copyFiles
+        retryOnStatus [status502, status504] 100 req >>= putErrOr \_ -> do
```
expede, line 107: Wait, does this work? Shouldn't this be...

```haskell
let req = App.update url cid copyFiles
retryOnStatus [status502, status504] 100 req
```

...because otherwise, you're just passing the result of the first request back through repeatedly, no?

library/Fission/CLI/Command/Watch.hs
expede: Because it was only being used in the one place. It's specific to this use case. But yeah, sure own module is fine 👍

library/Fission/Error.hs
```diff
@@ -25,3 +27,18 @@ fromMaybe err okHandler = maybe (openLeft err) (Right . okHandler)
...
+  -> m a
+  -> m a
+retryOnErr check times action = do
+  result <- action
+  check result >>= \case
+    True -> return result
+    False -> 
+      if times > 0
+        then retryOnErr check (times - 1) action
+        else return result
```
expede, line 44: Two things:

# 1. Formatting

Let's try to keep the formatting consistent:

```haskell
  result <- action
  check result >>= \case
    True -> 
      return result

    False -> 
      if times > 0
        then retryOnErr check (times - 1) action
        else return result
```

# 2. Multiple Heads

The current version is pretty imperative. The idiomatic way to write this is:

```haskell
retryOnErr _     0          action = action
retryOnErr check retryCount action =
  action >>= check >>= \case
    True  -> return result
    False -> retryOnErr check (times - 1) action
```

or

```haskell
retryOnErr _     0          action = action
retryOnErr check retryCount action =
  ok <- check <$> action
  if ok
    then return result
    else retryOnErr check (times - 1) action
```

library/Fission/Internal/Fixture/Web.hs
```diff
@@ -0,0 +1,36 @@
...
+
+import Fission.Prelude
+
+import Servant.Client
+import Servant.Client.Core.Request
+import Servant.Client.Internal.HttpClient
+
+import Network.HTTP.Types.Status
+import Network.HTTP.Types.Version
+import RIO.Seq as Seq
```
expede, line 16: Nitpick: I'd make this `import qualified`, since we're only using it that way. Prevents pain down the road with ambiguous imports. Typically importing with `as` will have a `qualified` unless you're re-exporting the module.

library/Fission/Internal/Mock/Effect/Types.hs
```diff
@@ -143,3 +145,7 @@ data DestroyLoosePin
   = DestroyLoosePin     UserId IPFS.CID
   | DestroyLoosePinById UserId LoosePinId
   deriving (Eq, Show)
+
+data APICall 
+  = APICall
```
expede, line 150: Let's perhaps be clearer about this one. `FissionWebRequest` or the more general `WebRequest` if it can also do that. Thoughts?

library/Fission/Web/Client/Error.hs
```diff
@@ -0,0 +1,39 @@
...
+checkStatus retryOn = \case
+  Right _ -> return True
+  Left _err@(FailureResponse _req res) -> do
+    let code = responseStatusCode res
+    if elem code retryOn 
+      then do
+        logWarn $ "Got a " <> textShow code <> "; retrying..."
+        return False
+      else return True
+  Left _ -> return True
```
expede, line 39: Nitpick, but for consistency:

```haskell
checkStatus :: 
  MonadLogger m
  => [Status] 
  -> Either ClientError a 
  -> m Bool
checkStatus retryOn = \case
  Right _ -> 
    return True

  Left _err@(FailureResponse _req res) -> do
    let code = responseStatusCode res

    if elem code retryOn 
      then do
        logWarn $ "Got a " <> textShow code <> "; retrying..."
        return False

      else 
        return True

  Left _ ->
    return True
```

expede: Since you're not using the `_err`, you can remove it and the `@`:

```haskell
Left (FailureResponse _req res) ->
```

test/Test/Fission/Error.hs
```diff
@@ -0,0 +1,48 @@
...
+import Test.Fission.Prelude as Mock
+import Fission.Error
+
+
+tests :: IO TestTree
+tests = do
+
+  Mock.Session 
+    { effectLog = successEffectLog :: [OpenUnion '[LogMsg]]
+    , result = successResult :: (Natural)
```
expede, line 12: Remove the parentheses — they're meaningless here

library/Fission/CLI/Command/Watch.hs
dholms: I thought the same thing at first. But `App.update` returns type `m (ClientM NoContent)`. So this is unwrapping the top `m` but isn't touching the actual `ClientM` monad

`retryOnStatus` runs the actual `sendRequest` which is the function that does the monadic action

`sendRequestM` is actually a function that unwraps the top layer of the monad before passing it to `sendRequest`
```haskell
sendRequestM :: MonadWebClient m => m (ClientM a) -> m (Either ClientError a)
sendRequestM clientAction = sendRequest =<< clientAction
```

test/Test/Fission/Error.hs
```diff
@@ -0,0 +1,48 @@
...
+
+  Mock.Session 
+    { effectLog = successEffectLog :: [OpenUnion '[LogMsg]]
+    , result = successResult :: (Natural) 
+    } <- runMock defaultConfig do
+        retryOnErr equal5 100 $ pure 5
+
+  Mock.Session 
+    { effectLog = failEffectLog :: [OpenUnion '[LogMsg]]
+    , result = failResult :: (Natural)
```
expede, line 18: Same here

library/Fission/CLI/Command/Watch.hs
dholms: hoping I have this straight in my head, so let me know if that sounds right!

test/Test/Fission/Error.hs
```diff
@@ -0,0 +1,48 @@
...
+        it "should return the invalid result after retrying" do
+          failResult `shouldBe` 6
+
+
+equal5 :: 
+  MonadLogger m 
+  => Natural 
+  -> m Bool
+equal5 num = do
+  logWarn ("log" :: Text)
```
expede, line 46: Protip: you can also write this as `logWarn @Text "log"`. The `@Text` fills in the type variable for `logWarn`, and reads nicely. They're totally equivalent, so totally about style and which you prefer.

test/Test/Fission/Web/Client/Error.hs
```diff
@@ -0,0 +1,53 @@
...
+      it "is false when a disallowed error status code is returned" do
+        errResult `shouldBe` False
+
+      it "is true when an allowed error status code is returned" do
+        allowedErrResult `shouldBe` True
+
+
+
+
+
```
expede, line 53: Empty lines?

test/Test/Fission/Web/Client/Error.hs
```diff
@@ -0,0 +1,53 @@
+module Test.Fission.Web.Client.Error (tests) where
+
+import           Test.Fission.Prelude as Mock
+
+
+import Fission.Web.Client.Error
+import Network.HTTP.Types.Status
```
expede, line 7: Align these please! If you're not already, installing stylish-haskell and a plugin for VSCode should autoformat your imports.

https://github.com/vigoo/stylish-haskell-vscode
https://github.com/jaspervdj/stylish-haskell

test/Test/Fission/Web/Client/Error.hs
```diff
@@ -0,0 +1,53 @@
...
+        checkStatus [status502, status504] (Right (0 :: Integer))
+
+  Mock.Session 
+    { effectLog = _effectLog :: [OpenUnion '[LogMsg]]
+    , result = errResult 
+    } <- runMock defaultConfig do
+        checkStatus [status502, status504] (Left Fixture.failure502)
+
+  Mock.Session 
+    { effectLog = _effectLog :: [OpenUnion '[LogMsg]]
```
expede, line 30: If you're tossing the `_effectLog`, you may not need to draw the values (unless the compiler complains)

```haskell
Mock.Session {result = errResult } <- runMock defaultConfig do
  checkStatus [status502, status504] (Left Fixture.failure502)
```

library/Fission/CLI/Command/Watch.hs
expede: Ah, yes I see — it's because of the type of `authClient`. We need the `m` because we want ambient access to config variables (for the auth part without having to manually pass them around).

This may be clearer with naming:

```haskell
App.mkUpdate

-- or

App.updateM
```

or similar

dholms: ah cool I like it :+1:

library/Fission/Error.hs
dholms: oh yeah good point on the base case^^

I think I still need to get `result` out of `action` however. So:
```haskell
retryOnErr check times action = do
  result <- action
  check result >>= \case
    True -> return result
    False -> retryOnErr check (times - 1) action
```
altho it sounds like I should avoid blending `do` notation & `bind` notation
so maybe:
```haskell
retryOnErr _ 0 action = action
retryOnErr check times action = do
  result <- action
  ok <- check result
  if ok
    then return result
    else retryOnErr check (times - 1) action
````

Comment (expede) [reply library/Fission/Error.hs:44]:
```haskell
result <- action
check result >>= \case

-- is just sugar for

action >>= check >>= \case
```

library/Fission/Internal/Mock/Effect/Types.hs
dholms: Actually... we got rid of the mock that used this so I'm just going to get rid of the Effect type as well

library/Fission/Error.hs
expede: I'm fine with this version of yours:

```haskell
retryOnErr _ 0 action = action
retryOnErr check times action = do
  result <- action
  ok <- check result
  if ok
    then return result
    else retryOnErr check (times - 1) action
```

You're right about the types in my example with `<$>`. We could write it this way:

```haskell
do
  ok <- check =<< action
  if ok
    then return result
    else retryOnErr check (times - 1) action
```

or even this way if you're going to reuse that a lot in the body (probably not this case)

```haskell
do
  ok <- checkM action
  if ok
    then return result
    else retryOnErr check (times - 1) action
where
  checkM foo = check =<< foo
```

expede: These are all the same to the compiler. Which reads best to you?

test/Test/Fission/Error.hs
dholms: yes much nicer. thanks!

test/Test/Fission/Web/Client/Error.hs
dholms: :man_shrugging: :hocho:

test/Test/Fission/Web/Client/Error.hs
dholms: nope unfortunately the compiler complains
```
    • You require open sum type to contain the following element:
          LogMsg
      However, given list can store elements only of the following types:
          effs0
```

library/Fission/Error.hs
dholms: but in the `if` statement  you need to return `result` so you can't just bind it through to `check`

I think our options are:
```haskell
retryOnErr _ 0 action = action
retryOnErr check times action = do
  result <- action
  ok <- check result
  if ok
    then return result
    else retryOnErr check (times - 1) action
```
or
```haskell
retryOnErr _ 0 action = action
retryOnErr check times action = 
  action >>= \result ->
    check result >>= \case
      True -> return result
      False -> retryOnErr check (times - 1) action
```
if we want bind syntax

expede: Oooh, right we need to return `result`! Right right. I'd use the `do` version 👍

library/Fission/Internal/Fixture/Web.hs
```diff
@@ -0,0 +1,36 @@
...
+    , responseHttpVersion = http11
+    , responseBody = ""
+    }
+
+
+failure502 :: ClientError
+failure502 = failureResp status502
+
+failure504 :: ClientError
+failure504 = failureResp status504
```
expede, line 35: Leave this as is, but it could also read nicely as something like

`failWith status504`

...rather than specific functions.

library/Fission/Internal/Mock/Types.hs
```diff
@@ -190,9 +190,8 @@ instance MonadReflectiveServer (Mock effs) where
   getHost = Web.Host <$> parseBaseUrl "example.com"
 
 instance IsMember LogMsg effs => MonadLogger (Mock effs) where
-  monadLoggerLog loc src lvl msg = do
+  monadLoggerLog _loc _src lvl msg = do
```
expede, line 193: 👍 Good point

library/Fission/Web/Client/App.hs
```diff
@@ -20,3 +33,21 @@ type Update
...
+  )
+  => URL
+  -> CID
+  -> Bool
+  -> m (ClientM NoContent)
+mkUpdateReq url cid copyFiles =
+  authClient (Proxy @Update)
+    `withPayload` url
+    `withPayload` cid
+    `withPayload` (Just copyFiles)
```
expede, line 53: Probably don’t need the parens around `(Just copeFiles)`

expede:
Just a SUPER minor suggestion. This one is good to go!!

LOLOLOL it won’t let me approve it because I opened it 😆  @dholms this one is all yours to merge if/when you’re happy with it!

> Opened 20 days ago

WUUUUUUUUUUUUT

Issue #342, opened 2020-06-22
Docs say "Apache 2"


expede:
# Summary

## Problem

Our docs (runfission.com/docs) claim that we're licensed Apache 2, but we're actually AGPL 😅

## Solution

Fix... that

PR #345, opened 2020-06-24
License

dholms:
## Problem
We're licensed under AGPL but the swagger docs say Apache

## Solution
Update license in swagger docs

Closes https://github.com/fission-suite/fission/issues/342#event-3470176874


nikoPLP:
still no license file appearing in your repo

expede:
@nikoPLP Correct not at the root! There are licenses in each package in the repo.

https://github.com/fission-suite/fission/blob/main/fission-cli/LICENSE
https://github.com/fission-suite/fission/blob/main/fission-web-api/LICENSE

Issue #336, opened 2020-06-11
Increase CNAME caching time


expede:
# Summary

## Problem

Currently, `foo.fission.app` and friends are only cached for 10s. This makes sense for TXT records, but CNAMEs can be cached for _much_ longer.

### Impact

Worse performance

## Solution

Increase DNS cache length to 24 hours(?)

PR #346, opened 2020-06-24
Increase TTL for CNAME records

dholms:
## Problem
Currently, `foo.fission.app` and friends are only cached for 10s. This makes sense for TXT records, but CNAMEs can be cached for much longer.

## Solution
Increase DNS cache length to 24 hours

Closes https://github.com/fission-suite/fission/issues/336

### DIFF
--- a/library/Fission/Types.hs
+++ b/library/Fission/Types.hs
@@ -145,7 +145,7 @@ instance MonadRoute53 Fission where
 
         return $ changeResourceRecordSets (ResourceId zoneID) batch
 
-  set recType url@URL {..} (ZoneID zoneTxt) contents = do
+  set recType url@URL {..} (ZoneID zoneTxt) contents ttl = do
     logDebug $ "Updating DNS record at: " <> displayShow url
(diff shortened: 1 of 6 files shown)


library/Fission/Types.hs
```diff
@@ -204,12 +204,12 @@ instance MonadRoute53 Fission where
...
 instance MonadDNSLink Fission where
   set _userId url@URL {..} zoneID (IPFS.CID hash) = do
-    Route53.set Cname url zoneID (pure $ textDisplay gateway) >>= \case
+    Route53.set Cname url zoneID (pure $ textDisplay gateway) 86400 >>= \case
       Left err ->
         return $ Error.openLeft err
 
       Right _ ->
-        Route53.set Txt dnsLinkURL zoneID (pure dnsLink) <&> \case
+        Route53.set Txt dnsLinkURL zoneID (pure dnsLink) 10 <&> \case
```
expede, line 212: 👍 Good idea

expede: Not for right now, but down the line, we may want to make this user configurable, too. Like "on production, set TTL to an hour" or something.

Issue #341, opened 2020-06-22
Add route to set user's exchange keys


dholms:
## Problem
User's need to be able to share their public exchange keys over DNS

## Solution
Add a `PATCH /user/exchange` root

This will function similarly to `PATCH /user/did`

It should take an array of `did:key`s as a parameter, and update the TXT record at `_exchange.${username}.fission.name` to include those keys

PR #348, opened 2020-06-25
Exchange keys route

dholms:
## Problem
User's need to be able to share their public exchange keys over DNS

## Solution
Add a PATCH /user/exchange root

This will function similarly to PATCH /user/did

It should take an array of did:keys as a parameter, and update the TXT record at `_exchange.${username}.fission.name` to include those keys

Closes https://github.com/fission-suite/fission/issues/341

### DIFF
--- a/library/Fission/Internal/Fixture/User.hs
+++ b/library/Fission/Internal/Fixture/User.hs
@@ -10,14 +10,11 @@ import qualified Fission.User.Role.Types as User.Roles
 
 import           Fission.Internal.Fixture.Time
 import qualified Fission.Internal.Fixture.Key.Ed25519 as Ed25519
+import qualified Fission.Internal.RSA2048.Pair.Types  as RSA2048
 
 user :: User
 user = User
-  { userPublicKey = Just Ed25519.pk
-
-  --
-
-  , userEmail    = Just "test@fission.codes"
+  { userEmail    = Just "test@fission.codes"
   , userUsername = "testUser"
 
   --
@@ -26,6 +23,10 @@ user = User
   , userActive = True
   , userVerified  = False
 
+  --
+
+  , userPublicKey    = Just Ed25519.pk
+  , userExchangeKeys = [RSA2048.pk1]
   --
 
   , userDataRoot     = CID "QmW2WQi7j6c7UgJTarActp7tDNikE4B2qXtFCfLPdsgaTQ"
--- a/library/Fission/Internal/Mock/Types.hs
+++ b/library/Fission/Internal/Mock/Types.hs
@@ -259,6 +259,14 @@ instance IsMember ModifyUser effs => User.Modifier (Mock effs) where
     Effect.log $ ModifyUser uID
     return $ Right newPK
 
+  addExchangeKey uID key _ = do
+    Effect.log $ ModifyUser uID
+    return $ Right [key]
+
+  removeExchangeKey uID _ _ = do
+    Effect.log $ ModifyUser uID
+    return $ Right []
+
   setData uID _ _ = do
     Effect.log $ ModifyUser uID
     return ok
--- a/library/Fission/Internal/RSA2048/Pair/Types.hs
+++ b/library/Fission/Internal/RSA2048/Pair/Types.hs
@@ -1,4 +1,13 @@
-module Fission.Internal.RSA2048.Pair.Types (Pair (..)) where
+module Fission.Internal.RSA2048.Pair.Types 
+  ( Pair (..)
+  , pregenerated
+  , pk1
+  , sk1
+  , pk2
+  , sk2
+  , pk3
+  , sk3
+  ) where
 
 import Crypto.PubKey.RSA
 
--- a/library/Fission/Models.hs
+++ b/library/Fission/Models.hs
@@ -24,6 +24,7 @@ import           Fission.Platform.Heroku.Region.Types
 import           Fission.Security
 
 import qualified Fission.Key                          as Key
+import qualified Crypto.PubKey.RSA                    as RSA
 import           Fission.URL
 
 import           Fission.User.Email.Types
@@ -34,8 +35,9 @@ import           Fission.Challenge.Types
 import qualified Fission.AWS.Zone.Types               as AWS
 
 import           Fission.Internal.Orphanage.Bytes     ()
-import           Fission.Internal.Orphanage.CID       ()
-import           Fission.Internal.Orphanage.UUID      ()
+import           Fission.Internal.Orphanage.CID            ()
+import           Fission.Internal.Orphanage.UUID           ()
+import           Fission.Internal.Orphanage.RSA2048.Public ()
 
 share
   [ mkPersist       sqlSettings
@@ -57,6 +59,7 @@ HerokuAddOn
 
 User
   publicKey     Key.Public    Maybe
+  exchangeKeys  [RSA.PublicKey]        default=[]
 
   email         Email         Maybe
   username      Username
--- a/library/Fission/Types.hs
+++ b/library/Fission/Types.hs
@@ -62,10 +62,13 @@ import qualified Fission.Web.Auth.Token                as Auth.Token
 import           Fission.Web.Handler
 import           Fission.Web.Server.Reflective         as Reflective
 
-import           Fission.User.DID.Types
-import qualified Fission.User          as User
+import           Fission.User.DID            as DID
+import qualified Fission.User                as User
 import           Fission.User.Creator.Class
 import qualified Fission.User.Password as Password
+import qualified Fission.User.Modifier.Class as User.Modifier
+
+import qualified Fission.Key as Key
 
 import           Fission.Web.Auth.Token.Basic.Class
 import qualified Fission.Web.Auth.Token.JWT.RawContent as JWT
@@ -407,37 +410,34 @@ instance User.Creator Fission where
           Right _  -> Right userId
 
 instance User.Modifier Fission where
-  updatePassword userId pass now =
+  updatePassword uID pass now =
     Password.hashPassword pass >>= \case
       Left err ->
         return $ Left err
 
       Right secretDigest -> do
-        _ <- runDB $ User.updatePasswordDB userId secretDigest now
+        _ <- runDB $ User.updatePasswordDB uID secretDigest now
         return $ Right pass
 
-  updatePublicKey userId pk now =
-    runDB (User.getById userId) >>= \case
-      Nothing -> 
-        return . Error.openLeft $ NotFound @User
+  updatePublicKey uID pk now = 
+    runUserUpdate updatePK pkToText uID "_did"
+    where
+      updatePK = User.updatePublicKeyDB uID pk now
+      pkToText pk' = textDisplay (DID Key pk')
 
-      Just (Entity _ User { userUsername = Username rawUN }) -> do
-        _ <- runDB $ User.updatePublicKeyDB userId pk now
+  addExchangeKey uID key now =
+    runUserUpdate addKey keysToText uID "_exchange"
+    where
+      addKey = User.addExchangeKeyDB uID key now
+      keysToText keys = Text.intercalate "," (textDisplay . DID Key . Key.RSAPublicKey <$> keys)
 
-        domainName <- asks userRootDomain
-        zoneID     <- asks userZoneID
-
-        let
-          subdomain = Just $ Subdomain rawUN
-          url       = URL {domainName, subdomain = Just (Subdomain "_did") <> subdomain}
-          did       = textDisplay (DID pk Key)
-          didSegments = DNS.splitRecord did
-
-        Route53.set Txt url zoneID didSegments 10 <&> \case
-          Left serverErr -> Error.openLeft serverErr
-          Right _        -> Right pk
-
-  setData userId newCID now = 
+  removeExchangeKey uID key now =
+    runUserUpdate removeKey keysToText uID "_exchange"
+    where
+      removeKey = User.removeExchangeKeyDB uID key now
+      keysToText keys = Text.intercalate "," (textDisplay . DID Key . Key.RSAPublicKey <$> keys)
+      
+  setData userId newCID now = do
     runDB (User.getById userId) >>= \case
       Nothing ->
         return . Error.openLeft $ NotFound @User
@@ -547,40 +547,6 @@ instance App.Destroyer Fission where
     runDB (App.destroyByURL uId domainName maySubdomain now) >>= \case
       Left err   -> return $ Left err
       Right urls -> pullFromDNS urls
-  
-pullFromDNS :: [URL] -> Fission (Either App.Destroyer.Errors [URL])
-pullFromDNS urls = do
-  domainsAndZoneIDs <- runDB . select $ from \domain -> do
-    where_ $ domain ^. DomainDomainName `in_` valList (URL.domainName <$> urls)
-    return (domain ^. DomainDomainName, domain ^. DomainZoneId)
-
-  let
-    zonesForDomains :: [(DomainName, ZoneID)]
-    zonesForDomains =
-      domainsAndZoneIDs <&> \(SQL.Value domain, SQL.Value zone) -> (domain, zone)
-
-  foldM (folder zonesForDomains) (Right []) urls
- 
-  where
-    folder ::
-         [(DomainName, ZoneID)]            -- ^ Hosted zone map
-      -> Either App.Destroyer.Errors [URL] -- ^ Accumulator
-      -> URL                               -- ^ Focus
-      -> Fission (Either App.Destroyer.Errors [URL])
-
-    folder _ (Left err) _ =
-      return $ Left err
-
-    folder zonesForDomains (Right accs) url@URL {..} = do
-      case lookup domainName zonesForDomains of
-        Nothing -> do
-          logError $ "Unable to find zone for " <> textDisplay domainName
-          return . Error.openLeft $ NotFound @ZoneID
-
-        Just zoneId ->
-          AWS.clear Txt url zoneId <&> \case
-            Left err -> Error.openLeft err
-            Right _  -> Right (url : accs)
 
 instance Heroku.AddOn.Creator Fission where
   create uuid region now = runDB $ Heroku.AddOn.create uuid region now
@@ -619,3 +585,65 @@ instance MonadEmail Fission where
         }
 
     liftIO $ runClientM (Email.sendEmail apiKey emailData) env
+  
+pullFromDNS :: [URL] -> Fission (Either App.Destroyer.Errors [URL])
+pullFromDNS urls = do
+  domainsAndZoneIDs <- runDB . select $ from \domain -> do
+    where_ $ domain ^. DomainDomainName `in_` valList (URL.domainName <$> urls)
+    return (domain ^. DomainDomainName, domain ^. DomainZoneId)
+
+  let
+    zonesForDomains :: [(DomainName, ZoneID)]
+    zonesForDomains =
+      domainsAndZoneIDs <&> \(SQL.Value domain, SQL.Value zone) -> (domain, zone)
+
+  foldM (folder zonesForDomains) (Right []) urls
+ 
+  where
+    folder ::
+         [(DomainName, ZoneID)]            -- ^ Hosted zone map
+      -> Either App.Destroyer.Errors [URL] -- ^ Accumulator
+      -> URL                               -- ^ Focus
+      -> Fission (Either App.Destroyer.Errors [URL])
+
+    folder _ (Left err) _ =
+      return $ Left err
+
+    folder zonesForDomains (Right accs) url@URL {..} = do
+      case lookup domainName zonesForDomains of
+        Nothing -> do
+          logError $ "Unable to find zone for " <> textDisplay domainName
+          return . Error.openLeft $ NotFound @ZoneID
+
+        Just zoneId ->
+          AWS.clear Txt url zoneId <&> \case
+            Left err -> Error.openLeft err
+            Right _  -> Right (url : accs)
+
+runUserUpdate ::
+     Transaction Fission (Either User.Modifier.Errors a)
+  -> (a -> Text)
+  -> UserId
+  -> Text
+  -> Fission (Either User.Modifier.Errors a)
+runUserUpdate updateDB dbValToTxt uID subdomain =
+  runDB updateDB >>= \case
+    Left err ->
+      return $ Left err
+  
+    Right dbVal -> do
+      runDB (User.getById uID) >>= \case
+        Nothing ->
+          return . Error.openLeft $ NotFound @User
+
+        Just (Entity _ User { userUsername = Username rawUN }) -> do
+          domainName <- asks userRootDomain
+          zoneID     <- asks userZoneID
+
+          let
+            url = URL {domainName, subdomain = Just (Subdomain subdomain) <> Just (Subdomain rawUN)}
+            segments = DNS.splitRecord $ dbValToTxt dbVal
+
+          Route53.set Txt url zoneID segments 10 >>= \case
+            Left serverErr -> return $ Error.openLeft serverErr
+            Right _        -> return $ Right dbVal
--- a/library/Fission/User/DID.hs
+++ b/library/Fission/User/DID.hs
@@ -1,6 +1,3 @@
-module Fission.User.DID
-  ( -- * Reexports
-    module Fission.User.DID.Types
-  ) where
+module Fission.User.DID (module Fission.User.DID.Types) where
 
 import Fission.User.DID.Types
--- a/library/Fission/User/Modifier.hs
+++ b/library/Fission/User/Modifier.hs
@@ -1,6 +1,8 @@
 module Fission.User.Modifier
   ( updatePasswordDB
   , updatePublicKeyDB
+  , addExchangeKeyDB
+  , removeExchangeKeyDB
   , setDataDB
   , module Fission.User.Modifier.Class
   ) where
@@ -9,14 +11,18 @@ import           Fission.User.Modifier.Class
 
 import           Fission.Prelude
 import           Fission.Models
+import           Fission.Error
 
-import           Fission.Key           as Key
 import           Fission.Security.Types
+import qualified Fission.Key       as Key
+import qualified Crypto.PubKey.RSA as RSA
 
 import           Database.Persist as Persist
 
 import           Network.IPFS.CID.Types
 import           Network.IPFS.Bytes.Types
+
+import qualified RIO.List as List
 
 
 updatePasswordDB ::
@@ -36,13 +42,65 @@ updatePublicKeyDB ::
   => UserId
   -> Key.Public
   -> UTCTime
-  -> Transaction m ()
-updatePublicKeyDB userID pk now =
+  -> Transaction m (Either Errors Key.Public)
+updatePublicKeyDB userID pk now = do
   update userID
     [ UserPublicKey  =. Just pk
     , UserModifiedAt =. now
     ]
-  
+
+  return $ Right pk
+
+addExchangeKeyDB ::
+     MonadIO m
+  => UserId
+  -> RSA.PublicKey
+  -> UTCTime
+  -> Transaction m (Either Errors [RSA.PublicKey])
+addExchangeKeyDB userID key now =
+  Persist.get userID >>= \case
+    Nothing -> 
+      return . openLeft $ NotFound @User
+
+    Just user -> do
+      let keys = userExchangeKeys user
+      if List.elem key keys
+        then 
+          return $ Right keys
+
+        else do
+          let updated = keys ++ [key]
+          update userID
+            [ UserExchangeKeys =. updated
+            , UserModifiedAt   =. now
+            ]
+          return $ Right updated
+
+removeExchangeKeyDB ::
+    MonadIO m
+  => UserId
+  -> RSA.PublicKey
+  -> UTCTime
+  -> Transaction m (Either Errors [RSA.PublicKey])
+removeExchangeKeyDB userID key now =
+  Persist.get userID >>= \case
+    Nothing -> 
+      return . openLeft $ NotFound @User
+
+    Just user -> do
+      let keys = userExchangeKeys user
+      if List.elem key keys
+        then do
+          let updated = List.delete key keys
+          update userID
+            [ UserExchangeKeys =. updated
+            , UserModifiedAt   =. now
+            ]
+          return $ Right updated
+
+        else 
+          return $ Right keys
+
 setDataDB ::
      MonadIO m
   => UserId
--- a/library/Fission/User/Modifier.hs
+++ b/library/Fission/User/Modifier.hs
@@ -63,13 +63,14 @@ addExchangeKeyDB userID key now =
       return . openLeft $ NotFound @User
 
     Just user -> do
-      let keys = userExchangeKeys user
+      let 
+        keys = userExchangeKeys user
+        updated = [key] ++ keys
       if List.elem key keys
         then 
           return $ Right keys
 
         else do
-          let updated = keys ++ [key]
           update userID
             [ UserExchangeKeys =. updated
             , UserModifiedAt   =. now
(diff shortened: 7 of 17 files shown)


library/Fission/User/Creator/Class.hs
```diff
@@ -84,6 +84,7 @@ instance
   create username pk email now =
     User
       { userPublicKey     = Just pk
+      , userExchangeKeys  = Just []
```
dholms, line 87: should we pass exchange keys on user creation? Or create with an empty array and then flesh it out later?

library/Fission/Types.hs
```diff
@@ -413,18 +414,37 @@ instance User.Modifier Fission where
...
             Route53.set Txt url zoneID didSegments <&> \case
               Left serverErr -> Error.openLeft serverErr
               Right _        -> Right pk
-    where
-      splitter :: Char -> (Natural, NonEmpty Text) -> (Natural, NonEmpty Text)
-      splitter chr (255, txtList)       = splitter chr (0, "" `cons` txtList)
-      splitter chr (len, (txt :| more)) = (len + 1, (Text.cons chr txt) :| more)
+
+  updateExchangeKeys uID keys now =
+    runDB (User.updateExchangeKeys uID keys now) >>= \case
```
dholms, line 426: These 3 functions (`updatePublicKey`, `updateExchangeKeys` and `setData`) are all very similar, and I feel like we should be able to abstract out those similarities into a function. I've given it a go but I'm having issues with getting the types right on it :thinking:

dholms:
```haskell
runOperation ::
  ( MonadIO          m
  , MonadDB        t m
  , User.Retriever t
  )
  => m (Either User.Modifier.Errors Key.Public)
  -> UserId
  -> Text
  -> NonEmpty Text
  -> m (Either User.Modifier.Errors (NonEmpty Text))
runOperation dbFn uID subdomain txtSegments = 
  runDB dbFn >>= \case
    Left err ->
      return $ Left err
    Right _ -> do
      runDB (User.getById uID) >>= \case
        Nothing -> 
          return . Error.openLeft $ NotFound @User

        Just (Entity _ User { userUsername = Username rawUN }) -> do
          domainName <- asks userRootDomain
          zoneID     <- asks userZoneID

          let 
            url = URL {domainName, subdomain = Just (Subdomain subdomain) <> Just (Subdomain rawUN)}

          Route53.set Txt url zoneID txtSegments <&> \case
            Left serverErr -> Error.openLeft serverErr
            Right _        -> Right txtSegments
```
I think I'm missing something in the constraints

library/Fission/Internal/Text.hs
```diff
@@ -0,0 +1,15 @@
+module Fission.Internal.Text
+  ( split256
+  ) where
+
+import Fission.Prelude
+import RIO.NonEmpty
+import qualified RIO.Text as Text
+
+split256 :: Text -> NonEmpty Text
+split256 txt = snd $ Text.foldr splitter (0, ("" :| [])) txt
```
expede, line 10: https://www.stackage.org/haddock/lts-15.5/rio-0.1.14.1/RIO-Text.html#v:chunksOf

expede:
```haskell
foo :: Text -> Maybe (NonEmpty Text)
foo txt = nonEmpty $ Text.chunksOf 256 txt

-- or

import qualified RIO.NonEmpty.Partial as Partial.NonEmpty

bar :: Text -> NonEmpty Text
bar txt = Partial.NonEmpty.fromList $ Text.chunksOf 256 txt
```

I'd probably do the former, since it's safer. Also safer than the original version.

library/Fission/Models.hs
```diff
@@ -54,6 +54,7 @@ HerokuAddOn
 
 User
   publicKey     Key.Public    Maybe
+  exchangeKeys  [Key.Public]  Maybe
```
expede, line 57: From a purely information perspective

`Maybe [a] ~ [a]`

library/Fission/Web/User/UpdateExchangeKeys.hs
```diff
@@ -0,0 +1,33 @@
...
+  , MonadLogger   m
+  , MonadThrow    m
+  , User.Modifier m
+  )
+  => Authorization
+  -> ServerT API m
+server Authorization {about = Entity userID _} keys = do
+  now <- currentTime
+  Web.Error.ensureM $ User.updateExchangeKeys userID keys now
+  return NoContent
```
expede, line 33: This file looks pretty great 👍

library/Fission/Types.hs
expede: Lemme take a look 👀

library/Fission/Internal/Fixture/User.hs
```diff
@@ -22,8 +22,9 @@ user = User
 
   --
 
-  , userPublicKey = Just Ed25519.pk
-  , userDataRoot  = CID "QmW2WQi7j6c7UgJTarActp7tDNikE4B2qXtFCfLPdsgaTQ"
+  , userPublicKey    = Just Ed25519.pk
+  , userExchangeKeys = Just [Ed25519.pk]
```
expede, line 26: I thought we said that exchange keys need to be RSA2048 for compatability, no?

library/Fission/Types.hs
expede: Also, I'm a bit cautious in this specific case about making them all the same. You're right that the underlying implementation may be similar, but it feels brittle since they're _conceptually_ different effects.

The solution here may be to extract out a function that all 3 call. I'll play with the code you gave above in the meantime here, though.

expede: Working version:

```haskell

runOperation ::
     Fission (Either User.Modifier.Errors Key.Public)
  -> UserId
  -> Text
  -> NonEmpty Text
  -> Fission (Either User.Modifier.Errors (NonEmpty Text))
runOperation dbFn uID subdomain txtSegments =
  dbFn >>= \case
    Left err ->
      return $ Left err
  
    Right _ -> do
      runDB (User.getById uID) >>= \case
        Nothing ->
          return . Error.openLeft $ NotFound @User

        Just (Entity _ User { userUsername = Username rawUN }) -> do
          domainName <- asks userRootDomain
          zoneID     <- asks userZoneID

          let
            url = URL {domainName, subdomain = Just (Subdomain subdomain) <> Just (Subdomain rawUN)}

          Route53.set Txt url zoneID txtSegments <&> \case
            Left serverErr -> Error.openLeft serverErr
            Right _        -> Right txtSegments
```

expede: This probably seems odd. Where are the constraints? Well, in this scenario, we're making it concrete on `Fission`, _which we know_ is a `MonadIO`, `MonadDB`, and so on.

What happened to the `runDB`? Well, we probably want to just call the `User.getById` function or whatever. It calls `runDB` inside it, since we know the context it's going to get run in.

We also don't have to write an abstract version of this for the `Mock`. We just call this handy helper inside `Fission/Types.hs` and don't export it. We do this in a couple places (e.g. the big JWT module).

expede: In this implementation, I'd maybe even rename `dbFn` to `getPublicKey`, which looks like is the intent from the signature.

Also, we can probably tighten up which errors that part is allowed to return by changing to:

```haskell
Fission (Either User.Retriever.Errors Key.Public)
```

Which I believe is a subset of `User.Modifier.Errors`.

If we're always looking for the PK via the DB, we can even hardcode that DB call and drop the parameter.

library/Fission/Internal/Text.hs
dholms: beautiful

library/Fission/Models.hs
dholms: hmm yeah good point :thinking: 

If we have a required col in the Model will it automatically update every row in the DB to have an empty array or will we need to do that manually?

library/Fission/Types.hs
dholms: Nice this is perfect. exactly what I was shooting for!

library/Fission/Internal/Fixture/User.hs
dholms: hmmm yes good point :+1: 

Either that or we need to deterministically derive RSA2048 from Curve25519 keys. But I think just requiring RSA makes more sense

library/Fission/Models.hs
expede: You can set `default=` in the models [Reference](https://github.com/yesodweb/persistent/blob/master/docs/Persistent-entity-syntax.md#change-tablecollection-key-definition-field-name-andor-type-persistent--21)

Probably not needed here, but it's also possible to do a custom translation to/from the database with typeclasses [PersistField](https://www.stackage.org/haddock/lts-16.2/persistent-2.10.5.2/Database-Persist-Class.html#g:6) and friends

library/Fission/Internal/Fixture/User.hs
expede: In either case, the PK needs to be RSA. More flexible to not derive the key in case we need to rotate it? Unclear, but probably the easiest path.

library/Fission/Types.hs
```diff
@@ -394,37 +396,50 @@ instance User.Creator Fission where
           Left err -> Error.relaxedLeft err
           Right _  -> Right userId
 
+runUpdate ::
```
dholms, line 399: Okay 2 questions that I thought of here:

Does this function belong in a different module? Or is it fine here? This file is all interfaces so it seems a bit weird to have this function floating around in here, but in another file, I think we'll have to deal with a circular dependency

Should I try to include make this function work for `setData` as well? It'll definitely add some complexity since it's not quite as similar as the other 2 functions were. It uses `MonadDNSLink` instead of `MonadRoute53` and passes different parameters. I ended up not rewriting cause I thought it was more straightforward this way

expede: > This file is all interfaces so it seems a bit weird to have this function floating around in here

Yup, it's totally fine to have in here! It's not getting exported, so it's like a `private` function in OO. It's specific to the stuff in this module. 

> we'll have to deal with a circular dependency

Yup. You really don't have a choice. I also got stuck with some of this over in `JWT.hs` because it's a recursive structure.

> Should I try to include make this function work for `setData` as well?

I wouldn't worry about it right now. We can always break these functions into subfunctions for the common parts later, but this stuff is all so concrete and internal that I'm not too worried.

> I ended up not rewriting cause I thought it was more straightforward this way

Good call!

## Aside

All of the class instances for `Fission` need to live either here, or where the class is defined. Because we're going to split the repo into subprojects shortly, we don't want the classes to have too much dependency on `Fission`. The dependencies ideally flow the other way.

dholms: Great that all sounds good! :+1:

library/Fission/Internal/Fixture/Key/RSA2048.hs
```diff
@@ -0,0 +1,18 @@
...
+
+import           Fission.Prelude
+import qualified Crypto.PubKey.RSA as RSA
+import           Fission.Internal.Orphanage.RSA2048.Public ()
+
+pk :: RSA.PublicKey
+Right pk = parseUrlPiece rawPK
+
+rawPK :: Text
+rawPK = "MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAoZeA/ijID5iZqPCBJFdml7wfxY/bLKfEIaacpq7CW/Y4bm/v/peLGiRMD1+k9IrmoA0I9h8hczfegaHWCEPovlzxu5YErbtuaI14Yqm90Nq1T0wLNysdjLx995G7w4W+FGNSoIivxfmeY8RKt0v2Z8EykiWA81vSbberXOqyDeAo0cRqbvFVynS4oUUQZOeaKj+SWMbYtG1nAnu5frNknOCIQVqUr6NXEwXmSbk/vO/CqkqHRhv4p9wRWQbk7/GRMiaVu0m9qV4rlD1F9QL7Q9HYi16k7WmpSW7Wlw330ZgH5RLt4pTcuTyGMt1loCBFwPWSO5DbFDwS+EgldePMqwIDAQAB"
```
expede, line 18: We actually have a few of these already (as key pair fixtures). The file is horrifying, but works. Not sure if that helps or not for this use case, though

library/Fission/Internal/Mock/Types.hs
```diff
@@ -260,6 +260,10 @@ instance IsMember ModifyUser effs => User.Modifier (Mock effs) where
     Effect.log $ ModifyUser uID
     return $ Right newPK
 
+  updateExchangeKeys uID newKeys _ = do
+    Effect.log $ ModifyUser uID
+    return $ Right newKeys
```
expede, line 265: 👍

expede: (Can you see me getting lazy with the mock logs :P )

library/Fission/Internal/Text.hs
```diff
@@ -0,0 +1,18 @@
...
+import qualified RIO.Text as Text
+
+
+split :: Int -> Text -> Maybe (NonEmpty Text)
+split len txt = nonEmpty $ Text.chunksOf len txt
+
+split' :: Int -> Text -> NonEmpty Text
+split' len txt = 
+  split len txt
+  |> maybe (pure "") identity
```
expede, line 18: Much more idiomatic: 

```haskell
split’ :: Natural -> Text -> NonEmpty Text
split’ len txt = fromMaybe (pure “”) $ split len txt
```

--- 

Is `””` a valid value for this use case?

library/Fission/Models.hs
```diff
@@ -54,6 +56,7 @@ HerokuAddOn
 
 User
   publicKey     Key.Public    Maybe
+  exchangeKeys  [RSA.PublicKey]        default=[]
```
expede, line 59: 👍

library/Fission/User/DID.hs
```diff
@@ -1,6 +1,13 @@
...
     module Fission.User.DID.Types
   ) where
 
+import Fission.Prelude
 import Fission.User.DID.Types
+
+import qualified Fission.Key as Key
+
+txtFromKey :: Key.Public -> Text
+txtFromKey pk = textDisplay (DID pk Key)
```
expede, line 13: The convention would be to call this `toText`.

That said, what we actually want here is `textDisplay`:

```haskell
instance Display Key.Public where
  textDisplay pk = textDisplay (DID pk Key)
```

library/Fission/Internal/Fixture/Key/RSA2048.hs
dholms: >We actually have a few of these already (as key pair fixtures)

oh really? where?

library/Fission/Internal/Text.hs
dholms: > Is ”” a valid value for this use case?

1) we'll actually be ditching this method in favor of `splitRecords` from the `ordered-txt` branch
2) kind of.... I actually went back and forth on how to handle it. It's a state that shouldn't ever be reached (unless you create a CID/Key wrapped around an empty string which would be silly). An empty string wrapped in a `NonEmpty` won't cause an error tho, it'll just set the dns TXT record to an empty string. So it's not an expected situation, should never happen, but still technically valid

library/Fission/User/DID.hs
dholms: This actually takes a key, and returns the DID txt, which is not the same as just displaying the key

when I call it, I call `DID.txtFromKey`. maybe the clearer name would be `Key.toDIDText`?

library/Fission/Internal/Fixture/Key/RSA2048.hs
expede: This file: https://github.com/fission-suite/fission/blob/b9022546ddf4074d35462d328dd72ec99e49473d/library/Fission/Internal/RSA2048/Pair/Types.hs#L12

library/Fission/Internal/Text.hs
expede: > So it's not an expected situation, should never happen, but still technically valid

Ideally we make that state unrepresentable, unless it’s a HUGE pain to do so. Basically use the type system to get rid of edge cases.

One simple way to do this is to use a `newtype`, and a “smart constructor” — a function like `nonEmpty` that does whatever runtime checks it needs. You make sure to not export the actual regular constructor, so the user can’t stick other values in there. Then we know that the value, once constructed, conforms to our exectations.

expede: I can walk you through that in more detail if helpful. It’s a pretty common pattern.

library/Fission/User/DID.hs
expede: How about two functions, and compose them at the call site? (more idiomatic)

```haskell
textDisplay $ toDID pk
```

expede: A few reasons why this is idiomaic:

* Composition is considered par for the course
* Prevent a proliferation of `xToY` functions (grows factorially)
* Still reads well (“tells a story”)

dholms: :+1: 
I actually switched the order of the `DID` type from `DID pk Key` to `DID Key pk` to make it easier to compose, and we don't even need a `toDID` fn :ok_hand:

expede: Amazing 👍

library/Fission/Internal/Text.hs
dholms: yeah you've brought up "smart constructors" a couple times, so a walk through (or point to a good example) of that would probably be helpful

We'd want to do these checks on for CIDs & Keys

expede: > smart constructor

Sure thing! Here’s an example that’s actually possible to enforce the “regular” typed way, but is a huge pain to write that way: even numbers.

```haskell
module Even
  ( Even ()
  , toInt
  , mkEven
  ) where
newtype Even = Even {toInt :: Int}
  deriving newtype (Eq, Ord, Show, Display, Integral)
  
mkEven :: Integral a => a -> Maybe Even
mkEven a =
  if a `mod` 2 == 0
    then Just (Even a)
    else Nothing
```

The interesting part here is the exports. We don’t export the `Even` constructor . Normally you’d write the export line as `Even (..)`, but we want to keep that private to this module. No one can call it from outside.

In its place, we have `mkEven`, which does a runtime check that it is in fact even. If you’re able to get back an `Even` from that function, we now have a typesafe way of saying “this is an even number” without ever having to do the numerical check. `newtype`s get erased at compile time, and thus have zero runtime overhead.

The upshot here is that we only ever have to do the check once, and can have functions make the assumption that they have an even value. We can make this work with any arbitrary check (e.g. “is this email properly formatted?”). In this case, it would exclude the empty string case, and we don’t have to deal with that downstream, handle the errors of trying to pass an empty string to IPFS over the network, and so on. We restrict the possible space of errors by one, and have the compiler statically enforce an assumption about our code for us (it’s a”proof”).

expede: Also perhaps worth mentioning the anatomy of a datatype

```haskell
data TypeName = Constructor {field :: InnerType}
```

The left side (the part still getting exported) is the type (the thing after a `::`). The right side is the constructor (which you call as a value). So, by not exporting the constructor directly, users can depend on `TypeName`, but not make one with `Constructor`. They’re often the same name (“punning”) because they’re in different namespaces.

Sorry if that was review, but wanted to make sure 👍

library/Fission/Internal/RSA2048/Pair/Types.hs
```diff
@@ -1,4 +1,7 @@
-module Fission.Internal.RSA2048.Pair.Types (Pair (..)) where
+module Fission.Internal.RSA2048.Pair.Types 
+  ( Pair (..)
+  , pk1
```
expede, line 3: Why limit the number of keys exported? 🤨

library/Fission/Web/User/UpdateExchangeKeys.hs
```diff
@@ -0,0 +1,33 @@
...
+ 
+import qualified Fission.Web.Error as Web.Error
+
+import qualified Crypto.PubKey.RSA as RSA
+import qualified Fission.User      as User
+
+type API
+  =  Summary "Update Public Exchange Keys"
+  :> Description "Set currently authenticated user's root list of public exchange keys"
+  :> ReqBody '[JSON] [RSA.PublicKey]
```
expede, line 19: Legitimate question:is the user allowed to update their exchange keys to `[]`?

library/Fission/Types.hs
```diff
@@ -396,33 +399,48 @@ instance User.Creator Fission where
...
   updatePassword uID pass now =
     runDB $ User.updatePassword uID pass now
 
   updatePublicKey uID pk now =
-    runDB (User.updatePublicKey uID pk now) >>= \case
-      Left err ->
-        return $ Left err
+    runUpdate (User.updatePublicKey uID pk now) uID "_did" keyTxt 
+    where 
+      keyTxt = textDisplay $ DID Key pk
```
expede, line 438: 💅 Why bother with the `where` here? If you want a shorter line, we can break out more sections to make it more semantic. The where as a line break feels odd.

```haskell
-- Idiomatic line break:
updatePublicKey uID pk now = 
  runUpdate (User.updatePublicKey uID pk now) uID "_did" $
    textDisplay (DID Key pk)

-- Useful/clarifying wheres:
updatePublicKey uID pk now = runUpdate updatePK uID "_did" keyTxt
  where
    keyTxt = textDisplay $ DID Key pk
    updatePK = User.updatePublicKey uID pk now
```

library/Fission/Types.hs
```diff
@@ -396,33 +399,48 @@ instance User.Creator Fission where
...
-              did       = textDisplay (DID pk Key)
-              didSegments = DNS.splitRecord did
-
-            Route53.set Txt url zoneID didSegments 10 <&> \case
-              Left serverErr -> Error.openLeft serverErr
-              Right _        -> Right pk
+  updateExchangeKeys uID keys now =
+    runUpdate (User.updateExchangeKeys uID keys now) uID "_exchange" keysTxt
+    where
+      keysTxt = Text.intercalate "," $ textDisplay . DID Key . Key.RSAPublicKey <$> keys
```
expede, line 443: Yup, this works! I’ve found elsewhere that`x $ y . z <$> a` can be difficult to parse. Consider parens:

```haskell
updatePublicKey uID pk now = runUpdate updatePK uID "_exchange" keyTxt
  where
    updatePK = User.updatePublicKey uID pk now
    keysTxt = Text.intercalate “,” (textDisplay . DID Key . Key.RSAPublicKey <$> keys)
```

library/Fission/Web/User.hs
```diff
@@ -56,6 +58,11 @@ type UpdatePublicKeyRoute
     :> Auth
     :> UpdatePublicKey.API
 
+type UpdateExchangeKeysRoute
+  = "exchange"
```
expede, line 62: Hmm thinking a bit about naming here. `PUT /user/exchange` is a bit generic (is it a message exchange? Key exchange? Transfer of ownership?). Open to suggestions, but perhaps `PUT /user/did/exchange` or `PUT /user/exchange/keys` or similar?

library/Fission/Web/User/UpdateExchangeKeys.hs
```diff
@@ -0,0 +1,33 @@
...
+  :> Patch   '[PlainText, OctetStream, JSON] NoContent
```
expede, line 20: I think this is actually a `PUT` request, no? We’re replacing the entire record, not updating a portion of it. (Unless I misunderstand how it works — totally possible!)

library/Fission/Internal/Fixture/Key/RSA2048.hs
```diff
@@ -0,0 +1,9 @@
+{-# OPTIONS_GHC -fno-warn-incomplete-uni-patterns #-}
+
+module Fission.Internal.Fixture.Key.RSA2048 (pk) where
+
+import qualified Crypto.PubKey.RSA as RSA
+import qualified Fission.Internal.RSA2048.Pair.Types as Pair
+
+pk :: RSA.PublicKey
+pk = Pair.pk1
```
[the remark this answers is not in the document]
dholms: yeah I think I originally did this without realizing it was the exact same info and then never got rid of it. I'll just use the other module :ok_hand:

library/Fission/Internal/RSA2048/Pair/Types.hs
dholms: as in export all of the keys? I'm down for that. None were exported before, and I needed 1 so I exported that. But I can export them all :ok_hand:

library/Fission/Web/User/UpdateExchangeKeys.hs
dholms: Are they able to? Yes
Should they be able to? I'm not totally sure :thinking: Updating to `[]` is essentially the same thing as "deleting" your exchange keys (which we don't have a route for yet) I can imagine a use case for deleting your keys, but it seems very uncommon

library/Fission/Types.hs
dholms: yeah good call. I think I actually like the second one best

library/Fission/Web/User/UpdateExchangeKeys.hs
dholms: ah yup totally right :+1:

library/Fission/Web/User.hs
dholms: yeah makes sense. I went with `exchange/keys`

dholms:
I'm actually wondering if this route would make sense split into 2:
- one for adding an exchange key
- one for removing an exchange key

library/Fission/Web/User/UpdateExchangeKeys.hs
dholms: what I posted lower would address this:
> I'm actually wondering if this route would make sense split into 2:
> - one for adding an exchange key
> - one for removing an exchange key

expede:
> I'm actually wondering if this route would make sense split into 2:
> 
> * one for adding an exchange key
> * one for removing an exchange key

Ah, I see. Two PATCH requests, basically 🤔 It’s a bit of an API smell to see `/add` and `/remove` given `POST` and `DELETE` are RESTful. How do you want to model the API? Here’s one outline:

```REST
PUT /user/exchange_keys/did:key:abcdef
DELETE /user/exchange_keys/did:key:abcdef

# Alternate

PATCH /user/exchange_keys/did:key:abcdef/add
PATCH /user/exchange_keys/did:key:abcdef/remove
```

Given that the way we store the exchange keys doesn’t actually matter, and we have a direct reference to the key itself, there’s a good argument to be made for the `PUT`/`DELETE` version. Thoughts?

Stream of consciousness:

Yeah, either the SDK or the server will need to know how to read in the existing keys, and modify that array. My intuition say to put that on the SDK directly, which is more flexible and works with more [hypothetical] clients — but really at the end of the day it’s making an HTTP request to some service, which will each have their own quirks. Okay, yeah, I see the argument for putting it right on the server 👍

dholms:
sweet split the route into two separate routes:
```
PUT /user/exchange_keys/did:key:abcdef
DELETE /user/exchange_keys/did:key:abcdef
```

I made the routes idempotent, so no errors if you try to add a key that's already there/ delete a key that isn't there already & have them return the full list of keys (after update)

expede:
> I made the routes idempotent, so no errors if you try to add a key that's already there/ delete a key that isn't there already & have them return the full list of keys (after update)

Perfect! That’s the correct semantics for those methods 👍

library/Fission/User/Modifier.hs
```diff
@@ -36,13 +42,65 @@ updatePublicKeyDB ::
...
+      return . openLeft $ NotFound @User
+
+    Just user -> do
+      let keys = userExchangeKeys user
+      if List.elem key keys
+        then 
+          return $ Right keys
+
+        else do
+          let updated = keys ++ [key]
```
expede, line 72: This ordering is actually pretty inefficient, because lists are linked lists and here the innermost one is being changes. Two ways to make this faster:

# 1. Reverse the Ordering

```haskell
let updated = key : keys
```

2. Use a [Vector](https://www.stackage.org/haddock/lts-15.5/rio-0.1.14.1/RIO-Vector.html) [essentially an array]

```haskell
import RIO.Vector
-- ...
let updated = keys ++ singleton key
```

(where `++` is from `RIO.Vector`)

library/Fission/User/Modifier.hs
```diff
@@ -36,13 +42,65 @@ updatePublicKeyDB ::
...
+        then do
+          let updated = List.delete key keys
+          update userID
+            [ UserExchangeKeys =. updated
+            , UserModifiedAt   =. now
+            ]
+          return $ Right updated
+
+        else 
+          return $ Right keys
```
expede, line 102: No need to change, but a difference from say JS is that the following is totally valid and 

```haskell
let
  keys = userExchange user
  updated = key `List.delete` keys
  
if List.elem key keys
  then do
   -- ...
```

We never actually call `updated` until it’s needed, and since `let`s are totally pure they can’t do anything to the `keys` binding, but rather copies/structurally shares in the same way as IPLD.

library/Fission/Web/User/UpdateExchangeKeys.hs
```diff
@@ -0,0 +1,67 @@
...
+type AddAPI
+  =  Summary "Add Public Exchange Key"
+  :> Description "Add a key to the currently authenticated user's root list of public exchange keys"
+  :> Capture "public-did-key" RSA.PublicKey
+  :> Put     '[JSON] [RSA.PublicKey]
+
+type RemoveAPI
+  =  Summary "Remove Public Exchange Key"
+  :> Description "Remove a key from the currently authenticated user's root list of public exchange keys"
+  :> Capture "public-did-key" RSA.PublicKey
```
expede, line 32: “Private DID Key” isn’t a thing, so maybe just `did-key` or `did`? Actually, it may not be a bad idea to capture these as DIDs in case we later add more algorithms — can’t hurt since they’re self-describing. Thoughts?

library/Fission/Web/User/UpdateExchangeKeys.hs
```diff
@@ -0,0 +1,67 @@
...
+  ( MonadTime     m
+  , MonadLogger   m
+  , MonadThrow    m
+  , User.Modifier m
+  )
+  => UserId
+  -> ServerT RemoveAPI m
+removeKey userId key = do
+  now <- currentTime
+  Web.Error.ensureM $ User.removeExchangeKey userId key now
```
expede, line 67: 👍

library/Fission/Types.hs
```diff
@@ -619,3 +585,65 @@ instance MonadEmail Fission where
...
+    folder zonesForDomains (Right accs) url@URL {..} = do
+      case lookup domainName zonesForDomains of
+        Nothing -> do
+          logError $ "Unable to find zone for " <> textDisplay domainName
+          return . Error.openLeft $ NotFound @ZoneID
+
+        Just zoneId ->
+          AWS.clear Txt url zoneId <&> \case
+            Left err -> Error.openLeft err
+            Right _  -> Right (url : accs)
```
expede, line 621: Nice 👌

library/Fission/User/Modifier.hs
dholms: ah cool cool :+1:

library/Fission/User/Modifier.hs
dholms: lazy languages :raised_hands:

library/Fission/Web/User/UpdateExchangeKeys.hs
dholms: yup I like just `did` :+1:

dholms: I changed the name to `did-key-rsa` since that's the only thing we can accept at the moment. We can update that as we can accept more keys :ok_hand:

library/Fission/User/Modifier.hs
```diff
@@ -36,13 +42,67 @@ updatePublicKeyDB ::
...
+  -> Transaction m (Either Errors [RSA.PublicKey])
+addExchangeKeyDB userID key now =
+  Persist.get userID >>= \case
+    Nothing -> 
+      return . openLeft $ NotFound @User
+
+    Just user -> do
+      let 
+        keys = userExchangeKeys user
+        updated = [key] ++ keys
```
expede, line 68: Even faster: `key : keys`

expede: (it’s a linked list)

dholms: ah nice!

Issue #344, opened 2020-06-23
Email verification


bmann:
We need to have verified emails on file so we can help users reset their accounts, receive account / system related email, and optionally subscribe to product updates.

This is needed for both web signups and CLI signups. For web signups, the front end is handled by Auth. Auth includes checkbox to subscribe to Product Updates as part of flow.

Select a transactional email provider. We have accounts with Mailgun and Sendinblue. Slight preference for SendinBlue which we use for mailing lists. [SIB Docs](https://developers.sendinblue.com/docs)

On account signup, status of account / email should be set to unverified. Send a verification email where the user clicks on a link, received by the server, status set to verified (can we use Auth for this for CLI?)

PR #349, opened 2020-07-02
Email verification

dholms:
## Problem
We need to have verified emails on file so we can help users reset their accounts, receive account / system related email, and optionally subscribe to product updates.

## Solution
- On account creation, create a challenge (32 character string)
- Send a verification email through SendInBlue with a link to `user/email/verify/${challenge}`
- when visited, change a user's status in the DB to **verified**

Closes https://github.com/fission-suite/fission/issues/344

### DIFF
--- a/env.yaml.example
+++ b/env.yaml.example
@@ -34,3 +34,7 @@ fission_file_system:
 web_app:
   base_app_domain_name: fission.app
   app_placeholder_cid: QmRVvvMeMEPi1zerpXYH9df3ATdzuB63R1wf3Mz5NS5HQN # Fission logo on a blank page
+
+send_in_blue:
+  api_key: xkeysib-KEY
+  base_url: https://api.sendinblue.com 
--- a/library/Fission/Challenge.hs
+++ b/library/Fission/Challenge.hs
@@ -1,0 +1,23 @@
+module Fission.Challenge
+  ( module Fission.Challenge.Types
+  , module Fission.Challenge.Creator.Class
+  , module Fission.Challenge.Verifier.Class
+  , verificationLink
+  ) where
+
+import Fission.Challenge.Types
+import Fission.Challenge.Creator.Class
+import Fission.Challenge.Verifier.Class
+
+import Fission.Prelude
+import Servant         hiding (route)
+
+import qualified Fission.Web.Routes   as Web 
+import qualified Fission.Internal.API as API
+import qualified Fission.Web.User     as User
+
+verificationLink :: Challenge -> Text
+verificationLink challenge = 
+  toUrlPiece $ API.mkLink route challenge
+  where 
+    route = Proxy @(Web.UserPrefix :> User.VerifyEmailRoute)
--- a/library/Fission/Challenge/Creator/Class.hs
+++ b/library/Fission/Challenge/Creator/Class.hs
@@ -1,0 +1,29 @@
+module Fission.Challenge.Creator.Class (Creator (..)) where
+
+import           Fission.Prelude
+import           Fission.Error.Types
+import qualified Fission.Random as Random
+
+import           Fission.Models as Models
+import           Database.Esqueleto
+
+import           Fission.Challenge.Types
+
+
+class Monad m => Creator m where
+  create :: UserId -> m (Either (AlreadyExists UserChallenge) Challenge)
+
+instance MonadIO m => Creator (Transaction m) where
+  create userId = do
+    challenge <- mkChallenge
+    UserChallenge
+      { userChallengeUserId = userId
+      , userChallengeHash   = challenge
+      }
+      |> insertUnique
+      |> bind  \case
+        Nothing -> return $ Left AlreadyExists
+        Just _  -> return $ Right challenge
+
+mkChallenge :: MonadIO m => m Challenge
+mkChallenge = Challenge <$> Random.alphaNum 32
--- a/library/Fission/Challenge/Types.hs
+++ b/library/Fission/Challenge/Types.hs
@@ -1,0 +1,63 @@
+module Fission.Challenge.Types (Challenge(..)) where
+  
+import           Database.Persist.Class
+import           Database.Persist.Types
+import           Database.Persist.Sql
+
+import qualified RIO.ByteString.Lazy as Lazy
+import qualified RIO.Text            as Text
+
+import qualified Network.IPFS.Internal.UTF8 as UTF8
+
+import           Data.Swagger
+import           Servant
+
+import           Fission.Prelude
+
+
+newtype Challenge = Challenge { unChallenge :: Text }
+  deriving newtype  ( Eq
+                    , Show
+                    , IsString
+                    , ToHttpApiData
+                    )
+
+instance Arbitrary Challenge where
+  arbitrary = Challenge <$> arbitrary
+
+instance ToJSON Challenge where
+  toJSON (Challenge challenge') = String challenge'
+
+instance FromJSON Challenge where
+  parseJSON = withText "Challenge" (pure . Challenge)
+
+instance FromHttpApiData Challenge where
+  parseUrlPiece = Right . Challenge
+
+instance PersistField Challenge where
+  toPersistValue (Challenge challenge') = PersistText challenge'
+  fromPersistValue = \case
+    PersistText challenge' -> Right (Challenge challenge')
+    other                  -> Left ("Invalid Persistent Challenge: " <> Text.pack (show other))
+
+instance PersistFieldSql Challenge where
+  sqlType _pxy = SqlString
+
+instance MimeRender PlainText Challenge where
+  mimeRender _ = UTF8.textToLazyBS . unChallenge
+
+instance MimeUnrender PlainText Challenge where
+  mimeUnrender _ = bimap show Challenge . decodeUtf8' . Lazy.toStrict
+
+instance ToParamSchema Challenge where
+  toParamSchema _ = mempty |> type_ ?~ SwaggerString
+
+instance ToSchema Challenge where
+  declareNamedSchema _ =
+    mempty
+      |> type_       ?~ SwaggerString
+      |> description ?~ "A hash representing a user challenge. Sent in a verification email"
+      |> example     ?~ toJSON (Challenge "KyGig3J8lx3K07gZMK1lJV9M6Pr8w2RB")
+      |> NamedSchema (Just "Challenge")
+      |> pure
+
--- a/library/Fission/Challenge/Verifier/Class.hs
+++ b/library/Fission/Challenge/Verifier/Class.hs
@@ -1,0 +1,32 @@
+module Fission.Challenge.Verifier.Class (Verifier (..)) where
+
+import           Database.Persist
+
+import           Fission.Prelude
+import           Fission.Error as Error
+import           Fission.Models
+import           Fission.Challenge.Types
+
+
+class Monad m => Verifier m where
+  verify :: Challenge -> m (Either (NotFound UserChallenge) ())
+
+instance MonadIO m => Verifier (Transaction m) where
+  verify challenge = do
+    res <- selectFirst [ UserChallengeHash ==. challenge ] [] 
+    case res of
+      Nothing -> 
+        return $ Left NotFound
+
+      Just (Entity challengeId UserChallenge { userChallengeHash, userChallengeUserId }) -> 
+        if challenge == userChallengeHash
+          then do
+            update userChallengeUserId
+              [ UserVerified  =. True ]
+
+            delete challengeId
+
+            return ok
+
+          else
+            return $ Left NotFound
--- a/library/Fission/Config/Types.hs
+++ b/library/Fission/Config/Types.hs
@@ -20,12 +20,15 @@ import qualified Fission.AWS.Types as AWS
 import           Fission.URL.Types as URL
 import           Fission.Web.Types
 
+import qualified Fission.Email.SendInBlue.Types as SIB
+
 -- | The top level 'Fission' application 'RIO' configuration
 data Config = Config
   { processCtx     :: !ProcessContext
   , logFunc        :: !LogFunc
   --
   , httpManager    :: !HTTP.Manager
+  , tlsManager     :: !HTTP.Manager
   , dbPool         :: !(Pool SqlBackend)
   --
   , ipfsPath       :: !IPFS.BinPath
@@ -48,6 +51,10 @@ data Config = Config
   , userZoneID     :: !AWS.ZoneID
   , defaultDataCID :: !CID
   --
+  , sibApiKey      :: !SIB.ApiKey
+  , sibUrl         :: !Host
+  , sibTemplateId  :: !SIB.TemplateId
+  --
   , host           :: !Host
   , fissionDID     :: !DID
   , serverZoneID   :: !AWS.ZoneID
@@ -61,6 +68,7 @@ instance Show Config where
     , "  logFunc           = **SOME LOG FUNCTION**"
     --
     , "  httpManager       = **SOME HTTP MANAGER**"
+    , "  tlsManager        = **SOME HTTP/TLS MANAGER**"
     , "  dbPool            = " <> show dbPool
     --
     , "  ipfsPath          = " <> show ipfsPath
@@ -82,6 +90,10 @@ instance Show Config where
     , "  userRootDomain    = " <> show userRootDomain
     , "  userZoneID        = " <> show userZoneID
     , "  defaultDataCID    = " <> show defaultDataCID
+    --
+    , "  sibApiKey         = " <> show sibApiKey
+    , "  sibUrl            = " <> show sibUrl
+    , "  sibTemplateId     = " <> show sibTemplateId
     --
     , "  host              = " <> show host
     , "  fissionDID        = " <> show fissionDID
--- a/library/Fission/Email/Class.hs
+++ b/library/Fission/Email/Class.hs
@@ -1,0 +1,11 @@
+module Fission.Email.Class (MonadEmail (..)) where
+
+import           Fission.Prelude
+import           Servant.Client
+
+import           Fission.Email.Recipient.Types
+import           Fission.Email.SendInBlue.Types as SIB
+import           Fission.Challenge.Types
+
+class Monad m => MonadEmail m where
+  sendVerificationEmail :: Recipient -> Challenge -> m (Either ClientError SIB.Response)
--- a/library/Fission/Email/Recipient/Types.hs
+++ b/library/Fission/Email/Recipient/Types.hs
@@ -1,0 +1,16 @@
+module Fission.Email.Recipient.Types (Recipient(..)) where
+
+import Fission.Prelude
+import Fission.User.Email.Types
+import Fission.User.Username.Types
+
+data Recipient = Recipient 
+  { email :: !Email
+  , name  :: !Username
+  }
+
+instance ToJSON Recipient where
+  toJSON Recipient { email, name } =
+    Object [ ("email", toJSON email)
+           , ("name",  toJSON name)
+           ]
--- a/library/Fission/Email/SendInBlue/Client.hs
+++ b/library/Fission/Email/SendInBlue/Client.hs
@@ -1,0 +1,22 @@
+module Fission.Email.SendInBlue.Client 
+  ( SendEmailAPI
+  , sendEmail
+  ) where
+
+import Fission.Prelude
+import Fission.Email.SendInBlue.Types as Email
+
+import Servant
+import Servant.Client
+
+
+type SendEmailAPI
+  = "v3"
+  :> "smtp"
+  :> "email"
+  :> Header "api-key" Text
+  :> ReqBody '[JSON] Email.Request
+  :> Post    '[JSON] Email.Response
+
+sendEmail :: ApiKey -> Email.Request -> ClientM Email.Response
+sendEmail (ApiKey key) = (client $ Proxy @SendEmailAPI) (Just key)
--- a/library/Fission/Email/SendInBlue/Request/Types.hs
+++ b/library/Fission/Email/SendInBlue/Request/Types.hs
@@ -1,0 +1,24 @@
+module Fission.Email.SendInBlue.Request.Types (Request(..)) where
+
+import           Fission.Prelude hiding (to)
+
+import           Fission.Email.Recipient.Types
+import           Fission.Email.SendInBlue.TemplateOptions.Types
+import           Fission.Email.SendInBlue.TemplateId.Types
+
+import qualified RIO.NonEmpty as NonEmpty
+import qualified Data.Vector  as Vector
+
+
+data Request = Request
+  { templateId :: !TemplateId
+  , to         :: !(NonEmpty Recipient)
+  , params     :: !TemplateOptions
+  } 
+
+instance ToJSON Request where
+  toJSON Request { templateId, to, params } =
+    Object [ ("templateId", Number $ fromIntegral (unTemplateId templateId))
+           , ("to",         Array . Vector.fromList $ NonEmpty.toList (toJSON <$> to))
+           , ("params",     toJSON params)
+           ]
--- a/library/Fission/Email/SendInBlue/Response/Types.hs
+++ b/library/Fission/Email/SendInBlue/Response/Types.hs
@@ -1,0 +1,18 @@
+module Fission.Email.SendInBlue.Response.Types (Response(..)) where
+
+import Fission.Prelude
+
+
+newtype Response = Response { messageId :: Text }
+  deriving newtype  ( Eq
+                    , Show
+                    , IsString
+                    , Display
+                    )
+
+instance FromJSON Response where
+  parseJSON = withObject "Response" \obj -> do
+    messageId <- obj .:  "messageId"
+
+    return Response {..}
+
--- a/library/Fission/Email/SendInBlue/TemplateId/Types.hs
+++ b/library/Fission/Email/SendInBlue/TemplateId/Types.hs
@@ -1,0 +1,10 @@
+module Fission.Email.SendInBlue.TemplateId.Types (TemplateId(..)) where
+
+import Fission.Prelude
+
+
+newtype TemplateId = TemplateId { unTemplateId :: Int64 }
+  deriving newtype  ( Eq
+                    , Show
+                    , FromJSON
+                    )
--- a/library/Fission/Error/AlreadyExists/Types.hs
+++ b/library/Fission/Error/AlreadyExists/Types.hs
@@ -21,6 +21,9 @@ instance Display (AlreadyExists entity) => ToServerError (AlreadyExists entity) 
 instance Display (AlreadyExists User) where
   display _ = "User already exists"
 
+instance Display (AlreadyExists UserChallenge) where
+  display _ = "Challenge already exists"
+
 instance Display (AlreadyExists LoosePin) where
   display _ = "Loose pin already exists"
 
--- a/library/Fission/Models.hs
+++ b/library/Fission/Models.hs
@@ -28,6 +28,7 @@ import           Fission.URL
 import           Fission.User.Email.Types
 import           Fission.User.Role.Types
 import           Fission.User.Username.Types
+import           Fission.Challenge.Types
 
 import qualified Fission.AWS.Zone.Types               as AWS
 
@@ -60,6 +61,7 @@ User
 
   role          Role
   active        Bool
+  verified      Bool
 
   dataRoot      CID
 
@@ -72,6 +74,17 @@ User
   UniqueUsername  username
   UniquePublicKey publicKey !force
   -- UniqueEmail     email     !force -- FIXME
+
+  deriving Show Eq
+
+--------------------------------------------------------------------------------
+
+UserChallenge
+  userId    UserId
+  hash      Challenge
+
+  UniqueUserId    userId
+  UniqueChallenge hash
 
   deriving Show Eq
 
--- a/library/Fission/Types.hs
+++ b/library/Fission/Types.hs
@@ -75,6 +75,10 @@ import           Fission.Authorization.ServerDID.Class
 import           Fission.App.Content                   as App.Content
 import           Fission.App.Domain                    as App.Domain
 
+import           Fission.Challenge as Challenge
+import qualified Fission.Email     as Email
+import           Fission.Email.Class
+
 import qualified Fission.Domain as Domain
 
 -- | The top-level app type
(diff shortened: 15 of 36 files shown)


library/Fission/Types.hs
```diff
@@ -563,3 +568,29 @@ instance Domain.Creator Fission where
...
+  create email = 
+    runDB $ Challenge.create email
+
+instance Challenge.Verifier Fission where
+  verify challenge = 
+    runDB $ Challenge.verify challenge
+
+instance Mail.MonadMail Fission where
+  sendVerifyEmail recipient (Challenge hash) = do
+    httpManager <- liftIO $ HTTP.newManager HTTP.tlsManagerSettings
```
dholms, line 581: atleast testing locally, the `httpManager` is not TLS enabled. And looking at `server.Main.hs`, it looks like it's never TLS enabled

Do we need a second manager in `Config`? or is it fine to create one ephemerally here?

library/Fission/Challenge/Types.hs
```diff
@@ -0,0 +1,59 @@
...
+import           Servant
+
+import           Fission.Prelude
+
+
+newtype Challenge = Challenge { unChallenge :: Text }
+  deriving          ( Generic )
+  deriving anyclass ( ToSchema
+                    , ToParamSchema
+                    )
```
expede, line 22: There's probably a few of these left in the code, but if possible let's do explicit instances for `ToSchema` and `ToParamSchema`. You can likely also drop the `Generic` once that's done.

library/Fission/Challenge/Types.hs
```diff
@@ -0,0 +1,59 @@
...
+
+
+instance Arbitrary Challenge where
+  arbitrary = Challenge <$> arbitrary
+
+instance ToJSON Challenge where
+  toJSON (Challenge challenge') = String challenge'
+
+instance FromJSON Challenge where
+  parseJSON = withText "Challenge" \txt -> return (Challenge txt)
```
expede, line 37: 👍 

More a pro-tip (no change required), but you can also write this as:

```haskell
withText "Challenge" Challenge
```

library/Fission/Challenge/Verifier/Class.hs
```diff
@@ -0,0 +1,32 @@
...
+import           Database.Persist
+
+import           Fission.Prelude
+import           Fission.Error as Error
+import           Fission.Models
+import           Fission.Challenge.Types
+
+
+class Monad m => Verifier m where
+  verify :: Challenge -> m (Either (NotFound UserChallenge) ())
```
expede, line 12: 👍

library/Fission/Challenge/Verifier/Class.hs
```diff
@@ -0,0 +1,32 @@
...
+
+instance MonadIO m => Verifier (Transaction m) where
+  verify challenge = do
+    res <- selectFirst [ UserChallengeHash ==. challenge ] [] 
+    case res of
+      Nothing -> 
+        return $ Left NotFound
+
+      Just (Entity challengeId UserChallenge { userChallengeHash, userChallengeUserId }) -> 
+        if challenge /= userChallengeHash
```
expede, line 22: Compare with the positive case whenever possible. The logic here for the else case is "if not not X".

library/Fission/Config/Types.hs
```diff
@@ -83,6 +90,9 @@ instance Show Config where
     , "  userZoneID        = " <> show userZoneID
     , "  defaultDataCID    = " <> show defaultDataCID
     --
+    , "  sibApiKey         = " <> show sibApiKey
+    , "  sibUrl            = " <> show sibUrl
```
expede, line 94: No `sibTemplateId`?

library/Fission/Error/AlreadyExists/Types.hs
```diff
@@ -21,6 +21,9 @@ instance Display (AlreadyExists entity) => ToServerError (AlreadyExists entity)
 instance Display (AlreadyExists User) where
   display _ = "User already exists"
 
+instance Display (AlreadyExists UserChallenge) where
+  display _ = "Challenge already exists for email"
```
expede, line 25: Is it scoped per email address? My understanding was that they're assumed globally unique so we can look them up by the challenge text only, no?

library/Fission/Mail/Class.hs
```diff
@@ -0,0 +1,11 @@
...
+
+import           Fission.Prelude
+import           Servant.Client
+
+import           Fission.Mail.Recipient.Types
+import           Fission.Mail.SendInBlue.Types 
+import           Fission.Challenge.Types
+
+class Monad m => MonadMail m where
+  sendVerifyEmail :: Recipient -> Challenge -> m (Either ClientError SendEmailRes)
```
expede, line 11: `SendEmailRes`

Actually type out `Response`. You can break that out into a submodule as well to compress the text contextually:

```haskell
import qualified Fission.Mail.SendInBlue.Types as Email

foo :: Email.Response
```

I'd also recommend adding an `E` to `Mail`. Unambiguous, and it's one character.

library/Fission/Mail/Recipient/Types.hs
```diff
@@ -0,0 +1,17 @@
...
+data Recipient = Recipient 
+  { email :: !Email
+  , name :: Username
+  }
+
+instance ToJSON Recipient where
+  toJSON Recipient { email = email', name = name' } =
+    Object [ ("email", toJSON email')
+           , ("name", toJSON name')
+           ]
```
expede, line 16:
```haskell
  toJSON Recipient { .. } =
    Object [ ("email", toJSON email)
           , ("name",  toJSON name)
           ]
```

library/Fission/Mail/Recipient/Types.hs
```diff
@@ -0,0 +1,17 @@
+module Fission.Mail.Recipient.Types (Recipient(..)) where
+
+import Fission.Prelude
+import Fission.User.Email.Types
+import Fission.User.Username.Types
+
+data Recipient = Recipient 
+  { email :: !Email
+  , name :: Username
+  }
```
expede, line 10: Missing a `!` on `Username`

library/Fission/Mail/SendInBlue/ApiKey/Types.hs
```diff
@@ -0,0 +1,14 @@
+module Fission.Mail.SendInBlue.ApiKey.Types (ApiKey(..)) where
+
+import Fission.Prelude
+
+newtype ApiKey = ApiKey { unApiKey :: Text }
+  deriving          ( Generic )
```
expede, line 6: Why do we need `Generic` here?

library/Fission/Mail/SendInBlue/Client.hs
```diff
@@ -0,0 +1,24 @@
...
+import Servant.Client
+
+
+type SendEmailAPI
+  = "v3"
+  :> "smtp"
+  :> "email"
+  :> Header "api-key" Text
+  :> ReqBody '[JSON] SendEmailReq
+  :> Post    '[JSON] SendEmailRes
```
expede, line 19: `Email.Request` & `Email.Response`

library/Fission/Mail/SendInBlue/TemplateId/Types.hs
```diff
@@ -0,0 +1,10 @@
+module Fission.Mail.SendInBlue.TemplateId.Types (TemplateId(..)) where
+
+import Fission.Prelude
+
+newtype TemplateId = TemplateId { unTemplateId :: Int64 }
+  deriving          ( Generic )
```
expede, line 6: 🔪 `Generic`

library/Fission/Mail/SendInBlue/Types.hs
```diff
@@ -0,0 +1,53 @@
...
+import           Fission.User.Username.Types
+import           Servant.Client
+
+import qualified Data.Vector as Vector
+
+
+data SendEmailReq = SendEmailReq
+  { templateId :: !TemplateId
+  , to :: ![Recipient]
+  , params :: !CustomParams
```
expede, line 23: What is `CustomParams`? Template options?

library/Fission/Mail/SendInBlue/Types.hs
```diff
@@ -0,0 +1,53 @@
...
+data CustomParams = CustomParams
+  { verifyLink :: !BaseUrl 
+  , username   :: !Username 
+  } 
+
+instance ToJSON CustomParams where
+  toJSON CustomParams { verifyLink = verifyLink', username = username' } =
+    Object [ ("VERIFY_LINK", toJSON verifyLink' )
+           , ("USERNAME", toJSON username' )
+           ]
```
expede, line 42: Same as in an earlier module:

```haskell
toJSON CustomParams { .. } =
    Object [ ("VERIFY_LINK", toJSON verifyLink )
           , ("USERNAME", toJSON username )
           ]
```

library/Fission/Mail/SendInBlue/Types.hs
```diff
@@ -0,0 +1,53 @@
...
+
+
+data SendEmailRes = SendEmailRes
+  { messageId :: Text }
```
expede, line 46: 1. Since this is a single parameter, this should be a `newtype`
2. `deriving newtype (Eq, Show, Display)`

library/Fission/Models.hs
```diff
@@ -75,6 +77,14 @@ User
 
   deriving Show Eq
 
+--------------------------------------------------------------------------------
+
+UserChallenge
+  userId    UserId
+  hash      Challenge
+
+  deriving Show Eq
```
expede, line 86: Are these unique? i.e. can we have multiple active challenges per user?

library/Fission/Types.hs
expede: Our server doesn't use TLS directly anymore because we're using the AWS proxy for certs

expede: > Do we need a second manager in Config? or is it fine to create one ephemerally here?

If we're going to be making TLS requests, then yes, we should have one kicking around at all times. It takes up very little room in memory, and can get reused for lots of things. Plus it would be good to avoid the repeated setup/teardown.

library/Fission/User/Creator/Class.hs
```diff
@@ -88,6 +88,7 @@ instance
       , userEmail         = Just email
       , userRole          = Regular
       , userActive        = True
+      , userVerified      = False
```
expede, line 91: Hmmm this does make me wonder about having the two tables. Two ideas:

1. Have nullable text field for the current challenge. If it's null, the user has been verified. Write a helper `isVerified`.

2. Make users and emails a one-to-many (user has many email addresses). Keep the challenge on the `EmailAddresses` table.

Thoughts?

library/Fission/Web/User.hs
```diff
@@ -51,6 +56,10 @@ type VerifyRoute
     :> Auth
     :> Verify.API
 
+type VerifyEmailRoute
+  = "verify_email"
+    :> VerifyEmail.API
```
expede, line 61: Why not on `/user/verify`?

library/Fission/Web/User/Create.hs
```diff
@@ -31,26 +33,40 @@ withDID ::
...
-  Web.Err.ensureM $ User.create username publicKey email now
-  return NoContent
+  userId <- Web.Err.ensureM $ User.create username publicKey email now
+  challenge <- Web.Err.ensureM $ Challenge.create userId
+  res <- sendVerifyEmail (Recipient email username) challenge
+  case res of
+    Left err -> do
+      logError $ displayShow err
+      Web.Err.throw err500 { errBody = "Could not send verification email" }
+    Right _ -> return NoContent
```
expede, line 51: Remember that you can also write this as:

```haskell
sendVerifyEmail (Recipient email username) challenge >>= \case
  Left  err -> ...
  Right val -> ...
```

library/Fission/Mail/SendInBlue/TemplateId/Types.hs
dholms: oh really? when do we want `Generic` I thought we derived it anytime the type was a simple wrapper of some primitive

library/Fission/Types.hs
dholms: > Our server doesn't use TLS directly anymore because we're using the AWS proxy for certs
right right :ok_hand: 

> we should have one kicking around at all times in Config
ok cool. I'll move it to Config :+1:

library/Fission/Mail/SendInBlue/TemplateId/Types.hs
expede: > we derived it anytime the type was a simple wrapper of some primitive

Nope! It's very possible (and common!) to have a `Generic`-free codebase. Generic is only used by the compiler to help derive certain types of instances automatically (e.g. `ToParamSchema`). Since we're opting to make these explicit, we don't need the compiler to do the work of creating a bunch of extra compile-time annotations.

library/Fission/Challenge/Verifier/Class.hs
dholms: :+1:

library/Fission/Challenge/Types.hs
dholms: oh i like it!

library/Fission/Config/Types.hs
dholms: oops :stuck_out_tongue:

library/Fission/Error/AlreadyExists/Types.hs
dholms: oh yup I'll change the wording :+1: I was originally handling this differently and never changed the wording

yup they're assumed globally unique. we also shouldn't ever hit this error, it'd be about the same probability as a SHA1 collision

library/Fission/Mail/Class.hs
dholms: sounds good :+1:

library/Fission/Mail/SendInBlue/ApiKey/Types.hs
dholms: :man_shrugging: 
i'll be honest I'm not really sure what `Generic` does. I've just been throwing it on `newtype`s that are easily derivable from primitive-ish values :sweat_smile:

library/Fission/Mail/SendInBlue/Types.hs
dholms: yup exactly

I originally had it `Params` but that was so vague that I changed it to `CustomParams`, but `TemplateOptions` is a more descriptive type name

library/Fission/Web/User.hs
dholms: `/user/verify` is already taken (although that route is deprecated)
`/user/email/verify` would be fine tho :+1:

library/Fission/Web/User/Create.hs
dholms: yup :+1: 
I like this better, but I thought we were trying to be consistent within functions of either using `do` notation or `bind` notation

Is there a good rule of thing on when we should mix them? I feel like `case` statements lend themselves really nicely to binds

expede: I mean, they're totally equivalent. The main difference is that there's a branch point; we're not really using the variable anywhere else and it's mainly for control flow. Really either is fine, but `>>= \case` saves you a line and I find it a bit more expressive, _unless_ you're doing something like normalizing the response and continuing

```haskell
response <- request

normalized <- case response of
  Left  err -> defaultValue
  Right val -> foo val

bar normalized
```

expede: Also, a LOT of these `>>= \case`s will probably disappear when we have `rescue` integrated

expede: > I feel like case statements lend themselves really nicely to binds

Yes I'd say that's the exception to the rule, and is increasing popular with the community broadly

library/Fission/Models.hs
dholms: yes they are unique. good call :+1:

library/Fission/User/Creator/Class.hs
dholms: > Hmmm this does make me wonder about having the two tables
why does this make you wonder about two tables?

I'm down for either for those ideas though. Are we planning to make email addresses one-to-many at some point?

library/Fission/Web/User/Create.hs
dholms: okay great! :ok_hand:

library/Fission/Types.hs
```diff
@@ -563,3 +567,30 @@ instance Domain.Creator Fission where
...
+  sendVerifyEmail recipient@(Email.Recipient { name }) (Challenge hash) = do
+    httpManager      <- asks tlsManager
+    Host baseHostUrl <- asks host
+    Host sibUrl      <- asks sibUrl
+    apiKey           <- asks sibApiKey
+    templateId       <- asks sibTemplateId
+
+    let
+      env = mkClientEnv httpManager sibUrl
+      path = "user/verify_email/" <> Text.unpack hash
```
dholms, line 588: is there a better option instead of hard coding this? :thinking: I'm not seeing any methods in Servant to go from `API -> Text`

expede: I haven't tried to do this, but looking at Stackage: https://www.stackage.org/haddock/lts-16.3/servant-0.16.2/Servant-API.html#v:safeLink

expede: You can 100% use Servant for templating, so they for sure have some way of generating links. I'd assume relative links are an option.

expede: Also potentially helpful: https://docs.servant.dev/en/stable/cookbook/generic/Generic.html?highlight=link#links

Though we're not using `servant-generic` at the moment. Not opposed to it, though. It's known to be a bit on the slow side, but in this case we'll _probably_ generate it once it'll stay memoized for future use.

expede: Here's exactly the aformentioned relative link function:

https://www.stackage.org/haddock/lts-16.3/servant-lucid-0.9/Lucid-Servant.html#v:safeRelHref_

Lucid is one of the templating libraries. It's very fast, typesafe, and "good". If/when we want to serve HTML from the API, we'd probably use either Lucid or Blaze.

We don't lose anything by including Lucid as a dependency other than adding a bit of time downloading the package the first time. If this is the shortest path, do that 👍

expede: # ASIDE

While I'm looking at that those two:

https://mmhaskell.com/blog/2020/3/16/lucid-another-html-option

I've used Blaze before (years ago) and liked it. Lucid is newer. I vaguely remember comparing them around the New Year, and feeling like Lucid was probbaly the better choice, but they're _very_ similar.

library/Fission/Challenge/Creator/Class.hs
```diff
@@ -0,0 +1,29 @@
...
+  create userId = do
+    challenge <- mkChallenge
+    UserChallenge
+      { userChallengeUserId = userId
+      , userChallengeHash   = challenge
+      }
+      |> insertUnique
+      |> bind  \case
+        Nothing -> return $ Left AlreadyExists
+        Just _  -> return $ Right challenge
```
expede, line 26: Great 👍

library/Fission/Challenge/Creator/Class.hs
```diff
@@ -0,0 +1,29 @@
...
+
+mkChallenge :: MonadIO m => m Challenge
+mkChallenge = Challenge <$> Random.alphaNum 32
```
expede, line 29: Noice

library/Fission/Challenge/Types.hs
```diff
@@ -0,0 +1,67 @@
...
+
+instance ToSchema Challenge where
+  declareNamedSchema _ =
+    mempty
+      |> type_       ?~ SwaggerString
+      |> description ?~ "A hash representing a user challenge. Sent in a verification email"
+      |> example     ?~ toJSON (Challenge "KyGig3J8lx3K07gZMK1lJV9M6Pr8w2RB")
+      |> NamedSchema (Just "Challenge")
+      |> pure
+
```
expede, line 67: This entire file:

image
)

library/Fission/Challenge/Verifier/Class.hs
```diff
@@ -0,0 +1,32 @@
...
+          then do
+            update userChallengeUserId
+              [ UserVerified  =. True ]
+
+            delete challengeId
+
+            return ok
+
+          else
+            return $ Left NotFound
```
expede, line 32: Awesome possum

library/Fission/Email/Class.hs
```diff
@@ -0,0 +1,11 @@
...
+
+import           Fission.Prelude
+import           Servant.Client
+
+import           Fission.Email.Recipient.Types
+import           Fission.Email.SendInBlue.Types as SIB
+import           Fission.Challenge.Types
+
+class Monad m => MonadEmail m where
+  sendVerifyEmail :: Recipient -> Challenge -> m (Either ClientError SIB.Response)
```
expede, line 11: Nitpick, feel free to ignore: `sendVerificationEmail` or `sendEmailVerification`

library/Fission/Email/Recipient/Types.hs
```diff
@@ -0,0 +1,16 @@
...
+data Recipient = Recipient 
+  { email :: !Email
+  , name  :: !Username
+  }
+
+instance ToJSON Recipient where
+  toJSON Recipient { email, name } =
+    Object [ ("email", toJSON email)
+           , ("name",  toJSON name)
+           ]
```
expede, line 16: 👍

library/Fission/Email/SendInBlue/Client.hs
```diff
@@ -0,0 +1,22 @@
...
+import Servant.Client
+
+
+type SendEmailAPI
+  = "v3"
+  :> "smtp"
+  :> "email"
+  :> Header "api-key" Text
+  :> ReqBody '[JSON] Email.Request
+  :> Post    '[JSON] Email.Response
```
expede, line 19: Awesome. Super legible 👍

library/Fission/Email/SendInBlue/Request/Types.hs
```diff
@@ -0,0 +1,23 @@
...
+import           Fission.Email.Recipient.Types
+import           Fission.Email.SendInBlue.TemplateOptions.Types
+import           Fission.Email.SendInBlue.TemplateId.Types
+
+import qualified Data.Vector as Vector
+
+
+data Request = Request
+  { templateId :: !TemplateId
+  , to         :: ![Recipient]
```
expede, line 14: Can it be empty?

library/Fission/Email/SendInBlue/Request/Types.hs
```diff
@@ -0,0 +1,23 @@
...
+  , params     :: !TemplateOptions
+  } 
+
+instance ToJSON Request where
+  toJSON Request { templateId = templateId', to = to', params = params' } =
```
expede, line 19: `toJSON Request { templateId, to, params } = ...`

(No need for the `=` and `’`)

library/Fission/Email/SendInBlue/Response/Types.hs
```diff
@@ -0,0 +1,18 @@
...
+                    , Show
+                    , IsString
+                    , Display
+                    )
+
+instance FromJSON Response where
+  parseJSON = withObject "Response" \obj -> do
+    messageId <- obj .:  "messageId"
+
+    return Response {..}
```
expede, line 17: Totally great to leave as is. Just showing an alternate:

```haskell
parseJSON = withObject “Response” (Response <$> obj .: “messageId”)
```

library/Fission/Email/SendInBlue/TemplateId/Types.hs
```diff
@@ -0,0 +1,11 @@
+module Fission.Email.SendInBlue.TemplateId.Types (TemplateId(..)) where
+
+import Fission.Prelude
+
+
+newtype TemplateId = TemplateId { unTemplateId :: Int64 }
+  deriving          ( Generic )
```
expede, line 7: Generic?

library/Fission/Types.hs
```diff
@@ -563,3 +567,30 @@ instance Domain.Creator Fission where
...
+      env = mkClientEnv httpManager sibUrl
+      path = "user/email/verify/" <> Text.unpack hash
+      verifyUrl = baseHostUrl { baseUrlPath = path }
+      emailData = Email.Request
+        { templateId = templateId
+        , to = [recipient]
+        , params = Email.TemplateOptions verifyUrl name
+        }
+
+    liftIO $ runClientM (Email.sendEmail apiKey emailData) env
```
expede, line 596: 🎉

library/Fission/User/Creator/Class.hs
expede: Oh I see. Yeah, almost certainly many-to-one down the road, yeah. I asked Bori about this last week, and he said to keep it to whatever is going to be less error prone for a single email for now, since we don’t know the priority of multi-email. Could be a while.

library/Fission/Web.hs
```diff
@@ -114,7 +123,10 @@ bizServer ::
   , MonadDNSLink              m
   , MonadLogger               m
   , MonadTime                 m
+  , MonadEmail                m
   , User.CRUD                 m
+  , Challenge.Creator         m
+  , Challenge.Verifier        m
```
expede, line 129: Nice

library/Fission/Web/User/Create.hs
```diff
@@ -28,29 +30,46 @@ type PasswordAPI
...
   now <- currentTime
-  Web.Err.ensureM $ User.createWithPassword username pass email now
-  return ()
+  userId <- Web.Err.ensureM $ User.createWithPassword username pass email now
+  challenge <- Web.Err.ensureM $ Challenge.create userId
+
+  sendVerifyEmail (Recipient email username) challenge >>= \case
+    Left _ -> 
+      Web.Err.throw err500 { errBody = "Could not send verification email" }
+    Right _ ->
```
expede, line 74: Oh crap, right. Password registration is for Heroku only. We need to figure this out how to get a user email from them. Otherwise this will break the Jeroku add on!

library/Fission/Web/User/VerifyEmail.hs
```diff
@@ -0,0 +1,31 @@
...
+  , Challenge.Verifier m
+  )
+ => ServerT API m
+server challenge =
+  Challenge.verify challenge >>= \case
+    Left _ ->
+      Web.Err.throw err404 { errBody = "User does not exist" }
+
+    Right _ -> 
+      redirect "https://fission.codes?verified=1"
```
expede, line 31: For clarity, let’s make it `=true` or `=t`

expede:
Overall looking pretty rad! We need to figure out that Heroku thing, though 😱

library/Fission/Challenge/Types.hs
dholms: :pray:

library/Fission/Email/SendInBlue/Request/Types.hs
dholms: :thinking: wouldn't make much sense for it to be

dholms: I'll change it to `NonEmpty`

library/Fission/Email/SendInBlue/TemplateId/Types.hs
dholms: left over from when I didn't know what it was :grimacing:

library/Fission/Web/User/Create.hs
dholms: oh yeahhhh good call. we could just not verify email? and auto-verify the user if it was created through heroku

dholms: On that note, should we even keep around this password registration route? Heroku doesn't use it, it uses `POST /heroku/resource`

expede: Yeah, I realized that much later 😛 

I guess the actual question is if we’re able to require an email in the database until we’ve figured the Heroku stuff out.

The password-based endpoint is deprecated. Let’s leave it there for now. I’d rather not cut it until we have a major version bump.

library/Fission/Types.hs
dholms: sweet got it workin :man_dancing:

env.yaml.example
```diff
@@ -34,3 +34,7 @@ fission_file_system:
 web_app:
   base_app_domain_name: fission.app
   app_placeholder_cid: QmRVvvMeMEPi1zerpXYH9df3ATdzuB63R1wf3Mz5NS5HQN # Fission logo on a blank page
+
+send_in_blue:
+  api_key: xkeysib-KEY
+  base_url: https://api.sendinblue.com
```
expede, line 40: 👍

library/Fission/Challenge.hs
```diff
@@ -0,0 +1,23 @@
...
+
+import qualified Fission.Web.Routes as Web 
+import qualified Fission.Web.User   as User
+
+verificationLink :: Challenge -> Text
+verificationLink challenge = 
+  toUrlPiece $ link challenge
+  where 
+    route = Proxy @(Web.UserPrefix :> User.VerifyEmailRoute)
+    link = safeLink (Proxy @Web.API) route
```
expede, line 23: This could be a good helper to use in many places:

```haskell
apiLink :: forall route . Proxy route -> MkLink route Link
apiLink pxy = safeLink (Proxy @Web.API) pxy
```

library/Fission/Challenge/Types.hs
```diff
@@ -0,0 +1,67 @@
...
+instance MimeRender PlainText Challenge where
+  mimeRender _ = UTF8.textToLazyBS . unChallenge
+
+instance MimeUnrender PlainText Challenge where
+  mimeUnrender _proxy bs =
+    bs
+      |> Lazy.toStrict
+      |> decodeUtf8'
+      |> bimap show Challenge
```
expede, line 54: A slightly more compact version:

```haskell
mimeUnrender _ = bimap show Challenge . decodeUtf8’ . Lazy.toStrict
```

library/Fission/Types.hs
```diff
@@ -563,3 +567,30 @@ instance Domain.Creator Fission where
...
+instance Challenge.Creator Fission where
+  create email = 
+    runDB $ Challenge.create email
+
+instance Challenge.Verifier Fission where
+  verify challenge = 
+    runDB $ Challenge.verify challenge
+
+instance MonadEmail Fission where
+  sendVerificationEmail recipient@(Email.Recipient { name }) challenge = do
```
expede, line 579: Protip: I’m 99.9999% sure that you don’t need the parens after the `@` because the record syntax keeps the fields “in” the constructor:

```haskell
recipient@Email.Recipient {name}
```

expede:
(I’m assuming that it actually sends email. I haven’t tried it, but the code looks good :P )

library/Fission/Challenge.hs
dholms: good call :+1:

library/Fission/Types.hs
dholms: yup you right :ok_hand:

PR #351, opened 2020-07-03
File stats

dholms:
## Problem
We don't currently have stats on the size of a user's storage

## Solution
Track size of user's data root & size of each app in DB
IPLD keeps track of the cumulative size of a node (including the entire subtree), so we can easily query that value each time the user updates their data root or the dnslink for an app

### DIFF
--- a/library/Fission/App/Creator.hs
+++ b/library/Fission/App/Creator.hs
@@ -1,6 +1,7 @@
 module Fission.App.Creator
   ( module Fission.App.Creator.Class
   , createWithPlaceholder
+  , createDB
   ) where
 
 import           Fission.Models
@@ -13,6 +14,11 @@ import           Fission.App.Creator.Class
 
 import           Fission.URL.Subdomain.Types
 
+import           Network.IPFS.Bytes.Types
+import           Network.IPFS.CID.Types
+
+import           Database.Esqueleto hiding ((<&>))
+
 createWithPlaceholder ::
   ( App.Creator        m
   , AppCID.Initializer m
@@ -23,3 +29,22 @@ createWithPlaceholder ::
 createWithPlaceholder ownerId now = do
   defaultCID <- AppCID.placeholder
   create ownerId defaultCID now
+
+createDB :: 
+     MonadIO m
+  => UserId 
+  -> CID 
+  -> Bytes 
+  -> UTCTime 
+  -> Transaction m AppId
+createDB ownerId cid size now = do
+  appId <- insert App
+    { appOwnerId    = ownerId
+    , appCid        = cid
+    , appSize       = size
+    , appInsertedAt = now
+    , appModifiedAt = now
+    }
+
+  _ <- insert $ CreateAppEvent appId ownerId cid size now
+  return appId
--- a/library/Fission/App/Creator/Class.hs
+++ b/library/Fission/App/Creator/Class.hs
@@ -3,10 +3,10 @@ module Fission.App.Creator.Class
   , Errors
   ) where
 
-import           Database.Esqueleto hiding ((<&>))
+import           Servant
 
 import           Network.IPFS.CID.Types
-import           Servant
+import qualified Network.IPFS.Get.Error as IPFS.Stat
 
 import           Fission.Prelude
 import           Fission.Error as Error
@@ -25,21 +25,14 @@ type Errors = OpenUnion
    , ActionNotAuthorized URL
    , NotFound            URL
 
+   , IPFS.Stat.Error
+
    , InvalidURL
    ]
 
 class Monad m => Creator m where
-  create :: UserId -> CID -> UTCTime -> m (Either Errors (AppId, Subdomain))
-
-instance (MonadIO m, App.Domain.Initializer m) => Creator (Transaction m) where
-  create ownerId cid now = do
-    appId <- insert App
-      { appOwnerId    = ownerId
-      , appCid        = cid
-      , appInsertedAt = now
-      , appModifiedAt = now
-      }
-
-    App.Domain.associateDefault ownerId appId now <&> \case
-      Left  err       -> Error.relaxedLeft err
-      Right subdomain -> Right (appId, subdomain)
+  create :: 
+       UserId 
+    -> CID 
+    -> UTCTime 
+    -> m (Either Errors (AppId, Subdomain))
--- a/library/Fission/Internal/Fixture/User.hs
+++ b/library/Fission/Internal/Fixture/User.hs
@@ -1,6 +1,7 @@
 module Fission.Internal.Fixture.User (user) where
 
 import           Network.IPFS.CID.Types
+import           Network.IPFS.Bytes.Types
 
 import           Fission.Prelude
 import           Fission.Models
@@ -12,8 +13,12 @@ import qualified Fission.Internal.Fixture.Key.Ed25519 as Ed25519
 
 user :: User
 user = User
-  { userUsername = "testUser"
+  { userPublicKey = Just Ed25519.pk
+
+  --
+
   , userEmail    = Just "test@fission.codes"
+  , userUsername = "testUser"
 
   --
 
@@ -23,8 +28,8 @@ user = User
 
   --
 
-  , userPublicKey = Just Ed25519.pk
-  , userDataRoot  = CID "QmW2WQi7j6c7UgJTarActp7tDNikE4B2qXtFCfLPdsgaTQ"
+  , userDataRoot     = CID "QmW2WQi7j6c7UgJTarActp7tDNikE4B2qXtFCfLPdsgaTQ"
+  , userDataRootSize = Bytes 0
 
   --
 
--- a/library/Fission/Internal/Orphanage/Bytes.hs
+++ b/library/Fission/Internal/Orphanage/Bytes.hs
@@ -1,0 +1,26 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+module Fission.Internal.Orphanage.Bytes () where
+
+import           Fission.Prelude
+
+import           Network.IPFS.Bytes.Types
+
+import           Database.Persist.Class
+import           Database.Persist.Types
+import           Database.Persist.Sql
+
+
+instance PersistField Bytes where
+  toPersistValue (Bytes bytes') = PersistInt64 $ fromIntegral bytes'
+  fromPersistValue = \case
+    PersistInt64 bytes' -> 
+      if bytes' >= 0
+        then Right . Bytes $ fromIntegral bytes'
+        else Left ("Invalid Persistent Bytes: " <> textShow bytes')
+
+    other -> 
+      Left ("Invalid Persistent Bytes: " <> textShow other)
+
+instance PersistFieldSql Bytes where
+  sqlType _pxy = SqlInt64
--- a/library/Fission/Models.hs
+++ b/library/Fission/Models.hs
@@ -15,6 +15,7 @@ import           Data.Swagger
 import           Data.UUID
 
 import           Network.IPFS.CID.Types
+import           Network.IPFS.Bytes.Types
 
 import qualified Fission.Internal.UTF8                as UTF8
 import           Fission.Prelude
@@ -32,6 +33,7 @@ import           Fission.Challenge.Types
 
 import qualified Fission.AWS.Zone.Types               as AWS
 
+import           Fission.Internal.Orphanage.Bytes     ()
 import           Fission.Internal.Orphanage.CID       ()
 import           Fission.Internal.Orphanage.UUID      ()
 
@@ -64,6 +66,7 @@ User
   verified      Bool                 default=false
 
   dataRoot      CID
+  dataRootSize  Bytes
 
   herokuAddOnId HerokuAddOnId Maybe
   secretDigest  SecretDigest  Maybe
@@ -93,10 +96,12 @@ UserChallenge
 ------------
 
 UpdateUserDataRootEvent
-  userId      UserId
-  newDataRoot CID
+  userId          UserId
 
-  insertedAt  UTCTime
+  newDataRoot     CID
+  newDataRootSize Bytes
+
+  insertedAt      UTCTime
 
   deriving Show Eq
 
@@ -134,6 +139,8 @@ App
   ownerId     UserId
   cid         CID
 
+  size        Bytes
+
   insertedAt  UTCTime
   modifiedAt  UTCTime
 
@@ -144,8 +151,11 @@ App
 ------------
 
 CreateAppEvent
+  appId       AppId
   ownerId     UserId
+
   cid         CID
+  size        Bytes
 
   insertedAt  UTCTime
 
@@ -158,8 +168,10 @@ DestroyAppEvent
   deriving Show Eq
 
 SetAppCIDEvent
-  appId  AppId
-  newCID CID
+  appId      AppId
+
+  newCID     CID
+  newCIDSize Bytes
 
   insertedAt UTCTime
 
--- a/library/Fission/Types.hs
+++ b/library/Fission/Types.hs
@@ -23,6 +23,7 @@ import qualified Network.IPFS.Process.Error as IPFS.Process
 import qualified Network.IPFS.Pin           as IPFS.Pin
 import qualified Network.IPFS.Process       as IPFS
 import qualified Network.IPFS.Peer          as Peer
+import qualified Network.IPFS.Stat          as IPFS.Stat
 
 import           Fission.Config.Types
 import           Fission.Prelude
@@ -44,7 +45,6 @@ import qualified Fission.App.Destroyer as App.Destroyer
 import qualified Fission.Web.Error                     as Web.Error
 import           Fission.Web.Types
 
-import qualified Fission.App.Creator as App
 
 import           Fission.IPFS.Linked
 import qualified Fission.Platform.Heroku.AddOn.Creator as Heroku.AddOn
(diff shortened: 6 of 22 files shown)


library/Fission/Internal/Fixture/User.hs
```diff
@@ -12,8 +12,12 @@ import qualified Fission.Internal.Fixture.Key.Ed25519 as Ed25519
 
 user :: User
 user = User
-  { userUsername = "testUser"
+  { userPublicKey = Just Ed25519.pk
```
dholms, line 15: just reorganizing these to match the layout of the Model

library/Fission/App/Creator/Class.hs
```diff
@@ -29,17 +29,25 @@ type Errors = OpenUnion
...
+  create ownerId cid size now = do
     appId <- insert App
       { appOwnerId    = ownerId
       , appCid        = cid
+      , appSize       = size
       , appInsertedAt = now
       , appModifiedAt = now
       }
 
+    _ <- insert $ CreateAppEvent appId ownerId cid size now
```
dholms, line 49: we weren't inserting `CreateAppEvent`, so I added that in
I also added `appId` to the model

library/Fission/App/Creator.hs
```diff
@@ -22,4 +22,4 @@ createWithPlaceholder ::
   -> m (Either Errors (AppId, Subdomain))
 createWithPlaceholder ownerId now = do
   defaultCID <- AppCID.placeholder
-  create ownerId defaultCID now
+  create ownerId defaultCID 0 now
```
expede, line 25: 👍

I was like “magic number?!” I guess zero is pretty magical 😂

library/Fission/Models.hs
```diff
@@ -62,6 +62,7 @@ User
   active        Bool
 
   dataRoot      CID
+  dataRootSize  Natural
```
expede, line 65: Let’s add a unit!

```haskell
newtype Bytes = Natural
-- And then instances to and from SQL fields. There are examples elsewhere in the code.
```

There’s a nontrivial chance that this already exists in a package. I’ll look it up!

library/Fission/Models.hs
```diff
@@ -80,10 +81,12 @@ User
 ------------
 
 UpdateUserDataRootEvent
-  userId      UserId
-  newDataRoot CID
+  userId          UserId
 
-  insertedAt  UTCTime
+  newDataRoot     CID
+  newDataRootSize Natural
```
expede, line 87: Same here

library/Fission/Models.hs
```diff
@@ -121,6 +124,8 @@ App
   ownerId     UserId
   cid         CID
 
+  size        Natural
```
expede, line 127: And here. For example, is this the tree size (number of links) or the size on disk (bytes). A newtype wrapper can help!

library/Fission/Models.hs
expede: Ah, right the fancy version (optional) is something like this:

```haskell
data Bytes  -- No need for a `=` because we only need the type
data KiloBytes
data Meters
data Grams

newtype Quantity unit = Quantity Natural
  deriving newtype (Show, Eq, Integral, AndEvenMoreStuff)

4 :: Quantity Bytes
56 :: Quantity Meters
```

library/Fission/Types.hs
```diff
@@ -446,9 +446,14 @@ instance User.Modifier Fission where
...
-              Left err -> Error.relaxedLeft err
-              Right _  -> ok
+            DNSLink.set userId url zoneID newCID >>= \case
+              Left err -> 
+                return $ Error.relaxedLeft err
+
+              Right _  -> 
+                IPFS.Pin.add newCID >>= \case
+                  Right _  -> return ok
+                  Left err -> return $ Error.openLeft err
```
expede, line 456: Ya, agreed, let’s standardize on `>>=`

library/Fission/Types.hs
```diff
@@ -478,8 +483,8 @@ instance App.Creator Fission where
           Right _   -> Right (appId, subdomain)
 
 instance App.Modifier Fission where
-  setCID userId url newCID copyFiles now = do
-    runDB (App.setCID userId url newCID copyFiles now) >>= \case
+  setCID userId url newCID size copyFiles now = do
```
expede, line 486: Let’s be super clear: `bytes`

library/Fission/User/Creator/Class.hs
```diff
@@ -105,7 +111,7 @@ instance
               determineConflict username (Just pk)
 
             Just userId ->
-              User.setData userId App.Content.empty now >>= \case
+              User.setData userId App.Content.empty 0 now >>= \case
```
expede, line 114: We’re using this in a few places, eh? Thoughts on this:

```haskell
User.createData = User.setData userId App.Content.empty 0 now
```

library/Fission/User/Modifier/Class.hs
```diff
@@ -21,6 +24,9 @@ type Errors = OpenUnion
    , NotFound            URL
    , ActionNotAuthorized URL
 
+   , IPFS.Pin.Error
+   , IPFS.Stat.Error
```
expede, line 28: Yaaaaas

library/Fission/User/Modifier/Class.hs
```diff
@@ -4,9 +4,12 @@ module Fission.User.Modifier.Class
   ) where
 
 import           Database.Persist as Persist
-import           Network.IPFS.CID.Types
 import           Servant.Server
 
+import qualified Network.IPFS.Add.Error as IPFS.Pin
+import qualified Network.IPFS.Get.Error as IPFS.Stat
```
expede, line 10: Stat or Get? Looks like both modules exist 🤔

library/Fission/Web/App.hs
```diff
@@ -32,6 +33,7 @@ server ::
   , MonadTime               m
   , MonadLogger             m
   , MonadDNSLink            m
+  , MonadRemoteIPFS         m
```
expede, line 36: 🤔 How was this not here before?!

expede: Oooh, we had actually abstracted away from this with the app classes. I’m tempted to actually make a `App.Stats` class, or even just a method on `App.Retriever`.

library/Fission/Web/App/Update.hs
```diff
@@ -23,16 +27,18 @@ type API
...
+  , MonadTime       m
+  , MonadRemoteIPFS m
+  , App.Modifier    m
   )
   => Authorization
   -> ServerT API m
 update Authorization {about = Entity userId _} url newCID copyDataFlag = do
   now <- currentTime
-  Web.Error.ensureM $ App.setCID userId url newCID copyFiles now
+  size <- Web.Error.ensureM $ IPFS.Stat.getSizeRemote newCID
```
expede, line 40: Yeah, this one (comment above). Why not make `getSizeRemote` `User.getSize` and `App.getSize`, which reference these?

expede: I like the use of `ensureM`, though 👍

library/Fission/Web/DNS.hs
```diff
@@ -26,13 +29,15 @@ type API
...
 -- WITHOUT the `files` prefix
 server ::
-  ( MonadTime     m
-  , MonadThrow    m
-  , MonadLogger   m
-  , User.Modifier m
+  ( MonadTime       m
+  , MonadThrow      m
+  , MonadLogger     m
+  , MonadRemoteIPFS m
```
expede, line 35: Same comment as earlier

stack.yaml
```diff
@@ -44,7 +44,9 @@ extra-deps:
 - cryptostore-0.2.1.0
 - happy-1.19.12
 - hspec-wai-json-0.10.1
-- ipfs-1.0.3
+# - ipfs-1.0.3
+- git: https://github.com/fission-suite/ipfs-haskell.git
+  commit: 0b52e01fc53f05912350a413be6e246bc8d0e919
```
expede, line 49: Should we publish this? 👀

dholms: yup! but probably want a review first: https://github.com/fission-suite/ipfs-haskell/pull/15

expede: Approved

library/Fission/Types.hs
dholms: if the type is `Bytes` do we also want the variable named `Bytes`. I feel like `size: Bytes` is most clear

library/Fission/User/Modifier/Class.hs
dholms: There's not `Stat.Error`

`Stat` returns the same type of errors as `Get` (in the same way that `Pin` returns the same type of errors as `Add` if you look at the prev input)

These could probably use a renaming in the ipfs pkg :thinking: maybe `Modify.Error` & `Retrieval.Error` since pins & adds are both "modifies" & gets & stats are both "retrievals". and neither of those words can be confused with actual ipfs commands

library/Fission/Web/App.hs
dholms: yeah i'll put a method on `App.Retriever` I like the sound of that :+1:

dholms: actually just kidding. `App.Retriever` is all DB functions and I feel like we should have a separation in our head between "hit the DB and find something out about this appID" vs "figure out some information about the app at this CID". So maybe `App.Stats` does make more sense

dholms: hmm and the same function is used to update a user's data root. So I either we're using a `MonadRemoteIPFS` there or we're using `App.Stats` there which doesn't seem to make that much sense.

So maybe we have a `MonadIPFSStat` or similar? But that's essentially just a subset of `MonadRemoteIPFS` :thinking:

dholms: Actually the more i think about it, we're doing the same thing for a user's data root as well. So as you mention below, we could do App.Stats & User.Stats, but we're actually doing all of these operations on a CID before the CID is associated with the user/app.  So I think it makes most sense to think of it an IPFS operation not a user/app operation. 

It might make more sense to have something like `MonadIpfsStat` or similar. But at that point, it would essentially just be some subset of `MonadRemoteIPFS`'s functionality (and less descriptive, because "Remote IPFS" gives you more info about where it's coming from). idk, thoughts? :thinking:

expede: > actually just kidding. `App.Retriever` is all DB functions and I feel like we should have a separation in our head between "hit the DB and find something out about this appID" vs "figure out some information about the app at this CID". So maybe `App.Stats` does make more sense

Kind of the idea is that you don’t care if it’s hitting the DB or returning a hardcoded value, or pinging another API. It’s meant to be fully abstracted away from these details. You’re “retrieving user data”, not “accessing postgres”

expede: > we're actually doing all of these operations on a CID before the CID is associated with the user/app

I’m not following. Aren’t we getting the stats for what the user is responsible for in general? Not the CID, the user, the app, &c. I can imagine us having things like “certain CIDs are free because deduplication or part of our standard library”. We’re still checking the remote underneath, but the semantics we want are “check the user” not “check the CID”. We still need the remote IPFS _underneath_, but it’s not the effect that we care about — it’s a detail. Unless I’m missing something?

library/Fission/User/Modifier/Class.hs
expede: Okay, cool cool. Just wanted to make sure it was right. Maybe add a note. No need to change on the IPFS side right now 👍

library/Fission/Web/App.hs
dholms: ok I see what you're saying. I guess it's weird to me because we'd have `App.Stat` & `User.Stat` monads that have a method 
`getSize :: CID -> m (Either Error Bytes)` with the exact same implementation that doesn't take into account if it's an App or User.

But I guess the point is we're abstracting the IPFS details away and getting it set up in such a way that we could make it take into account some app/user details in the future

> we're actually doing all of these operations on a CID before the CID is associated with the user/app

what i meant by this is, the order of operations goes:
- user sends CID that they want to update their data root to
- we check the size of the CID in IPFS
- we update the cid & size of the user's data root in the DB (& do any other side effects like DNS)

so step 2 (which is what we're discussing) is totally agnostic to if this is a user/app & in our system, the CID isn't even associated with a user/app yet

expede: But is that what we’re trying to do? This is for stats on the user, not stats on arbitrary CIDs. In the future, we could easily want to call this function at other times than on data update, or do other calculation as part of that number (shared storage across multiple users, company accounts, and so on)

expede: Also happy to hop on a call (tomorrow) to chat about the tradeoffs here

expede: Thinking about this a bit more (and I need to read the code again), for the use case you mentioned above where we’re checking on write, we can probably just drop the constraint completely since it’s part of the same action.

Let’s chat more tomorrow 👍

library/Fission/Internal/Orphanage/Bytes.hs
```diff
@@ -0,0 +1,24 @@
...
+
+
+instance PersistField Bytes where
+  toPersistValue (Bytes bytes') = PersistInt64 $ fromIntegral bytes'
+  fromPersistValue = \case
+    PersistInt64 bytes' -> 
+      if bytes' >= 0
+        then Right . Bytes $ fromIntegral bytes'
+        else Left ("Invalid Persistent Bytes: " <> textShow bytes')
+    other               -> Left ("Invalid Persistent Bytes: " <> textShow other)
```
expede, line 21: No need to change, but in general we want the layout to be the same. If you’re dropping a line for `PersistInt64`, then drop a line for `other`

library/Fission/Types.hs
expede: Yup, great 👍

library/Fission/User/Creator/Class.hs
```diff
@@ -105,7 +112,7 @@ instance
               determineConflict username (Just pk)
 
             Just userId ->
-              User.setData userId App.Content.empty now >>= \case
+              User.setData userId App.Content.empty (Bytes 0) now >>= \case
```
expede, line 115: 👍

library/Fission/Web/DNS.hs
```diff
@@ -26,10 +27,10 @@ type API
...
 -- WITHOUT the `files` prefix
 server ::
-  ( MonadTime     m
-  , MonadThrow    m
-  , MonadLogger   m
-  , User.Modifier m
+  ( MonadTime       m
+  , MonadThrow      m
+  , MonadLogger     m
+  , User.Modifier   m
```
expede, line 33: Why?

library/Fission/Web/User.hs
```diff
@@ -67,11 +67,11 @@ type ResetRoute
...
-  ( MonadDNSLink  m
-  , MonadLogger   m
-  , MonadTime     m
-  , User.Modifier m
-  , User.Creator  m
+  ( MonadDNSLink    m
+  , MonadLogger     m
+  , MonadTime       m
+  , User.Modifier   m
+  , User.Creator    m
```
expede, line 74: ?

library/Fission/Web/DNS.hs
```diff
@@ -4,19 +4,20 @@ module Fission.Web.DNS
...
 import           Fission.Prelude
 
 import           Fission.URL                 as URL
-import           Fission.User.Username.Types
 import           Fission.Web.Error           as Web.Err
 
+import           Fission.User.Username.Types
 import qualified Fission.User.Modifier       as User
 
+import           Network.IPFS.CID.Types
```
expede, line 19: No need to change, but FYI the import layout is `general -> specific use case`. This import would be up at line 8.

library/Fission/Models.hs
```diff
@@ -62,6 +64,7 @@ User
   active        Bool
 
   dataRoot      CID
+  dataRootSize  Bytes
```
expede, line 67: Yaaaaaaaaaaaaaas

library/Fission/Web/DNS.hs
dholms: old formatting :unamused: 
really need to get `stylish-haskell` working

library/Fission/Web/DNS.hs
dholms: ah cool :+1: hadn't quite figured out the rhyme or reason to import layout

library/Fission/Models.hs
dholms: :man_dancing:

library/Fission/User/Creator.hs
```diff
@@ -1,7 +1,159 @@
...
+        , userInsertedAt    = now
+        , userModifiedAt    = now
+        }
+        |> insertUnique
+        |> bind \case
+          Just userId -> return $ Right userId
+          Nothing -> determineConflict username Nothing
+
+
+createWithHerokuDB ::
```
expede, line 99: How likely are we to use this funtion in multiple places? Why not just stick it right in `Fission.Types`?

expede: If it’s “that file is getting long” — this is another case of “when we break up the subprojects, we can use less abstraction, and thus less in that file”

expede:
Don’t forget to bump the version number!

library/Fission/User/Creator.hs
dholms: yeah it was "this file is getting long" & I liked maintaining that separation of "database" & "everything else"

I do use `createDB` in `server/Main` to ensure the default user is created since we're not in `Fission` yet

Issue #352, opened 2020-07-06
Add prefix to TXT records so we can join them in order


icidasset:
## Problem

When we lookup a user's root DID using the `_did.username.fission.name` domain, the DID is split into multiple TXT records (in the case of an RSA DID). The problem is that the order of these records is random, sometimes the first record is `did:key:...`, other times it's a different part of the DID. Tested with https://dnsrecords.io and the `dig` command.

## Solution

Add an ordering prefix to the record.
For example: `1:2:qQxvcYxBbJZ51qc5FWGu8u5Vcgh4G78CnD2AAkfXp35rsjuAAYCcqvKREdGBUe6GADyp1FRGvePwnixamyjkoyQtmpgJFHudRXH`.
Format: `which_key:which_part:chunk`

expede:
It would also lower the parsing overhead to have a fixed size header. We know that TXT can take up to 255 fields. We can either count in hex or decimal:

Decimal:
```
001;did:key:z12345
220;did:key:z12345
```

Hex
```
01;did:key:z12345
DC;did:key:z12345
```

dholms:
Cool I like just the one number (and finding key breaks by looking for `did:key` or similar) & using a semicolon as a delimiter to avoid confusion

> hex or decimal

my initial reaction was hex just cause we're in crypto-land and things are seldom in decimal + slight space savings. but as you said it doesn't really matter, and it's only one character. And it's prefixed in front of the human readable "did:key" so it would also make sense to have the index be a human readable decimal number

dholms:
Follow up question:

dholms:
oh also 0-indexed or 1-indexed?

expede:
> oh also 0-indexed or 1-indexed?

Yeah, good point. For sure 0-indexed 💯

> I'll probably start on decimal unless I hear otherwise

Decimal is also my feeling on this. Let's just manage it in software. Pretty easy to keep it straight on that side, PLUS we can't put more parts than TXT allows, so great

PR #353, opened 2020-07-06
Ordered TXT records

dholms:
## Problem
When we lookup a user's root DID using the `_did.username.fission.name` domain, the DID is split into multiple TXT records (in the case of an RSA DID). The problem is that the order of these records is random, sometimes the first record is `did:key:`..., other times it's a different part of the DID. Tested with https://dnsrecords.io and the `dig` command.

## Solution
Add an fixed-size decimal prefix to the record that indicates it's ordering:

```
[
  "000;did:key:zhdjsakd3uia",
  "001;hsajhdjashkjdsajkldj",
  "002;djksaljdklasjldasjkd",
  "003;did:key:z12sadsha345",
  "004;hjsdkauismadml"
]
```

_Note: do we need the delimiter if the prefix is fixed-size?_

Closes https://github.com/fission-suite/fission/issues/352


library/Fission/DNS.hs
```diff
@@ -0,0 +1,34 @@
...
+  if Text.length txt <= 256
+    then 
+      pure txt
+
+    else 
+      Text.chunksOf 252 txt
+      |> List.zip [0..]
+      |> (joinRecord <$>)
+      |> nonEmpty
+      |> maybe (pure "") identity
```
expede, line 23: Two quick points on style:

1. Pipes _must_ be indented. The above wold not work in a `do` block
2. When piping, start with a bae argument (in this case, `txt |> Text.chunksOf`
3. I find the piped version here much harder to read
4. Why not just keep it consistent with a prefix always?
5. It can’t hurt to break these steps out with names

```haskell 
splitRecord txt = 
  fromMaybe (pure “”) records
  where
    records = joinRecord <$> indexed
    indexed = List.zip [0..] Text.chunksOf 252 txt
```

library/Fission/DNS.hs
```diff
@@ -0,0 +1,34 @@
...
+
+joinRecord :: (Integer, Text) -> Text
+joinRecord (index, txt) =  
+  textShow index
+  |> padToLen 3 '0'
+  |> (<> ";" <> txt)
```
expede, line 29: Also would loose the pipes here, especially with the `<>`s

```haskell
padToLen 3 ‘0’ (textDisplay index) <> “;” <> txt
```

library/Fission/DNS.hs
```diff
@@ -0,0 +1,34 @@
...
+
+padToLen :: Int -> Char -> Text -> Text
+padToLen len char txt = do
+  let paddingLen = max 0 $ len - Text.length txt
+  Text.replicate paddingLen (Text.singleton char) <> txt
```
expede, line 34: I didn’t even know that Text was a monad :P This version is fine, but you have two other options:

```haskell

padToLen len char txt =
  let paddingLen = max 0 $ len - Text.length txt
  in Text.replicate paddingLen (Text.singleton char) <> txt

padToLen len char txt =
  Text.replicate paddingLen (Text.singleton char) <> txt
  where
    paddingLen = max 0 (len - Text.length txt)
```

library/Fission/Types.hs
```diff
@@ -416,15 +418,11 @@ instance User.Modifier Fission where
               subdomain = Just $ Subdomain rawUN
               url       = URL {domainName, subdomain = Just (Subdomain "_did") <> subdomain}
               did       = textDisplay (DID pk Key)
-              (_, didSegments) = Text.foldr splitter (0, ("" :| [])) did
+              didSegments = DNS.splitRecord did
```
expede, line 421: 👍

test/Test/Fission/DNS.hs
```diff
@@ -0,0 +1,46 @@
...
+
+import qualified Fission.DNS          as DNS
+import qualified RIO.NonEmpty.Partial as NonEmpty.Partial
+
+tests :: IO TestTree
+tests =
+  testSpec "Fission.DNS" do
+    describe "splitRecord" do
+      it "does not split a string under 256 chars" do
+        DNS.splitRecord "asdf" `shouldBe` pure "asdf"
```
expede, line 13: Why not make these prop tests? You can control the size a couple ways, but the most straightforward without having to learn the library would be something like:

```haskell
-- Helper type, maybe right in the test file
newtype SmallText = SmallText Text
  deriving newtype (Show, Eq)

instance Arbitrary SmallText where
  arbitrary = do
    txt <- arbitrary
    return . SmallText $ Text.take 256 txt

-- Tests

testSpec “Fission.DNS” d
  describe “...” do
    context “under 256 characters” do
      itsProp’ “length does not change ” \SmallText txt ->
        DNS.splitRecord txt `shouldBe pure txt
````

File: library/Fission/DNS.hs [hunk 6]
```diff
@@ -0,0 +1,34 @@
...
+      |> nonEmpty
+      |> maybe (pure "") identity
+
+joinRecord :: (Integer, Text) -> Text
+joinRecord (index, txt) =  
+  textShow index
+  |> padToLen 3 '0'
+  |> (<> ";" <> txt)
+
+padToLen :: Int -> Char -> Text -> Text
```
expede, line 31: I’d make this a `Natural` to avoid having to deal with negative integers

library/Fission/DNS.hs
expede: You may actually be able to get rid of `padToLen` with some laziness:

```haskell
Text.takeEnd 3 (Text.replicate 2 “0” <> textDisplay index) <> ”;” <> txt
```

expede: As mentioned elsewhere, `Natural` would also be good here. We don’t have negative indexes.

Also, I’m using `textDisplay` instead of `textShow` since that’s actually built-in to RIO, and likely is very efficient.

dholms: > Text.takeEnd 3 (Text.replicate 2 “0” <> textDisplay index) <> ”;” <> txt
oh yeah this is slick

library/Fission/DNS.hs
dholms: Nice! tbh I kinda forgot about `where` clauses :sweat_smile:  this reads much clearer

dholms: > Pipes must be indented. The above wold not work in a do block 

:+1:

dholms: > Why not just keep it consistent with a prefix always?
:thinking: we _can_ but TXT records aren't normally treated that way. so I thought this was actually more straightforward: If you get one record back, that's the record, no parsing needed (which is what you expect from DNS). If you get back multiple records where order matters, then prefixes help order them

test/Test/Fission/DNS.hs
dholms: cool :+1:  this is really helpful

Should I be prop testing longer strings as well? seems like i'll essentially be re-coding the function to ensure that it works, but maybe that's good?

PR #354, opened 2020-07-14
Fix DNS.splitRecord tests

dholms:
## Problem
The test for splitting arbitrary large DNS records was hanging

## Solution
Change the arbitrary instance so it doesn't hang (using `Text.replicate`)

### DIFF
--- a/test/Test/Fission/DNS.hs
+++ b/test/Test/Fission/DNS.hs
@@ -23,6 +23,8 @@ tests =
           where
             splitCount = NonEmpty.length . DNS.splitRecord
             expectedCount txt = (Text.length txt `div` 252) + 1
+      
+      -- add a serialize/deserialize test once we add `DNS.combineRecords`
 
 newtype SmallText = SmallText Text
   deriving newtype (Show, Eq)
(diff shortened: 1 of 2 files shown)


test/Test/Fission/DNS.hs
```diff
@@ -13,15 +16,13 @@ tests =
...
-      --     splitCount txt `shouldBe` expectedCount txt
+      context "over 256 characters" do
+        itsProp' "length does not change" \(LargeText txt) -> 
+          splitCount txt `shouldBe` expectedCount txt
 
-      --     where
-      --       splitCount = NonEmpty.length . DNS.splitRecord
-      --       expectedCount txt = (Text.length txt `div` 252) + 1
+          where
+            splitCount = NonEmpty.length . DNS.splitRecord
```
expede, line 24: Yep looking good! Let’s also add a prop test that it serializes and deserializes correctly (I’m assuming that this should be trivial — if not just make a note)

expede: i.e.

```haskell
fromDNS (forDNS txt) `shouldBe` txt
```

expede: But yeah, I’ll let you use your judgement on that. LGTM!

dholms: we don't have a function for recombining records yet (I think that functionality is only needed in the SDK at the moment). so I'll hold off on that & add a note for when/if we add it

PR #358, opened 2020-07-17
Fix setup bug

dholms:
## Problem
If setup fails, the key sticks around (even tho it hasn't been linked to an account) and confuses the CLI

## Solution
On setup, check if the local key is associated with an account (`whoami`) instead of just if it exists. Also revamp the message to show the username that you're currently logged in with

### DIFF
--- a/library/Fission/CLI/Command/Setup.hs
+++ b/library/Fission/CLI/Command/Setup.hs
@@ -18,6 +18,7 @@ import           Fission.Web.Auth.Token
 import           Fission.Web.Client      as Client
 import qualified Fission.Web.Client.User as User
 
+import qualified Fission.User.Username.Types     as User
 import           Fission.User.Registration.Types
 import           Fission.User.Email.Types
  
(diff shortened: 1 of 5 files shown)


library/Fission/CLI/Command/Setup.hs
```diff
@@ -56,13 +57,13 @@ setup ::
...
-      CLI.Success.putOk "You are already connected"
+setup =
+  sendRequestM (authClient $ Proxy @User.WhoAmI) >>= \case
+    Right User.Username {username} ->
+      CLI.Success.alreadyLoggedInAs username
 
-    False -> do
-      maybe createAccount upgradeAccount =<< Env.Override.findBasicAuth -- NOTE Deprecated
+    Left _ ->
+      maybe createAccount upgradeAccount =<< Env.Override.findBasicAuth
```
expede, line 66: Good call

library/Fission/CLI/Display/Success.hs
```diff
@@ -30,10 +31,15 @@ dnsUpdated domain = do
   UTF8.putText "📝 DNS updated! Check out your site at: \n"
   UTF8.putText $ "🔗 " <> textDisplay domain  <> "\n"
 
-loggedInAs :: MonadIO m => Text -> m ()
-loggedInAs username = liftIO do
-  UTF8.putText "💻 Currently logged in as: "
+currentlyLoggedInAs :: MonadIO m => Text -> m ()
+currentlyLoggedInAs = loggedInAs "Currently logged in as: "
+
+alreadyLoggedInAs :: MonadIO m => Text -> m ()
```
expede, line 37: 👍

PR #366, opened 2020-07-31
Set up Niv

expede:


expede:
Failing CI because of the `nix: true` but no Nix in the GH setup script.

library/Fission/CLI/Config/Base/Types.hs
```diff
@@ -161,6 +161,7 @@ instance MonadWebAuth FissionBase Ed25519.SecretKey where
 instance MonadLocalIPFS FissionBase where
   runLocal opts arg = do
     IPFS.BinPath ipfs <- asks ipfsPath
+    logInfo $ ">>>>>>>>>>>>>>>>>" <> Text.pack ipfs
```
brainrape, line 164: oops

expede: Ha! Thanks :)

PR #372, opened 2020-08-04
Connect to multiple peers

dholms:
## Problem
The CLI is not connecting to our remote node

## Solution
There are two fixes here:
- make sure that we're writing the entire peerlist from the server to `~/.fission.yaml` (we were just writing the tail for some reason)
- attempt connection to the entire peerlist (we were just attempting to connect to the head of the list)

### DIFF
--- a/library/Fission/CLI/Config/Connected.hs
+++ b/library/Fission/CLI/Config/Connected.hs
@@ -10,6 +10,7 @@ import qualified Crypto.PubKey.Ed25519 as Ed25519
 import           Network.IPFS
 
 import           Fission.Prelude
+import qualified RIO.NonEmpty as NonEmpty
  
 import           Fission.Authorization.ServerDID
 import qualified Fission.Key as Key
@@ -28,7 +29,6 @@ import           Fission.CLI.Environment.Types as Environment
 import qualified Fission.CLI.Environment       as Environment
 import qualified Fission.CLI.IPFS.Connect      as Connect
 
-import qualified RIO.NonEmpty as NonEmpty
 
 -- | Ensure we have a local config file with the appropriate data
 --
--- a/library/Fission/CLI/Config/Connected.hs
+++ b/library/Fission/CLI/Config/Connected.hs
@@ -28,6 +28,8 @@ import           Fission.CLI.Environment.Types as Environment
 import qualified Fission.CLI.Environment       as Environment
 import qualified Fission.CLI.IPFS.Connect      as Connect
 
+import qualified RIO.NonEmpty as NonEmpty
+
 -- | Ensure we have a local config file with the appropriate data
 --
 -- Takes a @Connected@-dependant action, and lifts it into an environment that
@@ -64,13 +66,14 @@ liftConfig BaseConfig {..} = do
     Right secretKey -> do
       config <- Environment.get
  
-      Environment.getOrRetrievePeer config >>= \case
+      maybePeers <- Environment.getOrRetrievePeers config 
+      case NonEmpty.nonEmpty maybePeers of
         Nothing -> do
           CLI.Error.notConnected PeersNotFound
           return $ Left PeersNotFound
 
-        Just peer ->
-          Connect.swarmConnectWithRetry peer 1 >>= \case
+        Just peers ->
+          Connect.swarmConnectWithRetry peers 1 >>= \case
             Left err -> do
               logDebug $ displayShow err
               Connect.couldNotSwarmConnect
--- a/library/Fission/CLI/Config/Connected/Types.hs
+++ b/library/Fission/CLI/Config/Connected/Types.hs
@@ -43,7 +43,7 @@ data ConnectedConfig = ConnectedConfig
   , processCtx   :: !ProcessContext
   , ipfsPath     :: !IPFS.BinPath
   , ipfsTimeout  :: !IPFS.Timeout
-  , peer         :: !IPFS.Peer
+  , peers        :: !(NonEmpty IPFS.Peer)
   , ignoredFiles :: !IPFS.Ignored
   }
 
--- a/library/Fission/CLI/Environment.hs
+++ b/library/Fission/CLI/Environment.hs
@@ -5,7 +5,7 @@ module Fission.CLI.Environment
   , getPath
   , couldNotRead
   , removeConfigFile
-  , getOrRetrievePeer
+  , getOrRetrievePeers
  
   -- * Reexport
 
@@ -94,29 +94,30 @@ removeConfigFile = do
 
 -- | Retrieves a Fission Peer from local config
 --   If not found we retrive from the network and store
-getOrRetrievePeer ::
+getOrRetrievePeers ::
   ( MonadUnliftIO  m
   , MonadLogger    m
   , MonadWebClient m
   )
   => Environment
-  -> m (Maybe IPFS.Peer)
-getOrRetrievePeer Environment {peers = (peer : _)} = do
-  logDebug @Text "Retrieved Peer from .fission.yaml"
-  return $ Just peer
-
-getOrRetrievePeer Environment {peers = []} =
+  -> m [IPFS.Peer]
+getOrRetrievePeers Environment {peers = []} =
   Peers.getPeers >>= \case
     Left err -> do
       logError $ displayShow err
       logDebug @Text "Unable to retrieve peers from the network"
-      return Nothing
+      return []
 
     Right nonEmptyPeers -> do
-      logDebug @Text "Retrieved Peer from API"
       path <- globalEnv
-      Override.writeMerge path $ mempty { peers = NonEmpty.toList nonEmptyPeers }
-      return . Just $ NonEmpty.head nonEmptyPeers
+      let peers = NonEmpty.toList nonEmptyPeers
+      logDebug $ "Retrieved Peers from API, and writing to ~/.fission.yaml: " <> textShow peers
+      Override.writeMerge path $ mempty { peers }
+      return peers
+
+getOrRetrievePeers Environment {peers = peers} = do
+  logDebug $ "Retrieved Peers from .fission.yaml: " <> textShow peers
+  return peers
 
 ignoreDefault :: IPFS.Ignored
 ignoreDefault =
--- a/library/Fission/CLI/Environment.hs
+++ b/library/Fission/CLI/Environment.hs
@@ -115,7 +115,7 @@ getOrRetrievePeers Environment {peers = []} =
       Override.writeMerge path $ mempty { peers }
       return peers
 
-getOrRetrievePeers Environment {peers = peers} = do
+getOrRetrievePeers Environment {peers} = do
   logDebug $ "Retrieved Peers from .fission.yaml: " <> textShow peers
   return peers
 
--- a/library/Fission/CLI/Environment/Override.hs
+++ b/library/Fission/CLI/Environment/Override.hs
@@ -83,7 +83,7 @@ write path env = writeBinaryFileDurable path $ YAML.encode env
 writeMerge :: MonadIO m => FilePath -> Override -> m ()
 writeMerge path newEnv = do
   currentEnv <- decode path
-  writeBinaryFileDurable path $ YAML.encode (currentEnv <> newEnv)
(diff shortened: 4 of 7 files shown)


library/Fission/CLI/Config/Connected.hs
```diff
@@ -28,6 +28,8 @@ import           Fission.CLI.Environment.Types as Environment
 import qualified Fission.CLI.Environment       as Environment
 import qualified Fission.CLI.IPFS.Connect      as Connect
 
+import qualified RIO.NonEmpty as NonEmpty
```
expede, line 31: Nit: for consistency let’s put this at the top of the file (general -> specific)

library/Fission/CLI/Config/Connected/Types.hs
```diff
@@ -43,7 +43,7 @@ data ConnectedConfig = ConnectedConfig
   , processCtx   :: !ProcessContext
   , ipfsPath     :: !IPFS.BinPath
   , ipfsTimeout  :: !IPFS.Timeout
-  , peer         :: !IPFS.Peer
+  , peers        :: !(NonEmpty IPFS.Peer)
```
expede, line 46: Oooooooh

library/Fission/CLI/Environment/Override/Types.hs
```diff
@@ -48,7 +47,7 @@ instance Monoid Override where
 
 instance ToJSON Override where
   toJSON Override {..} = object $ catMaybes
-    [ ("peers"      .=) <$> tailMaybe peers
```
dholms, line 51: I really can't figure out why this would've been written like this in the first place? It seems like extra work :thinking:

library/Fission/CLI/IPFS/Connect.hs
```diff
@@ -7,7 +7,7 @@ module Fission.CLI.IPFS.Connect
 import           Fission.Prelude
 
 import qualified System.Console.ANSI as ANSI
-import           Data.List.NonEmpty as NonEmpty hiding ((<|))
+import           RIO.NonEmpty as NonEmpty hiding ((<|))
```
dholms, line 10: We prefer RIO right?

library/Fission/CLI/Config/Connected.hs
dholms: :+1:

library/Fission/CLI/Environment.hs
```diff
@@ -94,29 +94,30 @@ removeConfigFile = do
...
-      logDebug @Text "Retrieved Peer from API"
       path <- globalEnv
-      Override.writeMerge path $ mempty { peers = NonEmpty.toList nonEmptyPeers }
-      return . Just $ NonEmpty.head nonEmptyPeers
+      let peers = NonEmpty.toList nonEmptyPeers
+      logDebug $ "Retrieved Peers from API, and writing to ~/.fission.yaml: " <> textShow peers
+      Override.writeMerge path $ mempty { peers }
+      return peers
+
+getOrRetrievePeers Environment {peers = peers} = do
```
expede, line 118:
```haskell
getOrRetrievePeers Environment {peers = peers}
```

Can be written without the `=`

```haskell
getOrRetrievePeers Environment {peers}
```

or even as a wildcard:

```haskell
getOrRetrievePeers Environment {..}
```

library/Fission/CLI/Environment.hs
```diff
@@ -94,29 +94,30 @@ removeConfigFile = do
...
+      return []
 
     Right nonEmptyPeers -> do
-      logDebug @Text "Retrieved Peer from API"
       path <- globalEnv
-      Override.writeMerge path $ mempty { peers = NonEmpty.toList nonEmptyPeers }
-      return . Just $ NonEmpty.head nonEmptyPeers
+      let peers = NonEmpty.toList nonEmptyPeers
+      logDebug $ "Retrieved Peers from API, and writing to ~/.fission.yaml: " <> textShow peers
+      Override.writeMerge path $ mempty { peers }
```
expede, line 115: 👍

library/Fission/CLI/Environment/Override.hs
```diff
@@ -83,7 +83,7 @@ write path env = writeBinaryFileDurable path $ YAML.encode env
 writeMerge :: MonadIO m => FilePath -> Override -> m ()
 writeMerge path newEnv = do
   currentEnv <- decode path
-  writeBinaryFileDurable path $ YAML.encode (currentEnv <> newEnv)
+  write path $ currentEnv <> newEnv
```
expede, line 86: Remind me what the difference is? Why don’t we want a durable write here?

library/Fission/CLI/Environment/Override/Types.hs
```diff
@@ -48,7 +47,7 @@ instance Monoid Override where
...
+    [ ("peers"      .=) <$> pure peers
```
expede, line 50: Oh lol yeah that would do it

expede: This looks fairly intentional. I wonder why we had it like that? Just a quick “the server is updated and we’ll do the CLI at some stage”?

library/Fission/CLI/IPFS/Connect.hs
expede, line 10: If you import it qualified (which is how you’re using it AFAICT), you can sidestep conflicting imports

```haskell
import qualified RIO.NonEmpty as NonEmpty
```

library/Fission/CLI/IPFS/Connect.hs
```diff
@@ -28,25 +30,34 @@ swarmConnectWithRetry ::
...
-swarmConnectWithRetry _peer (-1) =
+swarmConnectWithRetry _peers (-1) =
   return . Left $ toException UnableToConnect
 
-swarmConnectWithRetry peer tries =
-  IPFS.Peer.connect peer >>= \case
-    Right _ ->
+swarmConnectWithRetry peers tries = do
+  attempts <- connectToPeers $ NonEmpty.toList peers
+  if any isRight attempts
```
expede, line 41: I almost want to know if any of them fail in an `infoDebug`. Thoughts?

library/Fission/CLI/Environment/Override.hs
dholms: It's still a durable write, The diff doesn't show the full context: 
```haskell

-- | Writes partial environment to path
write :: MonadIO m => FilePath -> Override -> m ()
write path env = writeBinaryFileDurable path $ YAML.encode env

-- | Merge a partial env with an existing env at the path
writeMerge :: MonadIO m => FilePath -> Override -> m ()
writeMerge path newEnv = do
  currentEnv <- decode path
  write path $ currentEnv <> newEnv
```
we just had repeat logic

library/Fission/CLI/IPFS/Connect.hs
```diff
@@ -28,25 +30,34 @@ swarmConnectWithRetry ::
...
-          swarmConnectWithRetry (head peers) (tries - 1)
+          swarmConnectWithRetry retryPeers (tries - 1)
+
+connectToPeers ::
+  ( MonadUnliftIO  m
+  , MonadLocalIPFS m
+  )
+  => [IPFS.Peer]
+  -> m [Either IPFS.Peer.Error ()]
+connectToPeers peers = sequence $ IPFS.Peer.connect <$> peers
```
expede, line 60: No retry?

library/Fission/CLI/Environment/Override/Types.hs
dholms: right I noted that it looks intentional as well. But I can't figure out why we'd do it either....

> the server is updated and we’ll do the CLI at some stage

that seems most likely, but even so you'd think we'd do `head` instead of `tail`

library/Fission/CLI/IPFS/Connect.hs
dholms: the retry is in `swarmConnectWithRetry` which calls this function

library/Fission/CLI/IPFS/Connect.hs
dholms: yeah that makes sense to me :+1: I'll just log the full connection results

library/Fission/CLI/IPFS/Connect.hs
expede:
```haskell
connectToPeers peers = sequence $ connector <$> peers
  where
    connector :: IPFS.Peer -> m ()
    connector peer =  IPFS.Peer.connectWithRetry peer 1
 
-- Or with a monadic fold:

connectToPeers peers = foldM connector (Right ()) peers
  where
    connector :: () -> IPFS.Peer -> m ()
    connector _ peer =  IPFS.Peer.connectWithRetry peer 1
```

We can also attempt these in parallel:

```haskell
sequence $ parMap IPFS.Peer.connect peers
```

expede: (mainly unsure how the parMap will interact with go-ipfs)

library/Fission/CLI/IPFS/Connect.hs
expede: Yes 👍 We probably didn’t realize that it was in there at the time

A lot of this code is _pretty old_ 😅 It’s actually pretty great that it’s mostly just worked all this time (IPFS annoyances aside)

library/Fission/CLI/Environment/Override.hs
expede: Sweet 🍬

library/Fission/CLI/Environment/Override/Types.hs
expede: Well, looks much more correct now 👍

library/Fission/CLI/IPFS/Connect.hs
expede: > the retry is in swarmConnectWithRetry which calls this function

Ah, gotcha 👍

expede:
Looking pretty great overall 👍 Just a handful of tweaks

library/Fission/CLI/IPFS/Connect.hs
dholms: I'm confused why we need to retry in this function? It's only called in `swarmConnectWithRetry` which handles the retry logic. I could make it a helper function on `swarmConnectWithRetry` to make it more clear?

dholms: `parMap` makes sense to me though as long as it works with go-ipfs (which I imagine it should)

expede: Well, specifically will the daemon try in parallel since there seems to be a lot of singl-threadedness causing headaches. We can probably test this by opening a couple tabs, and asking for peers that don’t exist, and seeing how long between failures (close means actually parallel, spread out means one timed out then the next)

expede: Though really adding that one function and testing in the CLI is also pretty lightweight :P

dholms: just tried it out. it seems that they run in parallel

dholms: getting compiler errors with `parMap` that I'm not sure how to manage:
```haskell
  attempts <- sequence $ parMap IPFS.Peer.connect (NonEmpty.toList peers)
```
```
    • Couldn't match type ‘Either Network.IPFS.Peer.Error.Error ()’
                     with ‘IPFS.Peer’
      Expected type: Control.Parallel.Strategies.Strategy
                       (Either Network.IPFS.Peer.Error.Error ())
        Actual type: IPFS.Peer
                     -> Control.Parallel.Strategies.Eval
                          (Either Network.IPFS.Peer.Error.Error ())
```
will continue to mess around :thinking:

library/Fission/CLI/Environment.hs
expede: Resolved?

dholms: I.... I swear I fixed this. like have a distinct memory of doing it. spooky :ghost: 
ah well thanks for double checking me :sweat_smile:

Issue #371, opened 2020-08-04
Global ignore file


bmann:
The global `~/.fission.yaml` file no longer has an ignore section.

I think these are the base defaults:
https://github.com/fission-suite/fission/blob/0bee1438bf9aba34adcc89034607dcaba225c3f4/library/Fission/CLI/Environment.hs#L121-L126

I think this needs to be connected to where peer IDs are fetched?

PR #373, opened 2020-08-04
Init env with peers and default ignores

dholms:
## Problem
The global `~/.fission.yaml` file no longer has an ignore section and is not created on `setup`

## Solution
Run `Env.init` on a successful `fission setup`

Closes https://github.com/fission-suite/fission/issues/371


expede:
Sure, makes sense

PR #374, opened 2020-08-04
Only print connected error once

dholms:
## Problem
When a user runs a command and is not logged in we print the error twice:
```
🚫 Not logged in yet! Try running `fission setup`
🚫 Not logged in yet! Try running `fission setup`
```
This is base the error message is printed in `runConnected` & we attempt to print the error message again in another `runConnected`

## Solution
Remove the extra `put'`

### DIFF
--- a/library/Fission/CLI.hs
+++ b/library/Fission/CLI.hs
@@ -38,14 +38,10 @@ cli baseCfg = liftIO do
 
     runConnected_ :: Command FissionConnected input () -> Command.Leaf
     runConnected_ =
-      runWith \actn -> do
-        result <- runConnected baseCfg do
+      runWith \actn ->
+        void $ runConnected baseCfg do
           logDebug @Text "Setting up connected"
           actn
-
-        case result of
-          Right _  -> return ()
-          Left err -> void . runConnected baseCfg $ put' err
 
 summary :: String
 summary = "CLI to interact with Fission services"
--- a/package.yaml
+++ b/package.yaml
@@ -1,5 +1,5 @@
 name: fission
-version: '2.6.6'
+version: '2.6.7'
 category: API
 author:
   - Brooklyn Zelenka


library/Fission/CLI.hs
```diff
@@ -38,15 +38,11 @@ cli baseCfg = liftIO do
 
     runConnected_ :: Command FissionConnected input () -> Command.Leaf
     runConnected_ =
-      runWith \actn -> do
-        result <- runConnected baseCfg do
+      runWith \actn ->
+        void $ runConnected baseCfg do
```
expede, line 42: Oh, the error handling moved inside. Are we sure that we want to do the logging inside the action, and not out here?

expede: i.e. keep the error logging together

expede: (opposite of this)

expede: (also to be clear — not saying “not this”, just “have we considered both approaches?“)

dholms: > Oh, the error handling moved inside
It actually already was inside. `put'` prints a generic "Something went wrong" error message (but it never actually gets printed cause it's run in `runConnected` which throws the error again)

I can see both arguments. My thought was that we already had all of the logs inline, so there seemed to be some desire for that style. Plus it gives a bit more flexibility with error handling. For instance `CLI.Error.notConnected` prints a diff message based on if the user has upgraded their CLI from an old pre-key version (granted, we could handle this with error types as well)

So yeah I can really go either way :man_shrugging:

expede: Yup, makes sense. It’s sometimes nice to propagate the error up to be able to handle or interpret it (for example, to know which command you’re in), but I can see both arguments. These commands are so short lived and concrete that I can see bth arguments. Okay, cool let’s merge this!

PR #396, opened 2020-08-31
Set DNSLink *after* pinning

expede:
# Existing Code

1. Update database record
2. Set DNSLink
3. Pin

# Problem

Well, in practice this is completely backwards! We have situations where the files aren’t synced, the user closes their window / stops transfer. Since the DNSLink is already set, there’s no way to get the new data.

# Updated Code

1. Pin
2. Set DNSLink
3. Update database record

### DIFF
--- a/fission-core/library/Fission/Types.hs
+++ b/fission-core/library/Fission/Types.hs
@@ -448,25 +448,23 @@ instance User.Modifier Fission where
             return $ Error.openLeft err
 
           Right size -> do
-            _ <- runDB $ User.setDataDB userId newCID size now
-
-            userDataDomain <- asks userRootDomain
-            zoneID         <- asks userZoneID
-
-            let
-              url = URL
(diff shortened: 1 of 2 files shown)


expede:
Open question — do we want to do this with app as well? Kind of nice that they are live “instantly”, and devs will generally wait for stuff to be live. Will be especially good once there’s streaming % feedback.

Apps tend to be fairly small, though, so maybe also an overoptimization? 👍

bmann:
I think sharing the “local” hash solves the instant bit.

expede:
> I think sharing the “local” hash solves the instant bit.

Yeah absolutely for user _data_. Because apps get served over the gateway, that strategy doesn’t actually work for them. After reflecting on it a bit, data integrity is more important than raw speed, our DNS is actually _ surprisingly quick_ most of the time, and people can fall back to the CID if running IPFS locally. Making the change 👍

PR #397, opened 2020-09-03
Global config & Fission-managed IPFS

expede:
This is a dependency for doing CLI linking. Doing it as a separate task to keep PRs from always getting huge 😅 

Since we don’t seem to be shipping files anytime soon, and we don’t have to do our own manual IPFS ignore anymore, removing recursive lookups and manual recursive adds.

All file access is atomic and durable (hard to accidentally corrupt these files while writing).

This PR adds:

* Global directory: `~/.config/fission`
  * Config at `~/.config/fission/config.yaml`
  * Key at `~/.config/fission/keys/machine_id.ed25519`
  * Cache server DID
* App config at `./fission.yaml`
  * Does not contain anything that can’t be uploaded — all public ok
  * Includes the app URL (by popular request)
* IPFS
  * Use CIDv1
  * Download/install/setup `go-ipfs` at `~/.config/fission/bin/ipfs`
  * Data repo at `~/.config/fission/ipfs/[repo]`
  * Managed daemon at a port that won’t conflict with other system IPFS
  * Safe daemon startup and teardown
  * Use `—ignore-rules-path` instead of building our own DAG
* Use CLI’s own version number (separate from the server’s)
* Remove unused functions from keystore module


expede:
Rebasing

fission-cli/library/Fission/CLI/App/Environment.hs
```diff
@@ -0,0 +1,70 @@
...
+absPath = do
+  pwd <- getCurrentDirectory
+  return $ pwd </> relPath
+
+relPath :: FilePath
+relPath = "fission.yaml"
+
+ignoreDefault :: [Text]
+ignoreDefault =
+  [ Text.pack relPath
```
expede, line 67: Nothing harmful about uploading this these days. I dunno. Remove?

PR #401, opened 2020-09-09
Spelling correction

bmann:
verififcation --> verification


expede (approved):
NOTE TO SELF: Boris is off, so bump version and merge

PR #402, opened 2020-09-17
GET User Data Route

dholms:
## Problem
DNS takes a while to propagate, and specifically only initial lookups, it can take > 15 min

## Solution
Add a route to our server that checks the value of a user's dataroot directly in DNS


fission-core/library/Fission/AWS/Route53.hs
```diff
@@ -1,3 +1,25 @@
...
+
+import           Fission.Prelude
+import           RIO
+import           Network.AWS.Route53
+
+verifyFirstResource
+  :: ListResourceRecordSetsResponse -> Text -> Maybe ResourceRecordSet
+verifyFirstResource resp name = case headMaybe recordSets of
+  Nothing  -> Nothing
+  Just rrs -> if (rrs ^. rrsName) == (name <> ".") then Just rrs else Nothing
```
dholms, line 18: Good news: I got stylish-haskell working
Bad news: it's a little overzealous some of the time

For instance, I'd like to split this `if` statement into multiple lines, but it always condenses it. Is this something I should fix locally or in the `.stylish-haskell.yaml`?

expede: @dholms which version of stylish-haskell are you running?

fission-core/library/Fission/AWS/Validate.hs
```diff
@@ -0,0 +1,26 @@
...
+import           Network.AWS.Route53
+
+import           Fission.Web.Error
+
+
+validate :: HasStatus a => a -> Either ServerError a
+validate res = if status >= 300 then Left (toServerError status) else Right res
+  where status = getStatus res
+
+class HasStatus a where
```
expede, line 19: A couple recommendations

#  Getting a status code is actually pretty widely applicable
 
* e.g. Client requests also have status codes
* Pull it into its own .Class` module

# Optionally: use `Has`

For example:

```haskell
instance Has StatusCode ChangeResourceRecordSetResponse where
  hasLens = crrsrsResponseStatus
```

# `getStatus` is a bit general

Other libraries make the distinction between a status _message_ and a status _code_

https://www.stackage.org/haddock/lts-16.14/http-types-0.12.3/Network-HTTP-Types-Status.html#t:Status

Maybe name it `StatusCode`?

# Optional: newtype

If you want to get super pedantic, can also wrap in a `StatusCode` newtype wrapper, but that may actually be more hassle than its worth for your use case?

fission-core/library/Fission/AWS/Route53.hs
dholms: 0.10

fission-core/library/Fission/DataRoot.hs
```diff
@@ -0,0 +1,6 @@
+module Fission.DataRoot
```
expede, line 1: I like it!

A couple things:

`where` flush left? Everywhere else we have it next to the closing `)`.

I like the impulse to break things into classes 🙌 🎉

`DataRoot` is a bit generic here, no? Is it the data root _in general_ on some IPLD, or over on DNS? Perhaps `MonadWNFS`?

```haskell
module Fission.WebNative.FileSystem.Class where

-- ...

class Monad m => MonadWNFS m where
  getDataRootForUser :: m (Either ServerError CID)
```

fission-core/library/Fission/DataRoot/Class.hs
```diff
@@ -0,0 +1,14 @@
...
+
+import           Fission.Prelude
+import           Servant
+
+import           Fission.AWS.Route53
+import           Fission.User.Username.Types
+import           Network.IPFS.CID.Types
+
+class MonadRoute53 m => MonadDataRoot m where
+  get :: Username -> m (Either ServerError CID)
```
expede, line 14: `ServerError`, not `ClientError`?

expede: Super minor tip, no need to change:

The class definition doesn’t necessarily need to have `MonadRoute53` on it! You _can_ rely on Route53 in the `instance MonadDataRoot Fission`, however, since `Fission` has that defined.

By not requiring `MonadRoute53`, we can later move this lookup to IPNS, the database, or similar. We also don’t need to define `MonadRoute53` if we just want to test the data root in a testing scenario.

fission-core/library/Fission/AWS/Validate.hs
dholms: nice I like it :+1: 

I ended up doing the `newtype` but did not do `Data.Has` since it seems like we've been kind of moving away from that pattern

fission-core/library/Fission/DataRoot.hs
dholms: > where flush left? Everywhere else we have it next to the closing ).

that was a formatting tool thing. working on it :ok_hand: I'll override it for now

dholms: hmmmm yeah I wasn't totally happy with `MonadDataRoot`, I like `MonadWNFS` :+1:

fission-core/library/Fission/DataRoot/Class.hs
dholms: > The class definition doesn’t necessarily need to have MonadRoute53 on it! You can rely on Route53 in the instance MonadDataRoot Fission, however, since Fission has that defined.

Nice! Yeah that's great!

dholms: I did `ServerError` here because that is what MonadRoute53 returns in it's other methods (`set` & `clear`)

I can see that it would make sense for those to be Client Errors until they get to the route logic which then returns them as a `ServerError`

fission-core/library/Fission/AWS/Validate.hs
expede: We are not moving away from `Has` altogether

expede: `Has` is more convenient than writing specialized `HasHTTPManager`, `HasFoo`, `HasBar` classes

expede: It may be worth chatting about the distinction here when you’re done driving, if you’re interested

fission-core/library/Fission/DataRoot/Class.hs
expede: Ah, okay, that’s our doing in `Fission.AWS.validate`. We use `toServerError` (a creation of our own) to turn the AWS error into something we can pass back to the user.

Probably should do this later in the stack. When we move to MonadRescue, that could be a good idea.

But yeah, reads a bit strange right now but it does the correct things, so 🤷‍♀️

fission-core/library/Fission/AWS/Validate.hs
dholms: Yeah definitely interested!

dholms:
All tests passing:
Screenshot from 2020-09-21 14-50-52
Screenshot from 2020-09-21 14-51-09

PR #406, opened 2020-09-22
CLI Linking OLD VERSION HERE TO MAKE PORTING CHANGES EASIER

expede:
# 1. System Setup
* [x] Start pubsub with daemon
* [x] Add RSA exchange key to `~/.config/fission/keys/exchange.rsa2048`
* ~~[ ] Ensure that RSA-OAEP DIDs work (as well as PCKS15)~~ It’s the same
* [x] Add exchange key DID to user account DNS on signup
* [x] Enable WS in IPFS node

`~/.ipfs/config` needs these set:

```js
  "Swarm": {
    "AddrFilters": null,
    "ConnMgr": {
      "GracePeriod": "20s",
      "HighWater": 900,
      "LowWater": 600,
      "Type": "basic"
    },
    "DisableBandwidthMetrics": false,
    "DisableNatPortMap": false,
    "DisableRelay": false,
    "EnableRelayHop": true
  }
```

However my IPFS config has different values.

# 2. Pubsub Session

* [ ] ~~Connect IPFS daemon to WebRTC signal server as needed~~ Not actually possible "because GoLang" 🤷‍♀️ 
* [ ] Bootstrap secure pubsub session (handshake)
* [ ] Requestor send challenge code from screen display

# 3. Exchange

* [ ] Provider — send UCAN
* [ ] Provider — send read key
* [ ] Requestor — accept & store UCAN in `~/.config/fission/auth/${did}/ucan/${hash}.ucan.jose` [???]
* [ ] Requestor — accept & store read key in `~/.config/fission/auth/${did}/shared_with_me/${cid}.aes256` [???]

# 4. Check

* [ ] Can push changes to a shared app

---

# Notes

* Added `MonadKeyStore` (lots of confusing duplication without it)
* Broke out `fission-web-client`


fission-web-client/test/Test/Fission/CLI/Web/Client/Error.hs
```diff
@@ -0,0 +1,47 @@
+module Test.Fission.CLI.Web.Client.Error (tests) where
```
expede, line 1: Note to self: remove these from the CLI tests

fission-core/library/Fission/Internal/AES256/Types.hs
```diff
@@ -0,0 +1,3 @@
+module Fission.Internal.AES256.Types where
+
+import           Crypto.Cipher.AES (AES256)
```
expede, line 3: Delete this module

PR #410, opened 2020-10-16
Update UCAN to 0.5

expede:
This PR is pretty enormous. Let's talk about _why._ The TL;DR is that we make things easy for senders (convenient easy for users), but harder for validators (service providers need to handle the newly introduced detail).

* Moving to UCAN 0.5 means that each UCAN layer can now both "bow tie" and inherit. This means that validation is a bit trickier, but not overwhelming
* We now support more than super user rights (actually support rights delegation, not just "all")
* Can dynamically delegate all rights in scope, so checking is often left to runtime
* Can gather rights from multiple senders/roots, so can't infer who you're acting as directly — our usual "authorization not identity" — so that needs to be reflected in the server

---

Switching back to linking with UCAN 0.3.1

To know where to pick up:

* Structurally UCAN is done
* Need to tweak the current 0.4 to 0.5 to reflect changes that make the FE implementation easier and bump semver
* Auth session is all hooked up, but need to run it in the relevant handlers
* We can no long imply a single root user, so switch from root proof userId to the `Access resource` and username
  * Means that the web API needs to include the relevant user when changing an account
  * Apps & domains will probably stay roughly the same
  * Updating data root will.... need some thinking on how exactly to target that, since you may need to update multiple data roots atomically


expede:
Yeah, this is now firmly in "historical reasons". Closing.

PR #412, opened 2020-11-07
Downcase usernames

expede:
## Problem

Username creation is not currently case sensitive 😅 So, `james` and `James` are considered distinct. Both confusing, and dangerous for subdomains & DNSLink.

## Solution

Enforce the following invariants via the smart constructor `mkUsername`:

* Lowercase
* Not in deny list
* URL safe
* Follows subdomain rules


default.nix
```diff
@@ -0,0 +1,38 @@
...
+  # (1)
+  { pkgs ? pkgsNix.native }:
+    # (2)
+    pkgs.haskell-nix.project {
+      src = pkgs.haskell-nix.haskellLib.cleanGit {
+        name = "fission-cli";
+        src = ./. ;
+      };
+    }
+>>>>>>> 91c5d234db3ec882ec14b08d4d95d3d7397d9b06
```
expede, line 38: LOL okay will fix

expede:
Oopsie — rebasing

PR #428, opened 2020-12-02
Notify user of new cli versions

### LINKED ISSUE #330: Check for current version of CLI and message user
via @forgondolin, a suggestion on updating the CLI to the newest version.

We'll want to start by checking for the current released version in this Github repo, and message the user.

This could be as simple as `You have fission-cli v1.2.1, the current version is v1.2.3. Run brew upgrade to get the newest version`

This could get run whenever a "major" command like fission up is run, which is connecting to the Internet anyway. Might also include a mini-changelog, eg. `The new version includes faster uploads, domain name buying, and better docs`.

There are related issues on Linux and Windows/WSL about how the user has installed the CLI (via direct download, apt packages, etc), and how we can help people stay update.

walkah:
check latest release version on github and notify user if they should upgrade.

## Closing issues

Fixes #330

### DIFF
--- a/fission-cli/library/Fission/CLI/Release.hs
+++ b/fission-cli/library/Fission/CLI/Release.hs
@@ -1,24 +1,40 @@
 module Fission.CLI.Release (checkLatestRelease) where
 
+import qualified System.Console.ANSI                    as ANSI
+
 import           Fission.Prelude
+
+import qualified Fission.CLI.Meta                       as Meta
+import           Fission.CLI.Display.Text
+import qualified Fission.Internal.UTF8                  as UTF8
+import           Fission.SemVer.Types
 
 import           GitHub (github')
 import qualified GitHub
-import           GitHub.Data.Releases                           as Releases
+import           GitHub.Data.Releases                   as Releases
 
-checkLatestVersion ::
-  ( MonadLogger m
-  , MonadIO     m
+checkLatestRelease ::
+  ( MonadLogger  m
+  , MonadIO      m
+  , MonadCleanup m
   )
   => m ()
 checkLatestRelease = do
   -- check "last checked" timestamp to see if we need to check again
-  logInfo @Text "Checking for newer versions..."
+  logDebug @Text "Checking for newer versions..."
 
   possibleVersion <- liftIO $ github' GitHub.latestReleaseR "fission-suite" "fission"
   case possibleVersion of
     Left  _ -> return () -- just ignore errors here.
-    Right latestVersion -> do
-      -- check latest version against current version from Meta.package
-      logInfo $ "latest version is: " <> Releases.releaseTagName latestVersion
-      return ()
+    Right latestRelease -> do
+      let
+        currentVersion = getSemVer $ Meta.version =<< Meta.package
+        latestVersion  = getSemVer $ Just (Releases.releaseTagName latestRelease)
+      if currentVersion < latestVersion
+        then colourized [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Yellow] do
+          UTF8.putText $ "⚠️  A new version of Fission CLI is available: "
+          UTF8.putText $ Releases.releaseTagName latestRelease <> " (aka '" <> Releases.releaseName latestRelease <> "')\n"
+        else return ()
+
+getSemVer :: Maybe Text -> Maybe SemVer
+getSemVer maybeVersion = decode $ encode maybeVersion
--- a/fission-cli/library/Fission/CLI/Release.hs
+++ b/fission-cli/library/Fission/CLI/Release.hs
@@ -1,5 +1,6 @@
 module Fission.CLI.Release (checkLatestRelease) where
 
+import           Data.Versions
 import qualified System.Console.ANSI                    as ANSI
 
 import           Fission.Prelude
@@ -7,7 +8,6 @@ import           Fission.Prelude
 import qualified Fission.CLI.Meta                       as Meta
 import           Fission.CLI.Display.Text
 import qualified Fission.Internal.UTF8                  as UTF8
-import           Fission.SemVer.Types
 
 import           GitHub (github')
 import qualified GitHub
@@ -28,13 +28,19 @@ checkLatestRelease = do
     Left  _ -> return () -- just ignore errors here.
     Right latestRelease -> do
       let
-        currentVersion = getSemVer $ Meta.version =<< Meta.package
-        latestVersion  = getSemVer $ Just (Releases.releaseTagName latestRelease)
+        currentVersion = getVersion $ Meta.version =<< Meta.package
+        latestVersion  = getVersion $ Just (Releases.releaseTagName latestRelease)
       if currentVersion < latestVersion
         then colourized [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Yellow] do
           UTF8.putText $ "⚠️  A new version of Fission CLI is available: "
           UTF8.putText $ Releases.releaseTagName latestRelease <> " (aka '" <> Releases.releaseName latestRelease <> "')\n"
-        else return ()
+        else return () -- no new version
 
-getSemVer :: Maybe Text -> Maybe SemVer
-getSemVer maybeVersion = decode $ encode maybeVersion
+getVersion :: Maybe Text -> Maybe Versioning
+getVersion maybeVersion =
+  case maybeVersion of
+    Nothing -> Nothing
+    Just v  ->
+      case versioning v of
+        Left _ -> Nothing
+        Right version -> return version
--- a/fission-cli/library/Fission/CLI/Release.hs
+++ b/fission-cli/library/Fission/CLI/Release.hs
@@ -27,21 +27,23 @@ checkLatestRelease ::
   , MonadEnvironment m
   , m `Raises` YAML.ParseException
   , m `Raises` NotFound FilePath
-  , Display (OpenUnion (Errors m))
+  , Show (OpenUnion (Errors m))
   )
   => m ()
 checkLatestRelease = do
   attempt Env.get >>= \case
     Left  err -> CLI.Error.put err "Unable to parse config"
     Right Env {updateChecked} -> do
-      now            <- currentTime
-      nextUpdateTime <- addUTCTime 86400 updateChecked
+      now <- currentTime
+      let
+        nextUpdateTime = addUTCTime 86400 updateChecked
       unless (now < nextUpdateTime) do
         logDebug @Text "Checking for newer versions..."
         possibleVersion <- liftIO $ github' GitHub.latestReleaseR "fission-suite" "fission"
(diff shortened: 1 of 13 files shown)


fission-cli/library/Fission/CLI/Release.hs
```diff
@@ -0,0 +1,24 @@
...
+  -- check "last checked" timestamp to see if we need to check again
+  logInfo @Text "Checking for newer versions..."
+
+  possibleVersion <- liftIO $ github' GitHub.latestReleaseR "fission-suite" "fission"
+  case possibleVersion of
+    Left  _ -> return () -- just ignore errors here.
+    Right latestVersion -> do
+      -- check latest version against current version from Meta.package
+      logInfo $ "latest version is: " <> Releases.releaseTagName latestVersion
+      return ()
```
expede, line 24: NO NEED TO CHANGE, but fun fact: 

```haskell
logInfo :: MonadLogger m => Text -> m ()
```

So the resulting type of `logInfo` is `m ()`, which is the same as `return ()`. You can remove line 24 and it'll be the same.

I realize that the current code is WIP and line 24 is probably from when you wrote the initial branching. Also nothing wrong with leaving it in there (it'll have zero runtime cost). Just using it to highlight how to think in Haskell :)

expede: Oh, also before I forget: possibly another helpful tip...

If you need to just stub out a function or value as a "I'll look at that later", using the value `undefined` will fill literally any type. If the program executes that instruction, it'll crash, but helpful when developing as a placeholder.

expede:
So, I'm pretty sure that the tests will be failing (at minimum) the unit tests that have real world UCANs. You'll need to adjust the serialization of UCANs. There's two ways of doing this:

# 1. Use a `Maybe Word8` field fo the "tick" field

This will give you the ability to control how many digits a JSON encoded SemVer has. A `tick = Nothing` would have 3, `tick = Just 0` would have 4. You'd define that in the `ToJSON` and `FromJSON` instances. Something like:

```haskell
data SemVer = SemVer {major :: !Word8, minor :: !Word8, patch :: !Word8, tick :: !(Maybe Word8)}

instance ToJSON SemVer where
  toJSON SemVer {..} = JSON.String $ common <> ending
    where
      common = 
        intercalate "." $ fmap textDisplay [major, minor, patch]

      ending =
        case tick of
          Nothing -> ""
          Just x    -> "." <> textDisplay x
```

# 2. Edit the `JWT.Header` JSON

In the `instance ToJSON Header`, override the SemVer serialization with something like this:

```haskell
instance ToJSON Header where                                                                                                               
  toJSON Header {..} = object                                   
    [ "typ" .= typ                                                          
    , "alg" .= alg                                                           
    , "cty" .= cty                                                         
    , "uav" .= version
    ]    
  where
    version = intercalate "." $ fmap textDisplay
      [ major uav
      , minor uav
      , patch uav 
      ]
 ```

**Comment by walkah** [comment]
@expede I opted for option 3 which was to use the versioning package - which can handle all kinds of version strings and seems to be less generally disruptive to the rest of the codebase.

(we could potentially remove our own custom SemVer implementation in favour of this one too?)

**Comment by expede** [comment]
> @expede I opted for option 3 which was to use the versioning package - which can handle all kinds of version strings and seems to be less generally disruptive to the rest of the codebase.

Sounds great 👍 

> (we could potentially remove our own custom SemVer implementation in favour of this one too?)

Yes, that makes sense. As long as we can get a 3-number semver out of the package, all of the tests should pass.

File: shell.nix [hunk 2]
```diff
@@ -19,6 +19,7 @@ pkgs.mkShell {
     hinotify
 
     # Haskell
+    unstable.ghc
```
expede, line 22: Interesting. Why?

walkah: > Interesting. Why?

oh whoops. I was playing with `dante` (for emacs) which tries to launch`ghci` in a pure nix shell.  that doesn't need to be here.

expede: Yeah, I like `dante` (and Intero before it).

I tend to use [`ghcid`](https://github.com/ndmitchell/ghcid) these days, which continually checks your code and displays compile errors in its own window (which is _amazing_ when refactoring across modules). It's getting stuck on the Template Haskell that grabs the version. Works when compiling, but not in ghcid 🤷‍♀️ The solution is splitting that module out into the executable instead of the library, but just haven't done that yet.

fission-cli/library/Fission/CLI/Release.hs
```diff
@@ -0,0 +1,63 @@
...
+  , MonadCleanup     m
+  , MonadEnvironment m
+  , m `Raises` YAML.ParseException
+  , m `Raises` NotFound FilePath
+  , Show (OpenUnion (Errors m))
+  )
+  => m ()
+checkLatestRelease = do
+  attempt Env.get >>= \case
+    Left  err -> CLI.Error.put err "Unable to parse config"
```
expede, line 35: Super minor style OPTIONAL thing that I find useful but is nonstandard: If any branch in a `case` drops to the next line, have them all drop to the next line and place a blank line between. I find it makes visually parsing each branch a bit easier, but also maybe just me 🤷‍♀️

fission-cli/library/Fission/CLI/Release.hs
```diff
@@ -0,0 +1,63 @@
...
+    Right Env {updateChecked} -> do
+      now <- currentTime
+      let
+        nextUpdateTime = addUTCTime 86400 updateChecked
+      unless (now < nextUpdateTime) do
+        logDebug @Text "Checking for newer versions..."
+        possibleVersion <- liftIO $ github' GitHub.latestReleaseR "fission-suite" "fission"
+        case possibleVersion of
```
expede, line 43: _Totally fine as-is._

However, if you want to cut a small amount of boilerplate, you can switch to `>>=` here to "handle the result directly". You tend to get an `Either` back when talking to external services, because there's any number of things that can go wrong in that call, so this is a very common pattern.

```haskell
liftIO (github' GitHub.latestRelaseR "fission-suite" "fission") >>= \case
  Left _ ->
    return ()

  Right latestRelease -> do
    -- ...
```

fission-cli/library/Fission/CLI/Release.hs
```diff
@@ -0,0 +1,63 @@
...
+          Left  _ -> return () -- just ignore errors here.
```
expede, line 44: Let's also log the failure so we can see it with `--verbose`

```haskell
Left err ->
  logDebug $ show err

Right latestRelease -> do
  -- ...

File: fission-cli/library/Fission/CLI/Release.hs [hunk 6]
```diff
@@ -0,0 +1,63 @@
...
+          Right latestRelease -> do
+            Env.update \env -> env {updateChecked = now}
+            let
+              currentVersion = getVersion $ Meta.version =<< Meta.package
+              latestVersion  = getVersion $ Just (Releases.releaseTagName latestRelease)
+            if currentVersion < latestVersion
+              then colourized [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Yellow] do
+                UTF8.putText $ "⚠️  A new version of Fission CLI is available: "
+                UTF8.putText $ Releases.releaseTagName latestRelease <> " (aka '" <> Releases.releaseName latestRelease <> "')\n"
+              else return () -- no new version
```
expede, line 54: Also totally fine to have the manual branching here, but you can also use `when :: Monad m => Bool -> m a -> m a` which does exactly the same thing:

```haskell
when currentVersion < latestVersion do
  colourized [ANSI.SetColor {- ...-}] do
    UTF8.putText "text text text"
    -- ...
```

fission-cli/library/Fission/CLI/Release.hs
```diff
@@ -0,0 +1,63 @@
...
+        case possibleVersion of
+          Left  _ -> return () -- just ignore errors here.
+          Right latestRelease -> do
+            Env.update \env -> env {updateChecked = now}
+            let
+              currentVersion = getVersion $ Meta.version =<< Meta.package
+              latestVersion  = getVersion $ Just (Releases.releaseTagName latestRelease)
+            if currentVersion < latestVersion
+              then colourized [ANSI.SetColor ANSI.Foreground ANSI.Vivid ANSI.Yellow] do
+                UTF8.putText $ "⚠️  A new version of Fission CLI is available: "
```
expede, line 52: Again, no need to change per se, but this line doesn't need the `$` (since it's a single argument to the `putText` function)

fission-cli/library/Fission/CLI/Release.hs
```diff
@@ -0,0 +1,63 @@
...
+                UTF8.putText $ Releases.releaseTagName latestRelease <> " (aka '" <> Releases.releaseName latestRelease <> "')\n"
```
expede, line 53: There's also a `putTextLn` function that includes the newline, which is sometimes nicer than typing out the `\n`

fission-cli/library/Fission/CLI/Release.hs
```diff
@@ -0,0 +1,63 @@
...
+              else return () -- no new version
+
+getVersion :: Maybe Text -> Maybe Versioning
+getVersion maybeVersion =
+  case maybeVersion of
+    Nothing -> Nothing
+    Just v  ->
+      case versioning v of
+        Left _ -> Nothing
+        Right v' -> return v'
```
expede, line 63: Haskell, much like Elixir, has multiple function heads. You can write the above a few ways:

```haskell
getVersion Nothing = Nothing
getVersion (Just verTxt) =
  case versioning verTxt of
    Left   _ -> Nothing
    Right v -> Just v

---

getVersion = \case
  Nothing -> 
    Nothing

  Just verTxt ->
    case versioning verTxt of
      Left   _ -> Nothing
      Right v' -> return v'

---

-- This one is more of a party trick, though you will see it in the wild:

getVersion maybeVersion = do
  verTxt <- maybeVersion
  either (\_ -> Nothing) pure (versioning verTxt)
```

stack.yaml
```diff
@@ -6,7 +6,9 @@ packages:
 
 extra-deps:
   - alex-3.2.5
+  - binary-instances-1.0.1
```
expede, line 9: Which instances do we need from this package?

expede (approved):
A couple style/best practices, but overall this looks fantastic 🥳

fission-cli/package.yaml
```diff
@@ -107,6 +107,7 @@ dependencies:
   - bytestring
   - memory
   - utf8-string
```
expede, line 109: Don't forget to bump the version number at the top of this file!

expede: IRONIC!

stack.yaml
walkah: the `github` package claimed to need this 🤷

expede: Ah, `github` is coming from Hackage. Gotcha 👍

PR #431, opened 2021-01-02
Split libraries (no CLI dependencies) & GHC 8.10.3 (Big Sur)

expede:
# Changes

* Upgrade to GHC 8.10.3 via Stackage Nightly
  * Works on Big Sur
  * Works on M1 under automatic x86 emulation
* Globally enable the `StrictData` GHC extension
  * This means that all fields have implied `!`s on them
  * If you actually want a lazy field, use `~` in front
    * e.g. `data Foo = Foo { lazy :: ~Int }

# macOS `xz` Dependency via Brew

On Mac, we have to still build with `--no-nix`, which kinda sucks. The problem is `xz`/ libzma. For whatever reason, the version installed with Nix doesn't contain the C headers, but the Brew one does. In theory, once this lib is built the first time, we can use Nix. Will investigate later.

# Works on My Machine

Screen Shot 2021-01-02 at 1 47 36 PM

Screen Shot 2021-01-02 at 1 51 41 PM
Screen Shot 2021-01-02 at 1 51 45 PM

# Note

I get this message when compiling. It's complaining about an uppercase module name, which is likely due to macOS's file system (AFS) being case-insensitive. Will see if a clean build fixes this, but everything does appear to work.

Screen Shot 2021-01-02 at 1 47 40 PM


pbcopy
```diff
@@ -0,0 +1 @@
+ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIM5RGNvxkIOd7lbaCUIe4m2fOZeO0tlTvJXzMZZdtBfo hello@brooklynzelenka.com
```
walkah, line 1: 🙈

stack.yaml
```diff
@@ -1,11 +1,19 @@
-resolver: lts-16.23
+resolver: nightly-2021-01-02
```
walkah, line 1: ✨

pbcopy
expede: Public key, but yeah 😛

fission-cli/library/Fission/CLI/Parser/Command.hs
```diff
@@ -16,11 +16,12 @@ import qualified Fission.CLI.Parser.Command.Version as Version
...
 parser :: Parser Options
 parser = do
-  cmd                      <- shortcuts <|> subCommands <|> version
-  Remote.RemoteConfig {..} <- Remote.parser
+  cmd    <- shortcuts <|> subCommands <|> version
+  config <- Remote.parser
 
-  pure Options { fissionDID = mayDID
-               , fissionURL = target
+  -- NOTE GHC 8.10.3 doesn't like the wildcard. Use wildcards again later.
```
expede, line 22: Annoying, but whatever

fission-cli/library/Fission/CLI/Parser/Config/Remote/Types.hs
```diff
@@ -7,6 +7,6 @@ import           Fission.Prelude
 import           Fission.User.DID.Types
 
 data RemoteConfig = RemoteConfig
-  { target :: !BaseUrl     -- ^ Which remote peer / server to use
-  , mayDID :: !(Maybe DID) -- ^ Cached remote DID
+  { target :: BaseUrl   -- ^ Which remote peer / server to use
+  , mayDID :: Maybe DID -- ^ Cached remote DID
```
expede, line 11: Ah, right, added `StrictData`. Going to run through and remove the `!`s everywhere

walkah (approved):
Builds for me! (With split merged)

PR #435, opened 2021-01-15
New route for resending email verification

matheus23:
This is just so that it's up here.

At the moment there's still a bug in here.
When resending a verification email, I create a new challenge. So then there are multiple challenges associated with a user. But verification only `selectFirst`s the challenges.

So two solutions:
* Fetch all challenges when verifying, so `selectAll` or sth like that (possibly less secure)
* Send the same challenge when resending email verification as the one associated.

By the way - is having multiple challenges associated with a user intended? If not, couldn't it become a field of User?

---

First dip into server code yay :tada:

### DIFF
--- a/fission-web-api/library/Fission/Web/API/User/Email/Resend/Types.hs
+++ b/fission-web-api/library/Fission/Web/API/User/Email/Resend/Types.hs
@@ -1,0 +1,14 @@
+module Fission.Web.API.User.Email.Resend.Types (Resend) where
+
+import           Fission.Web.API.Prelude
+
+import qualified Fission.Web.API.Auth.Types as Auth
+
+type Resend = "resend" :> Check
+
+type Check
+  =  Summary "Resend verification email"
+  :> Description "Send a verification email to currently authenticated user."
+  --
+  :> Auth.HigherOrder
+  :> PostNoContent
--- a/fission-web-api/library/Fission/Web/API/User/Email/Types.hs
+++ b/fission-web-api/library/Fission/Web/API/User/Email/Types.hs
@@ -4,4 +4,10 @@ import           Fission.Web.API.Prelude
 
 import           Fission.Web.API.User.Email.Verify.Types
 
-type Email = "email" :> Verify
+import           Fission.Web.API.User.Email.Resend.Types
+
+type Email = "email" :> API
+
+type API
+  =    Verify
+  :<|> Resend
--- a/fission-web-server/library/Fission/Web/Server/Challenge.hs
+++ b/fission-web-server/library/Fission/Web/Server/Challenge.hs
@@ -1,23 +1,25 @@
 module Fission.Web.Server.Challenge
   ( module Fission.Web.Server.Challenge.Creator.Class
   , module Fission.Web.Server.Challenge.Verifier.Class
+  , module Fission.Web.Server.Challenge.Retriever.Class
   , verificationLink
   ) where
 
-import           Servant                                     hiding (route)
+import           Servant                                      hiding (route)
 
 import           Fission.Prelude
 
 import           Fission.Challenge.Types
 
-import qualified Fission.Web.API.User.Email.Types            as API.User
+import qualified Fission.Web.API.User.Email.Verify.Types      as API.User.Email
 
 import           Fission.Web.Server.Challenge.Creator.Class
 import           Fission.Web.Server.Challenge.Verifier.Class
+import           Fission.Web.Server.Challenge.Retriever.Class
 
-import qualified Fission.Web.Server.Link                     as API
+import qualified Fission.Web.Server.Link                      as API
 
 verificationLink :: Challenge -> Text
 verificationLink challenge = toUrlPiece $ API.mkLink route challenge
   where
-    route = Proxy @("user" :> API.User.Email)
+    route = Proxy @("user" :> "email" :> API.User.Email.Verify)
--- a/fission-web-server/library/Fission/Web/Server/Challenge/Retriever/Class.hs
+++ b/fission-web-server/library/Fission/Web/Server/Challenge/Retriever/Class.hs
@@ -1,0 +1,29 @@
+module Fission.Web.Server.Challenge.Retriever.Class (Retriever (..)) where
+
+import           Database.Persist
+
+import           Fission.Prelude
+
+import           Fission.Challenge.Types
+import           Fission.Error                    as Error
+
+import           Fission.Web.Server.Models
+import           Fission.Web.Server.MonadDB.Types
+
+class Monad m => Retriever m where
+  retrieve :: UserId -> m (Either (NotFound UserChallenge) Challenge)
+
+instance MonadIO m => Retriever (Transaction m) where
+  retrieve userId = do
+    res <- selectFirst [ UserChallengeUserId ==. userId ] []
+    case res of
+      Nothing ->
+        return $ Left NotFound
+
+      Just (Entity _ UserChallenge { userChallengeHash, userChallengeUserId }) ->
+        if userId == userChallengeUserId
+          then do
+            return $ Right userChallengeHash
+
+          else do
+            return $ Left NotFound
--- a/fission-web-server/library/Fission/Web/Server/Email/Error.hs
+++ b/fission-web-server/library/Fission/Web/Server/Email/Error.hs
@@ -1,0 +1,20 @@
+module Fission.Web.Server.Email.Error (CouldNotSend(..)) where
+
+import           Servant.Client
+import           Servant.Server
+
+import           Fission.Prelude
+
+import           Fission.Web.Server.Error.Class
+
+newtype CouldNotSend = CouldNotSend ClientError
+  deriving ( Show
+           , Eq
+           )
+
+instance Display CouldNotSend where
+  display (CouldNotSend _) = "An error occured while trying to send an email"
+
+
+instance ToServerError CouldNotSend where
+  toServerError couldNotSend = err500 { errBody = displayLazyBS couldNotSend }
--- a/fission-web-server/library/Fission/Web/Server/Handler/User.hs
+++ b/fission-web-server/library/Fission/Web/Server/Handler/User.hs
@@ -10,6 +10,7 @@ import           Fission.Web.Server.IPFS.DNSLink.Class          as DNSLink
 import           Fission.Web.Server.WNFS.Class
 
 import qualified Fission.Web.Server.Challenge.Creator.Class     as Challenge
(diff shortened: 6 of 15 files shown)


expede:
UserChallenge
  userId    UserId
  hash      Challenge

  UniqueUserId    userId
  UniqueChallenge hash

  deriving Show Eq
  



Looks like it has unique indexes on userID and the challenge. It’s a 1:{0,1} relationship, not one to many.

matheus23:
That could still become a nullable field then, though, right?

expede:
> That could still become a nullable field then, though, right?

Yes they’re totally isomorphic. I suppose the current layout uses less space, though we’re really talking on the order of a couple bytes.

it shouldn’t matter one way or the other, though I suppose the current approach is more flexible down the road. Is there a problem with the current layout, or just curious why it’s not on the same table? I didn’t write it, but have a vague memory of having the conversation “sure it doesn’t really matter” way back when 😛

matheus23:
Just curious

expede:
Glanced over the code; looking pretty good so far 🙌

fission-web-server/library/Fission/Web/Server/Challenge.hs
```diff
@@ -1,23 +1,25 @@
...
+import           Fission.Web.Server.Challenge.Retriever.Class
 
-import qualified Fission.Web.Server.Link                     as API
+import qualified Fission.Web.Server.Link                      as API
 
 verificationLink :: Challenge -> Text
 verificationLink challenge = toUrlPiece $ API.mkLink route challenge
   where
-    route = Proxy @("user" :> API.User.Email)
+    route = Proxy @("user" :> "email" :> API.User.Email.Verify)
```
expede, line 25: Other comment because of this ☝️

fission-web-api/library/Fission/Web/API/User/Email/Types.hs
```diff
@@ -4,4 +4,10 @@ import           Fission.Web.API.Prelude
...
 import           Fission.Web.API.User.Email.Verify.Types
 
-type Email = "email" :> Verify
+import           Fission.Web.API.User.Email.Resend.Types
+
+type Email = "email" :> API
+
+type API
+  =    Verify
+  :<|> Resend
```
expede, line 13: To keep things totally type safe (see comment below), we should probably do something along these lines:

```haskell
type Email
  =    Verify'
  :<|> Resend'

type Prefix = "email"

type Verify'   = Prefix :> Verify
type Resend' = Prefix :> Resend
```

... plus an exported prefix for `User`.

Thoughts?

fission-web-server/library/Fission/Web/Server/Challenge/Retriever/Class.hs
```diff
@@ -0,0 +1,29 @@
...
+      Nothing ->
+        return $ Left NotFound
+
+      Just (Entity _ UserChallenge { userChallengeHash, userChallengeUserId }) ->
+        if userId == userChallengeUserId
+          then do
+            return $ Right userChallengeHash
+
+          else do
+            return $ Left NotFound
```
expede, line 29: 👍

fission-web-server/library/Fission/Web/Server/Handler/User.hs
```diff
@@ -42,7 +45,7 @@ handler ::
 handler = Create.create
      :<|> WhoAmI.handler
      :<|> Verify.handler
-     :<|> VerifyEmail.handler
+     :<|> (VerifyEmail.handler:<|> ResendEmail.handler)
```
expede, line 48: Let's make an `Email.handler` to keep the tree clean as we add more routes

fission-web-server/library/Fission/Web/Server/Handler/User/ResendEmail.hs
```diff
@@ -0,0 +1,35 @@
...
+handler ::
+  ( MonadThrow          m
+  , MonadLogger         m
+  , MonadEmail          m
+  , Challenge.Retriever m
+  )
+  => ServerT API.Email.Resend m
+
+handler Authorization { about = Entity _ User { userEmail = Nothing } } =
+  Web.Err.throw err422 { errBody = "There is no email associated with the user" }
```
expede, line 26: Pretty sure we discussed this one in Discord. Let's `Web.Err.throw $ NotFound Challenge`

fission-web-server/library/Fission/Web/Server/Handler/User/ResendEmail.hs
```diff
@@ -0,0 +1,35 @@
...
+
+handler Authorization { about = Entity userId User { userUsername = username, userEmail = Just email } } = do
+  challenge <- Web.Err.ensureM $ Challenge.retrieve userId
+
+  sendVerificationEmail (Recipient email username) challenge >>= \case
+    Left _ ->
+      Web.Err.throw err500 { errBody = "Could not send verification email" }
+    Right _ ->
+      return NoContent
```
expede, line 35: Also, no need to change anything, but you may be able to save yourself some typing by using a single function head in this case. But literally doesn't matter 👍

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -610,6 +610,10 @@ instance Challenge.Creator Server where
   create email =
     runDB $ Challenge.create email
 
+instance Challenge.Retriever Server where
+  retrieve userId =
+    runDB $ Challenge.retrieve userId
```
expede, line 615: 👍

fission-web-api/library/Fission/Web/API/User/Email/Resend/Types.hs
```diff
@@ -0,0 +1,14 @@
...
+import qualified Fission.Web.API.Auth.Types as Auth
+
+type Resend = "resend" :> Check
+
+type Check
+  =  Summary "Resend verification email"
+  :> Description "Send a verification email to currently authenticated user."
+  --
+  :> Auth.HigherOrder
+  :> PostNoContent
```
expede, line 14: 😎 👍

expede:
Looking really _really_ good 🎉

Just a couple minor comments, but basically good to go 😁

fission-web-server/library/Fission/Web/Server/Handler/User/ResendEmail.hs
expede: If you want a special message about the email-user relationship, we can also leverage the typesafe DB and have a `NotFound UserEmail`. Your choice 👍

fission-web-server/library/Fission/Web/Server/Handler/User/ResendEmail.hs
expede: Also same thing here. Let's slap an `ensureM` on the `sendVerificationEmail`, and make sure that it has the correct error.

expede: Ideally we get something more specific than a 500 as well. But also realize that it's not always feasible, depending on what we get back from Send In Blue

expede (changes requested):
haha forgot to do this in a "Request Changes" transaction. See notes inline.

fission-web-api/library/Fission/Web/API/User/Email/Types.hs
matheus23: Re: Type safety: I don't think there's a type safety issue. If I write
```hs
route = Proxy @("usser" :> "somethingwrong" :> API.User.Email.Verify)
```
The code type for `API.mkLink` prevents this from compiling:
```haskell
mkLink :: (IsElem route Fission.API, HasLink route) => Proxy route -> MkLink route Link
```

expede: Excellent point! I knew we set all this stuff up for a reason 😅

fission-web-server/library/Fission/Web/Server/Handler/User/ResendEmail.hs
matheus23: Hm. When I try to do something like `NotFound @UserEmail`, I get this error:
```
Data constructor ‘UserEmail’ cannot be used here
    (it comes from a data family instance)
```

I'm unfortunately not super type/data family literate, so I'm not sure why this doesn't work.

And I wouldn't want the error message to be something like "No email found" (which I'd define `instance Display (NotFound Email)` as), as I feel this is slightly confusing if there's not even an email in the request.

fission-web-server/library/Fission/Web/Server/Handler/User/ResendEmail.hs
matheus23: > Ideally we get something more specific than a 500 as well. But also realize that it's not always feasible, depending on what we get back from Send In Blue

Yeah, that's a good point. At least for now I've kept it possible to have the error depend on the `Servant.Client.ClientError` we get from the SIB request. I might take a look at the SIB API to maybe give better error messages.

matheus23: > Let's slap an ensureM on the sendVerificationEmail

FTR, did that :+1:

expede (approved):

fission-web-server/library/Fission/Web/Server/Email/Error.hs
```diff
@@ -0,0 +1,20 @@
...
+import           Servant.Server
+
+import           Fission.Prelude
+
+import           Fission.Web.Server.Error.Class
+
+newtype CouldNotSend = CouldNotSend ClientError
+  deriving ( Show
+           , Eq
+           )
```
expede, line 13: Also pro-tip: do you know about `deriving newtype`? Doesn't make a huge difference here, but mainly an FYI

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -632,7 +636,8 @@ instance MonadEmail Server where
         , params = Email.TemplateOptions verifyUrl name
         }
 
-    liftIO $ runClientM (Email.sendEmail apiKey emailData) env
+    mapLeft Email.CouldNotSend <$>
+      liftIO (runClientM (Email.sendEmail apiKey emailData) env)
```
expede, line 640: Also, no need to change, but I used to write a lot of this kind of code. Coming back many months later means having to reconstruct what's actually happening.

```haskell
liftIO (runClientM (Email.sendEmail apiKey emailData) env) >>= \case
  Left clientErr -> return $ Left Email.CouldNotSend clientErr
  Right _ -> return ok
```

This is often just clearer, if more LOC.

PR #436, opened 2021-01-15
Return non-zero exit codes on failure 

### LINKED ISSUE #381: Return non-zero exit code on failures.
# Summary

When, e.g. `fission up` fails, return a non-zero exit code, such as `1`. This makes it easier to use `fission up` for scripting.

## Problem

`fission up` returns exit code 0, even when it fails, e.g. when it fails with:
```
$ fission up
🚫 Server unable to sync data
$ echo $?
0
```

### Impact

Using `fission up` in e.g. bash scripts is a little annoying.

## Solution

Return an exit code other than 0, when there's an error.

walkah:
When a handler errors, make sure the CLI exits with a non-zero exit code. This will help the github action report errors more accurately.

Closes #381


expede (approved):
Looks awesome 🎉

PR #437, opened 2021-01-15
Direct fix

expede:
Separate PR for teaching 👩‍🏫

### DIFF
--- a/fission-cli/library/Fission/CLI/Handler/User/Register.hs
+++ b/fission-cli/library/Fission/CLI/Handler/User/Register.hs
@@ -50,6 +50,7 @@ register ::
   , IsMember ClientError (Errors m)
   , IsMember Key.Error   (Errors m)
   , Show (OpenUnion (Errors m))
+  , Contains (Errors m) (Errors m)
   )
   => Maybe Username
   -> Maybe Email
@@ -82,6 +83,7 @@ createAccount ::
   , m `Raises` AlreadyExists Ed25519.SecretKey
   , m `Raises` Username.Invalid
   , Show (OpenUnion (Errors m))
+  , Contains (Errors m) (Errors m)
   )
   => Maybe Username
   -> Maybe Email
@@ -137,6 +139,4 @@ createAccount maybeUsername maybeEmail = do
 
       CLI.Error.put err $
         errMsg <> " Please try again or contact Fission support at https://fission.codes"
-        -- raise err
-
-      createAccount maybeUsername maybeEmail
+      raise err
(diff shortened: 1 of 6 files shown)


fission-cli/library/Fission/CLI/Handler/User/Register.hs
```diff
@@ -56,12 +53,12 @@ register ::
   )
   => Maybe Username
   -> Maybe Email
-  -> m (Either (OpenUnion Errs) Username)
+  -> m Username
 register maybeUsername maybeEmail = do
   attempt (sendAuthedRequest User.whoami) >>= \case
     Right username -> do
       CLI.Success.alreadyLoggedInAs $ textDisplay username
-      Just username
```
expede, line 64: Putting the `Just` here at this line specialized the entire contextual `m` to be `Maybe`. The type inference said "oh, you're return is a `Maybe`! So then the type is `Maybe (Either (OpenUnion Errs) Username)`, right?!".

```haskell
data Maybe a 
  = Just a 
  | Nothing -- Error case
```

`Maybe` can't carry the `ClientError`, because it has nowhere to put it in that `Nothing`. Because the only way to raise things with a `Maybe` throws away information, it can only raise `()` (showing that it's super lossy).

fission-cli/library/Fission/CLI/Handler/User/Register.hs
```diff
@@ -140,6 +137,6 @@ createAccount maybeUsername maybeEmail = do
 
       CLI.Error.put err $
         errMsg <> " Please try again or contact Fission support at https://fission.codes"
-        raise err
-      -- createAccount maybeUsername maybeEmail
+        -- raise err
 
+      createAccount maybeUsername maybeEmail
```
expede, line 142: Throwing them back into the beginning of the account recovery loop. Maybe you're trying to avoid that for the GH Action, though?

fission-cli/library/Fission/CLI/Handler/User/Register.hs
```diff
@@ -85,10 +83,11 @@ createAccount ::
   , m `Raises` AlreadyExists Ed25519.SecretKey
   , m `Raises` Username.Invalid
   , Show (OpenUnion (Errors m))
+  , Contains (Errors m) (Errors m)
```
expede, line 86: This line looks weird, right? Of course `Contains a a`!

expede: Well, it's not an assertion that `Errors m` contains itself. It loops through the elements at compile time, and makes sure that they match. We don't depend on ordering, so the `Contains` constraint happens to be the thing that does that. I've tried to get rid of it, because it reads weird, but alas...

PR #438, opened 2021-01-15
Error code on exit fix 2

expede:
👩‍🏫  More teaching 😁

### DIFF
--- a/fission-cli/library/Fission/CLI/Handler/App/Publish.hs
+++ b/fission-cli/library/Fission/CLI/Handler/App/Publish.hs
@@ -9,6 +9,7 @@ import           System.FSNotify                 as FS
 import           RIO.Directory
 
 import           Network.HTTP.Types.Status
+import qualified Network.IPFS.Process.Error      as IPFS.Process
 
 import           Network.IPFS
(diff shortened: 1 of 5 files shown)


fission-cli/library/Fission/CLI/Handler/App/Publish.hs
```diff
@@ -76,25 +72,30 @@ publish watchFlag runner appURL appPath _updateDNS updateData = do -- FIXME upda
       absBuildPath <- liftIO $ makeAbsolute buildDir
       logDebug $ "Starting single IPFS add locally of " <> displayShow absBuildPath
 
-      CLI.IPFS.Add.dir absBuildPath >>= putErrOr \cid@(CID hash) -> do
```
expede, line 79: I also expanded this out, because out little `>>= putOrErr` is just... ugly I think?

expede: We could also turn some of these inline `CLI.Error.put err; raise err` into showing the message next to the exit code (now that it's all set up like that). It would make it less noisy for the common case. Thoughts?

PR #440, opened 2021-01-20
WS Relay

expede:
Extracting out of CLI linking, because it seems to work super well and we already want it for other things 👍

This is everything from CLI linking minus the actual CLI bits

> +1,351 −558 

This actually looks worse than it is. There's a couple files that moved, and a bunch of places that the formatter auto-aligned things.

# Changelog

* Build a socket server
  * Route
  * Server handler
  * Server config (pass around map of channels)
* Better errors by wrapping generic types like `String` in `newtypes` for...
  * JSON
  * Servant
* Add `facts` to UCAN
* Switch to OpenSSL for RSA key generation
  * It's... _so much_ faster
* Update dependencies as needed (since on Stackage-Nightly)
* No more `--no-nix`
  * Since we're building inside `shell.nix`, that's more robust anyways
* Lots of `newtypes` to keep encryption straight (encryption & string encoding can be crazy-making)

# Most Relevant Bits

`fission-web-server` has the most important changes


fission-core/library/Fission/Internal/Orphanage/RSA2048/Private.hs
```diff
@@ -2,12 +2,70 @@
...
+
+      public_size = OpenSSL.rsaSize pair
+      public_n    = OpenSSL.rsaN pair
+      public_e    = OpenSSL.rsaE pair
+
+      private_pub = RSA.PublicKey {public_size, public_n, public_e}
+
+      private_d = OpenSSL.rsaD pair
+      private_p = OpenSSL.rsaP pair
+      private_q = OpenSSL.rsaQ pair
```
expede, line 31: Much, much, much faster. Depends on the OpenSSL lib being available, though.

stack.yaml
```diff
@@ -38,10 +41,3 @@ extra-deps:
   - primitive-unlifted-0.1.3.0
   - run-st-0.1.1.0
   - tuples-0.1.0.0
-
-nix:
-  enable: true
-  packages:
-    - lzma
-    - postgresql
-    - zlib
```
expede, line 47: Removing in favour of `shell.nix`. Also means that that you don't need to `--no-nix` anymore 👍

stack.yaml
```diff
@@ -12,18 +12,21 @@ packages:
...
   - hfsevents-0.1.6
   - ipfs-1.1.5.1
   - lzma-clib-5.2.2
   - raven-haskell-0.1.4.0
   - rescue-0.2.1 # FIXME explore 0.3.0, or switch to checked-exceptions
+  - servant-auth-0.4.0.0
+  - servant-auth-server-0.4.6.0
+  - servant-auth-swagger-0.2.10.1
+  - servant-swagger-ui-redoc-0.3.3.1.22.3
+  - servant-websockets-2.0.0
```
expede, line 29: Alphabets. How do they work?

fission-cli/library/Fission/CLI/Types.hs
```diff
@@ -5,64 +5,65 @@ module Fission.CLI.Types
...
 
-import           Fission.Internal.Orphanage.BaseUrl                ()
-import           Fission.Internal.Orphanage.DNS.DNSError           ()
-import           Fission.Internal.Orphanage.OpenUnion              ()
+import           Fission.Internal.Orphanage.BaseUrl      ()
+import           Fission.Internal.Orphanage.DNS.DNSError ()
+import           Fission.Internal.Orphanage.OpenUnion    ()
 
-import           Fission.Web.Client.Internal.Orphanage.ClientError ()
+import           Fission.Internal.Orphanage.ClientError  ()
```
expede, line 66: The one line that was _actually_ changed. The rest is the formatter lining stuff up 😅

fission-cli/library/Fission/CLI/Key/Ed25519.hs
```diff
@@ -16,3 +16,14 @@ parseSecretKey ::
...
+-- TODO remove as part of cli-linking
+
+parseKey ::
+     (ByteArray.ScrubbedBytes -> CryptoFailable a)
+  -> ByteArray.ScrubbedBytes
+  -> Either Key.Error a
+parseKey f bytes =
+  case f bytes of
+    CryptoPassed sk  -> Right sk
+    CryptoFailed err -> Left $ Key.ParseError err
```
expede, line 29: This was removed from `fission-core`. This gets completely cut in the CLI Linking branch. Here to keep thing comping for this PR.

fission-web-server/library/Fission/Web/Server.hs
```diff
@@ -41,16 +41,21 @@ import           Fission.Web.Server.Config.Types
...
 import           Fission.Web.Server.Reflective
+import           Fission.Web.Server.Relay.Store.Class
 
 import           Fission.Internal.Orphanage.OctetStream ()
 import           Fission.Internal.Orphanage.PlainText   ()
 
 -- | Top level web API type. Handled by 'server'.
-type API = Swagger.API :<|> Fission.API
+type API    = Swagger.API :<|> Fission.API :<|> LinkWS
+type LinkWS = "user" :> "link" :> RelayWS
```
expede, line 58: Unfortunately, the creators of the servant-ws library have client and server together, which brings in the XZ dependency that ideally we avoid in the CLI 😠  So putting this in the server directly.

fission-web-server/library/Fission/Web/Server/Auth/DID.hs
```diff
@@ -30,8 +30,9 @@ handler ::
   -> m DID
 handler req =
   case Token.get req of
-    Right (Token.Bearer token@(Bearer.Token jwt rawContent)) ->
-      JWT.check rawContent jwt >>= \case
+    Right (Token.Bearer token@(Bearer.Token jwt rawContent)) -> do
+      serverDID <- getServerDID
+      JWT.check serverDID rawContent jwt >>= \case
```
expede, line 35: No longer has an implicit recipient in the UCAN, so we can do CLI<>browser linking. Needs to be passed in explicitly here for the server use case now ☝️

fission-web-server/library/Fission/Web/Server/Auth/Token/Basic.hs
```diff
@@ -88,6 +88,6 @@ checkUser (BasicAuthData rawUsername password) =
 
 parseBasic :: Auth.Basic.Token -> Either Auth.Error BasicAuthData
 parseBasic (Auth.Basic.Token token) =
-  case Ch.split ':' (Base64.decodeBase64Lenient token) of
+  case Ch.split ':' (Base64.decodeLenient token) of
```
expede, line 91: There's a few of these around. It's from updating a dependency. Reads better, so okay by me 👍

fission-web-server/library/Fission/Web/Server/Handler/Relay.hs
```diff
@@ -0,0 +1,21 @@
...
+relay :: (MonadIO m, MonadLogger m, MonadRelayStore m) => ServerT API.RelayWS m
+relay did conn = do
+  logDebug $ "Connected to user link for " <> textDisplay did
+  storeVar <- getStoreVar
+  (sentBufferVar, chanIn, chanOut) <- atomically $ Relay.setup did storeVar
+
+  liftIO $ WS.withPingThread conn 30 noop do
+    concurrently_
+      (inbound  conn chanIn  sentBufferVar)
+      (outbound conn chanOut sentBufferVar)
```
expede, line 21: This is 90% of the actual WS relay logic

fission-web-server/library/Fission/Web/Server/Relay.hs
```diff
@@ -0,0 +1,84 @@
...
+
+toFingerprint :: WS.DataMessage -> Digest SHA1
+toFingerprint msg =
+  hash $ Lazy.toStrict rawMsg
+  where
+    rawMsg :: Lazy.ByteString
+    rawMsg =
+      case msg of
+        WS.Text   lbs _ -> lbs
+        WS.Binary lbs   -> lbs
```
expede, line 84: Relay logic internals

fission-web-server/library/Fission/Web/Server/Relay/Channel/Types.hs
```diff
@@ -0,0 +1,17 @@
...
+
+import           Control.Concurrent.STM.TChan
+import qualified Network.WebSockets           as WS
+
+import           Fission.Error.NotFound.Types
+
+import           Fission.Prelude
+
+newtype ChannelIn  = ChannelIn  { getChanIn  :: TChan WS.DataMessage }
+newtype ChannelOut = ChannelOut { getChanOut :: TChan WS.DataMessage }
```
expede, line 14: Split up so that senders don't get their own messages (see handler)

fission-web-server/library/Fission/Web/Server/Relay.hs
```diff
@@ -0,0 +1,84 @@
...
+inbound ::
+     WS.Connection
+  -> ChannelIn
+  -> TVar (Set (Digest SHA1))
+  -> IO ()
+inbound conn (ChannelIn chanIn) sentBufferVar =
+  forever do
+    msg <- WS.receiveDataMessage conn
+    atomically do
+      modifyTVar' sentBufferVar \msgs -> Set.insert (toFingerprint msg) msgs
```
expede, line 47: @expede but why? Plz write some notes!

expede:
Investigating an infinite spinner on user signup on staging. I don't think it's related to this PR, but going to investigate & fix  here before merging

Issue #386, opened 2020-08-21
Simple "Magic" command(s)


dholms:
Something we talked about on our call today:

There's something nice about the brevity of `fission up`. I remember being pretty wowed running `now` (from zeit) for the first time. I'm not sure that same magic would be there with long commands like `fission app publish`.

What commands could we add that would give that same "magic" feeling?

We talked about, for instance, a simple `fission` cmd that would walk you through setup, app init, & app up all in one. Maybe a TUI?

bmann:
As discussed — just `fission` (or win or whatever) should “do the right thing”. 

Long form commands are for experts or for other shell commands to call.

walkah:
Started #442 to tackle this, which more deals with shortcuts defaulting to help text so far.

What does everyone thing just `fission` *should* do?

expede:
Closed #442

PR #442, opened 2021-01-20
Shortcuts and defaults ("magic" commands)

walkah:
(Result of a pairing session with @expede )

* This adds a new shortcut for `fission whoami` (to `fission user whoami`)
* Also, will show the full `--help` text anytime the parsing fails, including just running `fission` (or `fission not-a-command`, etc)

Re: #386


expede (approved):

expede:
> +13 −10

...and like 11 of those lines are subtly adjusting alignment on imports 😆 Ah stylish-haskell

PR #445, opened 2021-01-25
Allow multiple remote peers

dholms:
## Problem
After upgrading to IPFS Cluster, we now want to return multiple peer IDs, for all the nodes in the cluster.

## Solution
Change `env.yaml` so that it takes a `NonEmpty` list of remote peers. Update `/ipfs/peers` to return this list


expede (approved):
Looks awesome 👍 🎉

PR #447, opened 2021-01-27
Upgrade Stack+Nix Integration

expede:
# TL;DR

* No reason to `--no-nix` any more 😁 
* Pure environment (even on Mac)
* Break up dependencies into named groups

# Tested On

✅ M1 macOS Big Sur (Brooke's machine)
✅ x86_64 macOS Big Sur (James' machine)
✅ NixOS (Digital Ocean)
✅ Ubuntu (Fission EC2)
✅ Arch Linux (Philip's machine)

# FYIs

NB: See note in README about Lorri 🤷‍♀️ It's fine, but a bit annoying from time to time.

P.S. We could later go deeper on a build shell vs active development environment dependencies. We don't have that many for now, so I think it's fine, but let me know if you disagree!


shell.nix
```diff
@@ -3,29 +3,50 @@ let
...
+    # Fun
+    pkgs.figlet
+    pkgs.lolcat
   ];
+
+   shellHook = ''
+    export LANG=C.UTF8
+  
+    echo "Welcome to the"
+    ${pkgs.figlet}/bin/figlet "Fission Shell" | lolcat -a -s 50
```
expede, line 50: It's baaaaaaaack 😛

stack.yaml
```diff
@@ -41,3 +41,8 @@ extra-deps:
   - primitive-unlifted-0.1.3.0
   - run-st-0.1.1.0
   - tuples-0.1.0.0
+
+nix:
+  enable:     true
+  pure:       true
+  shell-file: shell.nix
```
expede, line 48: The [Stack/Nix integration](https://docs.haskellstack.org/en/stable/nix_integration/) has really improved!

shell.nix
```diff
@@ -1,31 +1,79 @@
...
-    hinotify
-
-    # Haskell
-    unstable.ghcid
-    unstable.stack
-    unstable.stylish-haskell
-    unstable.haskellPackages.hie-bios
-    unstable.haskell-language-server
-    unstable.haskellPackages.implicit-hie
+unstable.haskell.lib.buildStackProject {
```
expede, line 60: Stack integration ☝️

README.md
```diff
@@ -41,3 +41,18 @@ CLI interface to our service
...
+Unfortunately, the Stack integration doesn't play well with [`lorri`](https://github.com/target/lorri). If you've run `lorri` recently, you will either need to build with `--no-nix` or unset the `STACK_IN_NIX_SHELL` environment variable.
+
+```bash
+# bash or zsh
+unset STACK_IN_NIX_SHELL
+```
+
+```fish
+set -e STACK_IN_NIX_SHELL
+```
```
expede, line 58: This looks weird, and you're probably worried that `lorri` won't work. Well, `lorri` appears to be fine. Because the build is now pure, you don't need `lorri` to do the build itself — it's only for the other tools. It runs, but sets `STACK_IN_NIX_SHELL` (somehow). Unsetting that envar doesn't seem to effect lorri, and everything seems totally fine.

shell.nix
```diff
@@ -1,31 +1,79 @@
...
+    ];
+   
+    macos =
+      if pkgs.stdenv.isDarwin then
+        [ unstable.darwin.apple_sdk.frameworks.CoreServices
+          unstable.darwin.apple_sdk.frameworks.Foundation
+          unstable.darwin.apple_sdk.frameworks.Cocoa
+        ]
+      else 
+        [];
```
expede, line 56: `unstable.darwin.apple_sdk.frameworks.CoreServices` 🤯 

Okay, I was misinformed previously

README.md
walkah: I didn't have to do this, but it begs the question - do we still need lorri or should we drop it? we could also update .envrc to just`use_nix` ?

stack.yaml
walkah: Totally! this is awesome.

Do we want to bump stack to LTS-17 here & now? or are there issues still with it?

walkah (approved):
I can add x64-darwin to the ✅ builds! (and was able to also build cleanly on nixos). 🙌 🥳 

tenor-63217704

README.md
expede: So, we still want Lorri for other tooling. For example, if folks are going to run the server locally, they'll need Postgres. Having the language server, ghcide, and so on is also helpful so that getting an environment set up is as simple as `lorri shell`.

I kind of wonder about splitting out build from dev environment, but feels like overkill when it's ~5 or so tools.

> we could also update .envrc to just `use_nix`

Yeah, I'm actually unclear what `use_nix` gets us over lorri?

stack.yaml
expede: Fair, yeah we may as well. I have that in a different branch, so I didn't think about it. Will do now 👍

expede: Will see if that builds ⚙️ My other branch has a couple of these pinned dependencies updated as well.

README.md
walkah: Was just reading more about it after I wrote that... and I'm not sure? Other than lorri watches for changes. Yea, like i said, my build actually worked fine even with direnv + lorri so probably not worth making changes.

Although, if we're gonna change anything we should probably switch to https://github.com/nmattia/sorri because 🇨🇦

matheus23:
It builds for me again! :)


Nice.

PR #448, opened 2021-01-27
Pin to cluster

dholms:
## Problem
We're currently pinning CIDs to the remote IPFS Node. We're upgrading to a cluster.

## Solution
Pin to the cluster so that all nodes sync the data


fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Error.hs
```diff
@@ -0,0 +1,29 @@
...
+    FailureResponse _ response ->
+      response
+        |> responseBody
+        |> decode
+        |> \case
+          Just ErrorBody {message} ->
+            ClusterDaemonErr <| UTF8.textShow message
+
+          _ ->
+            UnexpectedOutput <| UTF8.textShow err
```
expede, line 26: Two things:

1. We're moving away from `<|`
2. `ClientError` has an (orphan) `Display` instance now

```haskell
parseClientError err = do
  logError $ displayShow err
  case err of
    FailureResponse _ response ->
      case decode $ responseBody response of
          Just ErrorBody {message} ->
            return . ClusterDaemonErr $ textDisplay message

          _ ->
            return . UnexpectedOutput $ textDisplay err
    unknownClientError ->
      UnknownPinErr $ textDisplay unknownClientError
```

fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Error/Types.hs
```diff
@@ -0,0 +1,30 @@
+module Fission.Web.Server.IPFS.Cluster.Error.Types
+  ( ErrorBody (..)
+  , Error (..)
+  ) where
+
+import Fission.Prelude
+
+data ErrorBody = ErrorBody {message :: String}
```
expede, line 8: (I know that this is all WIP)

With a single field, it may as well be a `newtype`. Also don't forget to give it a `Show` and `Display` instance :)

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -503,7 +521,7 @@ instance User.Modifier Server where
             return $ Error.openLeft err
 
           Right size -> do
-            IPFS.Pin.add newCID >>= \case
+            Cluster.pin newCID >>= \case
```
expede, line 524: Nice 👍

fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Client.hs
```diff
@@ -0,0 +1,19 @@
...
+
+import           Servant
+import           Servant.Client
+
+import qualified Network.IPFS.CID.Types as IPFS
+
+type API
+  = "pins"
+  :> Capture "cid" IPFS.CID
+  :> PostNoContent
```
expede, line 16: 🗺 The HTTP API types are down in `fission-web-api`

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -298,6 +299,23 @@ instance MonadDNSLink Server where
...
+      query = Cluster.pinClient cid
+      env = mkClientEnv manager url
+    
+    (liftIO $ runClientM query env) >>= \case
+      Left err -> do
+        formattedErr <- Cluster.parseClientError err
+        return $ Left formattedErr
+
+      Right _ -> 
+        return $ Right ()
```
expede, line 317: 👍

fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Error.hs
dholms: ah yup you're right. this code was (heavily :stuck_out_tongue:) inspired by code from `ipfs-haskell` which is still using `<|`

fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Client.hs
dholms: I'm not sure... I went back and forth on it a bit. But decided not to. It seems a little bit strange to have only 1 Cluster method in there while the rest is just ipfs.

I could honestly got either way tho :man_shrugging:

dholms: Maybe to clarify, this is hitting the ipfs-cluster daemon's api not the ipfs daemon

fission-web-server/library/Fission/Web/Server/IPFS/Pinner/Class.hs
```diff
@@ -0,0 +1,17 @@
...
+import qualified Network.IPFS.Add.Error                as IPFS.Pin
+import qualified Network.IPFS.Types                    as IPFS
+
+type Errors' = OpenUnion
+  '[ IPFS.Pin.Error
+   , Cluster.Error
+   ]
+
+-- | A monad representing the ability to pin a cid to IPFS (cluster or invidual daemon)
+class Monad m => MonadIPFSPinner m where
```
dholms, line 16: Unconvinced of the name of this. I decided I didn't want to abstract out all the way to `MonadIPFS` because then we have to carry that monad around with us anywhere we want a single function. So this Monad is just in charge of pinning some content: whether it's to a cluster or to a peer

Open to rethinking that or just changing the name (just sounds a little clunky, but couldn't think of another term for it)

fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Client.hs
expede: For sure. Let's leave it like this for now. Nothing stopping us from moving it to `Network.IPFS.Cluster` down the road.

Speaking of, I feel like `ipfs-haskell` could use some TLC. It's not a priority right now, but updating some deps and cleaning up the UI now that we're more familiar with IPFS would be nice.

fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Error/Types.hs
```diff
@@ -0,0 +1,35 @@
...
+  deriving Show
+
+instance Display ErrorBody where
+  display = displayShow
+
+
+instance FromJSON ErrorBody where
+  parseJSON = withObject "ErrorBody" \obj -> do
+    message    <- obj .: "message"
+    return ErrorBody {..}
```
expede, line 18: Partly a question on this file, partly one on `Cluster.Error`: do we ever get an error response with more than `message` in it? If so, anything worth capturing in there? Any value in a `raw :: Text` field that just passes through the entire error??

fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Client.hs
dholms: yea for sure^^ especially since it got some attention recently

fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Error/Types.hs
dholms: I don't believe so. The only errors I've seen just have a code and a message: id `{"code":400,"message":"error decoding Cid: selected encoding not supported"}`

We could just pass through the encoded JSON, but I don't see a huge benefit to it :man_shrugging:

fission-web-server/library/Fission/Web/Server/IPFS/Pinner/Class.hs
```diff
@@ -0,0 +1,17 @@
...
+
+import qualified Fission.Web.Server.IPFS.Cluster.Error as Cluster
+
+import qualified Network.IPFS.Add.Error                as IPFS.Pin
+import qualified Network.IPFS.Types                    as IPFS
+
+type Errors' = OpenUnion
+  '[ IPFS.Pin.Error
+   , Cluster.Error
+   ]
```
expede, line 13: Protip, but _no need to change here_.

With `OpenUnion`, we can have a bunch of separate errors that get composed as needed. So, for example:

```haskell
newtype ClusterDaemonErr = ClusterDaemonErr Text
newtype UnknownPinErr = UnknownPinErr    Text
newtype UnexpectedOutput = UnexpectedOutput Text

type ClusterError = '[ClusterDaemonErr, UnknownPinErr, UnexpectedOutput]
type IPFSError = '[...]

type Errors' = OpenUnion (ClusterError ++ IPFSErrors)
```

Now we can reuse `UnexpectedOutput` in a bunch of contexts, without needing to make `UnexpectedIPFSOutput`, `UnexpectedClusterOutput`, `UnexpectedServerOutput`, and so on

fission-web-server/library/Fission/Web/Server/IPFS/Pinner/Class.hs
expede: yeah, a bit clunky, but agreed that it's pretty granular right now. We may want to add stuff to this pinner later, so we could go with something like `MonadCIDStore` to handle our CID storage details.

Alternately, if we _want_ the "pin" part to be front-and-center, you can make the `IPFS` part of the name optional:

```haskell
class Monad m => MonadPinner m where
  pin :: IPFS.CID -> m (Either Errors' ())

-- ...

import qualified Fission.Web.Server.IPFS.Pinner.Class as IPFS

foo :: IPFS.MonadPinner m => m ()
```

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -298,6 +301,32 @@ instance MonadDNSLink Server where
...
 
+instance MonadIPFSPinner Server where
+  pin cid = 
+    asks clusterURL >>= \case
+      Nothing -> 
+        IPFS.Pin.add cid >>= \case
+          Left err -> 
+            return $ Error.openLeft err
+          Right _ -> 
+            return $ Right ()
```
expede, line 312: I really need to bring `rescue` over to the server 😅

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -298,6 +301,32 @@ instance MonadDNSLink Server where
...
+          query = Cluster.pinClient cid
+          env = mkClientEnv manager url
+        
+        (liftIO $ runClientM query env) >>= \case
+          Left err -> do
+            formattedErr <- Cluster.parseClientError err
+            return $ Error.openLeft formattedErr
+
+          Right _ -> 
+            return $ Right ()
```
expede, line 326: Totally fine as-is, but if you want to save a couple keystrokes:

```haskell
liftIO (runClientM query env) >>= \case
  -- ...
  Right _ -> 
    return ok
```

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -574,9 +603,9 @@ instance App.Modifier Server where
               Right Domain {domainZoneId} -> do
                 result <- if copyFiles
                             then
-                              IPFS.Pin.add newCID >>= \case
+                              IPFS.Pinner.pin newCID >>= \case
                                 Right _  -> return ok
-                                Left err -> return $ openLeft err
+                                Left err -> return $ Error.relaxedLeft err
```
expede, line 608: Again with `rescue` cleaning this all up 😅 

> Aside: I've also been considering switching to [checked exceptions](https://www.well-typed.com/blog/2015/07/checked-exceptions/), which has some Servant-related packages, but we loose the ability to recover an `Either`.

expede (approved):
Just a couple thoughts along the way; nothing to change.

fission-web-server/library/Fission/Web/Server/IPFS/Pinner/Class.hs
dholms: Oh cool :ok_hand: :ok_hand:

fission-web-server/library/Fission/Web/Server/Types.hs
dholms: ah yea nice! forgot about `ok`

PR #450, opened 2021-02-07
Proof of Concept — Fission Desktop

expede:
This is purely me messing around while working on linking. Mainly to show that it's possible. Also worth noting that this compiled from inside a pure Nix environment.

This works via GTK3, so _in theory_ it's cross platform.

Screen Shot 2021-02-07 at 1 09 45 PM


expede:
Since we have a bunch of web UIs, an Electron/Tauri app would make much more sense 😅 🚀

PR #451, opened 2021-02-09
Make UCANs backward compatible with FLOOFS

dholms:
## Problem
We parse UCANs expecting a `wnfs` path, but old versions of the CLI send it as `floofs`

## Solution
Change server parser to accept both

### DIFF
--- a/fission-core/library/Fission/Web/Auth/Token/UCAN/Resource/Types.hs
+++ b/fission-core/library/Fission/Web/Auth/Token/UCAN/Resource/Types.hs
@@ -36,9 +36,7 @@ instance FromJSON Resource where
     app    <- fmap FissionApp        <$> obj .:? "app"
     url    <- fmap RegisteredDomain  <$> obj .:? "domain"
 
-    let fs = maybe floofs Just wnfs -- fallback to floofs if UCAN does not include wnfs
-
-    case fs <|> app <|> url of
+    case wnfs <|> floofs <|> app <|> url of
       Just parsed -> return parsed
       Nothing     -> fail "Does not match any known Fission resource"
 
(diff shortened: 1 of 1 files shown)


fission-core/library/Fission/Web/Auth/Token/UCAN/Resource/Types.hs
```diff
@@ -31,9 +31,12 @@ instance ToJSON Resource where
...
   parseJSON = withObject "Resource" \obj -> do
-    fs  <- fmap FissionFileSystem <$> obj .:? "wnfs"
-    app <- fmap FissionApp        <$> obj .:? "app"
-    url <- fmap RegisteredDomain  <$> obj .:? "domain"
+    wnfs   <- fmap FissionFileSystem <$> obj .:? "wnfs"
+    floofs <- fmap FissionFileSystem <$> obj .:? "floofs" -- keep around floofs for backward-compatibility
+    app    <- fmap FissionApp        <$> obj .:? "app"
+    url    <- fmap RegisteredDomain  <$> obj .:? "domain"
+
+    let fs = maybe floofs Just wnfs -- fallback to floofs if UCAN does not include wnfs
```
expede, line 39: This totally works, but you can also do:

`fs = wnfs <|> floofs`

expede: Or, to break it out further:

```haskell
wnfs   <- obj .:? "wnfs"
floofs <- obj .:? "floofs"

-- ...

let fs = fmap FissionFileSystem (wnfs <|> floofs)
```

expede (approved):
Just a small comment for consistency / so you know that `<|>` exists, but 👍 🚢

PR #452, opened 2021-02-10
Enforce username checks in all tyep classes

expede:
This one is fully on me

We're using the "smart constructor" technique where you don't export the base constructor, and instead use a custom function that does checks. It's a really convenient way of enforcing checks at runtime without having to like... write a constructive proof on the incoming data. The checks happen at runtime, and this are wrapped in an `Either`.

The problem here is that inside the module that declares a the data type, you have compete access to the internal field. This is also the module where (most) of your type classes live (needs to be here, or where the class itself if defined).

I created a smart constructor, but failed to update the type classes. So, `FromJSON` was able to bypass the checks and create invalid `Username`s 🤦‍♀️ That's what I get for not enforcing things at the type level, right? 😉 

---

It _is_ nice that the problem is for sure scoped to this module, though 👍

### DIFF
--- a/fission-core/library/Fission/User/Username/Types.hs
+++ b/fission-core/library/Fission/User/Username/Types.hs
@@ -97,7 +97,7 @@ instance MimeUnrender PlainText Username where
   mimeUnrender _proxy bs =
     case decodeUtf8' $ Lazy.toStrict bs of
       Left unicodeErr ->
-        Left mconcat
+        Left $ mconcat
           [ "Username "
           , show bs
           , " contains invalid non-unicode character(s): "
--- a/fission-core/library/Fission/User/Username/Types.hs
+++ b/fission-core/library/Fission/User/Username/Types.hs
@@ -36,7 +36,11 @@ mkUsername txt =
     normalized = Text.toLower txt
 
 instance Arbitrary Username where
-  arbitrary = Username <$> arbitrary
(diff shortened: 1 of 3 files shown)


fission-core/library/Fission/User/Username/Types.hs
```diff
@@ -47,8 +51,18 @@ instance ToHttpApiData Username where
 instance PersistField Username where
   toPersistValue (Username un) = PersistText un
   fromPersistValue = \case
-    PersistText un -> Right (Username un)
-    other          -> Left ("Invalid Persistent Username: " <> Text.pack (show other))
+    input@(PersistText txt) ->
+      -- NOTE this may break against old DB records, but we really should keep this strict
+      -- I'd rather fail fast so know about it so we can fix it ~expede
```
expede, line 56: This is the controversial bit. Basically, don't trust what's in the DB, even if that breaks old bad records. This is guarding against things like us trying to use non-URL-safe characters for DNS.

expede:
Waiting for tests, but in theory this is fixed now 😅

walkah (approved):
tenor-116279911

Happy to have this one fixed, like for real 🎉 🤗

expede:
Holy hell, we really need to figure out how to cache test artifacts. I'm going to run the tests locally and merge if it passes

PR #453, opened 2021-02-11
Only use "allow listed" app names in the Fission namespace

expede:
While working on #452, I realized that we now give the ability to specify an app name, but we're not enforcing that no one registers `admin.fission.app`. This corrects that, using the same rules as for `Username`s.


fission-core/library/Fission/URL/Validation.hs
```diff
@@ -0,0 +1,90 @@
+-- | Username validation
+module Fission.URL.Validation
```
expede, line 2: Extracted general URL safe validation from `Username`

fission-core/library/Fission/URL/Subdomain/Types.hs
```diff
@@ -72,226 +67,5 @@ instance MimeUnrender PlainText Subdomain where
...
-  , "sprite"
-  , "alien"
-  , "ghost"
-  , "martian"
-  , "princess"
-  , "prince"
-  , "chef"
-  , "barista"
-  , "ufo"
-  ]
```
expede, line 297: Moved this to `AppName`, since that's what it is. Subdomain is a much more general thing, and we may want to process subdomains on user domains which don't have the same `${appName}.fission.app` format

fission-core/library/Fission/Security/Digestable/Class.hs
```diff
@@ -0,0 +1,19 @@
...
+
+-- | Create a digest
+class Digestable a where
+  digest :: a -> SecretDigest
+
+instance Digestable ByteString where
+  digest bs = UTF8.textShow (hash bs :: Digest SHA3_512)
+
+instance Digestable Text where
+  digest = digest . encodeUtf8
```
expede, line 19: Following our pattern of breaking classes out of other modules. It just.. felt weird where it was.

fission-cli/library/Fission/CLI/Parser/Command/App/Init/Types.hs
```diff
@@ -2,18 +2,18 @@ module Fission.CLI.Parser.Command.App.Init.Types (Options (..)) where
...
-  , buildDir     :: Maybe FilePath
-  , maySubdomain :: Maybe URL.Subdomain
-  , ipfsCfg      :: IPFS.Config
-  , verboseFlag  :: VerboseFlag -- ^ Verbose flag
+  { appDir      :: FilePath
+  , buildDir    :: Maybe FilePath
+  , mayAppName  :: Maybe App.Name
+  , ipfsCfg     :: IPFS.Config
+  , verboseFlag :: VerboseFlag -- ^ Verbose flag
   } deriving (Show, Eq)
```
expede, line 17: Since we change the API to enforce the rules, we also validate this on the CLI to make sure we're passing valid data. This isn't only a good practice so we can validate CLI-side, but the `Web.API` types also enforce it so this change is nonoptional

expede:
> +1,049 −931

hahhahaha. That's ~98% moving the random app name generator to `Fission.App.Name`.

fission-core/library/Fission/URL/Validation.hs
```diff
@@ -0,0 +1,90 @@
...
+-- >>> isValid "with.dot"
+-- False
+--
+-- >>> isValid "has.two.dots"
+-- False
+--
+-- >>> isValid "name&with#chars"
+-- False
+isValid :: Text -> Bool
+isValid username =
```
walkah, line 66: minor, but this is no longer always a `username` correct?

fission-core/library/Fission/App/Name/Types.hs
```diff
@@ -0,0 +1,323 @@
...
+  deriving newtype (Eq, Show, Display)
+
+mkName :: Text -> Either App.Name.Invalid Name
+mkName txt =
+  if isValid normalized
+    then Right $ Name normalized
+    else Left App.Name.Invalid
+
+  where
+    normalized = Text.toLower txt
```
expede, line 29: @walkah here we are exporting the _type_ but not the _data constructor_. This means that to create a `Name` outside of this module, you _must_ go through `mkName`. This is a simple technique for enforcing runtime validation called a "smart constructor".

walkah: Similar to `mkUsername` / Username right?

walkah (approved):
I think I get it! 🧠 🚀 

tenor-112137262

fission-core/library/Fission/URL/Validation.hs
expede: Good catch. Will fix!

expede:
> Similar to mkUsername / Username right?

Yup, the exact same pattern as Username. Slightly different class instances. My first run at this just `newtype` wrapped (IIRC) `Username` with most instance delegated to `Username`

fission-core/library/Fission/App/Name/Types.hs
expede: Yup the exact same technique

PR #456, opened 2021-02-18
Cluster pin management

expede:
🎗️ NOTE: requires adding the URLs of nodes to pin to in the `env.yaml`


fission-core/library/Fission/Unit/Prefix/Types.hs
```diff
@@ -0,0 +1,116 @@
+module Fission.Unit.Prefix.Types
+  ( Unity (..)
+  , Deci  (..)
+  , Centi (..)
+  , Milli (..)
+  , Micro (..)
+  , Pico  (..)
+  ) where
```
expede, line 8: This module is me getting tired of going "is this in seconds, milliseconds, microseconds, or nanoseconds". Just tracking it with the types, and having it do order-of-magnitude conversions for us.

fission-web-server/library/Fission/Web/Server.hs
```diff
@@ -57,10 +56,6 @@ import           Fission.Internal.Orphanage.PlainText   ()
 type API    = Swagger.API :<|> Fission.API :<|> LinkWS
 type LinkWS = "user" :> "link" :> RelayWS
 
--- | Run actions described by a @Fission@ type
-runServer :: MonadIO m => Config -> Fission.Server a -> m a
-runServer cfg actions = runRIO cfg $ unServer actions
-
```
expede, line 63: Moved into the `Fisssion.Web.Server.Types` module, since we need it in there sometimes

fission-web-server/library/Fission/Web/Server/Challenge/Verifier/Class.hs
```diff
@@ -20,15 +20,11 @@ instance MonadIO m => Verifier (Transaction m) where
...
         return $ Left NotFound
 
-      Just (Entity challengeId UserChallenge { userChallengeHash, userChallengeUserId }) ->
+      Just (Entity _ UserChallenge { userChallengeHash, userChallengeUserId }) ->
         if challenge == userChallengeHash
           then do
-            update userChallengeUserId
-              [ UserVerified  =. True ]
-
-            delete challengeId
```
expede, line 29: While we have the hood up, stop removing the challenges. It's a one-way switch (unverified -> verified), so like... are they going to verify it again or something? It's fine.

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -397,15 +325,53 @@ instance IPFS.MonadLocalIPFS Server where
```
expede, line 325: Below are the main substantive changes 👇

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -397,15 +325,53 @@ instance IPFS.MonadLocalIPFS Server where
...
-    IPFS.URL url <- asks ipfsURL
-    manager      <- asks httpManager
+    clusterURLs       <- asks ipfsURLs
+    IPFS.Timeout secs <- asks ipfsTimeout
+    manager           <- asks httpManager
+
+    logDebug @Text "Running IPFS request across cluster"
+    requests <- forM clusterURLs \(IPFS.URL url) ->
+      Process.asyncFor (Seconds (Unity secs)) do
+        runClientM query $ mkClientEnv manager url
```
expede, line 335: Pretty straightforward: fork (on green threads) `Async` processes ___. You can think of these as having a similar interface as JS Promises — it's a thing that may or may not be done yet. We _can_ block with `wait`, but since we want to react on the _first_ completion, we use `poll` which returns `Nothing` if it's still running

expede: The `Seconds` is a `maxTime`

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -397,15 +325,53 @@ instance IPFS.MonadLocalIPFS Server where
...
+        -> IO (Either ClientError a)
+      untilDone asyncRefs = do
+        results <- sequence (getStatus <$> asyncRefs)
+        case Process.progress results of
+          Process.Failed err -> return $ Left err
+          Process.InProgress -> untilDone asyncRefs
+          Process.Success a  -> return $ Right a
+      -- NOTE letting GHC's scheduler handle the frequency of retries.
+      -- This should be pretty lightweight, since we're just waiting for processes to return.
+      -- That said, if we go to 99% CPU or something, this is probably a good place to look 😉
```
expede, line 351: Per the note: we're not manually delaying the threads. Since we're asking the `Async` container if it's done, that's extremely lightweight, and I believe that the GHC scheduler is pretty smart about this.

That said, if we find that this is just spinning like wild we can add a 100ms delay to the `InProgress` case.

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -397,15 +325,53 @@ instance IPFS.MonadLocalIPFS Server where
...
+
+              Right maybeTimedOut ->
+                case maybeTimedOut of
+                  Left Process.TimedOut -> -- ⏲️  Timed out
+                    return . Process.Failed . ConnectionError $ toException Process.TimedOut
+
+                  Right requestResult ->
+                    case requestResult of -- 🌐 HTTP request result
+                      Left  clientErr -> return $ Process.Failed  clientErr
+                      Right a         -> return $ Process.Success a
```
expede, line 374: Lots of layers happening in that `Async` 😛 This is a consequence of us wanting a first-past-the-post strategy. A `race` would have been much simpler, but we _want_ the rest of the requests to continue while we report success to the user. Which means we need to add a timeout so they don't run literally forever in the degenerate case.

1. The `Async` itself may not be done (`poll :: Async a -> IO (Maybe a)`)
2. The inner process may throw an IO-based exception (`Either SomeException b`)
3. We have a timeout watcher, so that's an `Either Process.Timeout c`
4. Finally, the result of the HTTP request — `Either ClientError d`

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -305,76 +303,6 @@ instance MonadDNSLink Server where
 instance MonadLinkedIPFS Server where
   getLinkedPeers = asks ipfsRemotePeers
 
-instance MonadIPFSPinner Server where
```
expede, line 308: Less useful when we're making all requests against the entire cluster.

There's a version where we only request against one node and have a separate cluster manager.

...ugh, okay I just convinced myself (again) that's a good idea. Will put this back. Sorry, Daniel 😛

fission-web-server/library/Fission/Web/Server/Types.hs
dholms: yeah was searching for a poll interval. that's super cool :ok_hand:

fission-web-server/library/Fission/Web/Server/Types.hs
expede: Made an `MonadIPFSCluster` class instead

fission-web-server/library/Fission/Web/Server/Types.hs
expede: Turns out that there's a function that does this already called `waitAny`. It's a bit more sophisticated under the hood as well, and definitely has special stuff in the runtime. Switched to that and ripped our all of this `Process.Success` stuff.

fission-core/library/Fission/Web/Async.hs
```diff
@@ -0,0 +1,93 @@
...
+-- | Wait for all cluster peers to complete.
+waitAll :: MonadIO m => NonEmpty (Async (Either ClientError a)) -> m (NonEmpty (Either ClientError a))
+waitAll asyncRefs = liftIO $ forConcurrently  asyncRefs \ref ->
+  normalizeResult <$> waitCatch ref
+
+normalizeResult :: Either SomeException (Either ClientError a) -> Either ClientError a
+normalizeResult = \case
+  Left someException       -> Left $ ConnectionError someException
+  Right (Left clientError) -> Left clientError
+  Right (Right val)        -> Right val
```
expede, line 93: @dholms this is the same stuff as last time you looked, but extracted out to work on in any async web context. Added a couple more strategies while I was in there 😜  (it was quick)

fission-core/library/Fission/Web/Async.hs
```diff
@@ -0,0 +1,93 @@
...
+    go lastKnown = do
+      Process.sleepThread idleTimeout
+      latest <- readTVarIO latestVar
+      if latest == lastKnown
+        then do
+          asyncRef `cancelWith` Process.TimedOut
+          return $ Left Process.TimedOut
+
+        else
+          go latest
```
expede, line 71: @dholms You know that idle timeout that we keep talking about writing? ☝️

fission-web-server/library/Fission/Web/Server/IPFS/Cluster.hs
```diff
@@ -1,7 +1,45 @@
...
+    Left err -> return $ Left err
+    Right _  -> return $ Right ()
+
+pinStream :: MonadIPFSCluster m => CID -> m (Either ClientError ())
+pinStream cid = do
+  pseudoStreams <- streamCluster $ (Streaming.client $ Proxy @PinComplete) (Just cid) (Just True)
+  let asyncRefs = fst <$> pseudoStreams
+  waitAnySuccessCatch asyncRefs >>= \case
+    Left err -> return $ Left err
+    Right _  -> return $ Right ()
```
expede, line 45: So that we can do the idle timeout (i.e. requires getting streaming updates from the request)

fission-web-server/library/Fission/Web/Server/IPFS/Streaming/Pin/Types.hs
```diff
@@ -0,0 +1,30 @@
...
+  { pins     :: [CID]
+  , progress :: Natural
+  }
+  deriving (Eq, Show)
+
+instance FromJSON PinStatus where
+  parseJSON = withObject "IPFS.PinStatus" \obj -> do
+    pins     <- obj .: "Pins"     .!= []
+    progress <- obj .: "Progress" .!= 0
+    return PinStatus {..}
```
expede, line 30: We should add some of this stuff to `ipfs-haskell`. It needs a facelift in general IMO.

fission-web-server/library/Fission/Web/Server/Internal/Production.hs
```diff
@@ -118,11 +119,32 @@ runInProd overrideVerbose action = do
...
   putStrLnIO "      🎛️  Configuring..."
-  httpManager <- HTTP.newManager HTTP.defaultManagerSettings
-                   { HTTP.managerResponseTimeout = HTTP.responseTimeoutMicro clientTimeout }
 
-  putStrLnIO "      ✨ Creating manager"
-  tlsManager  <- HTTP.newManager HTTP.tlsManagerSettings
+  let
+    httpSettings = HTTP.defaultManagerSettings
+      { HTTP.managerResponseTimeout = HTTP.responseTimeoutMicro clientTimeout
+      , HTTP.managerConnCount       = 200
```
expede, line 126: Each client HTTP manager can handle a certain number of open connections. The default is 10, which obviously won't work for us when we're making parallel requests to 5+ nodes per request. Bumped to 200.

The docs say that we should prefer one manager when possible (I think for efficiency and setting an upper limit on resource use). We can always create these in the cluster handler, but for now 200 (~40 concurrent requests) feels pretty reasonable.

expede: Broke out one for IPFS to handle the IPFS timeout for us. This may be unnecessary, but we probably want that as a separate dial from, say, making email requests.

fission-web-server/library/Fission/Web/Server/Config/Types.hs
```diff
@@ -32,6 +32,7 @@ data Config = Config
   --
   , httpManager       :: HTTP.Manager
   , tlsManager        :: HTTP.Manager
+  , ipfsHttpManager   :: HTTP.Manager
```
dholms, line 35: What's the benefit to having a separate http manager here?

dholms: Ah is it just being able to set a different timeout?

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
```diff
@@ -19,3 +39,71 @@ deleteByUsername userNameTxt =
...
+    combiner acc cid =
+      Cluster.pin cid >>= \case
+        Left err -> do
+          logWarn $ "Unable to pin " <> display err
+          return ((cid, err) : acc)
+
+        Right _ ->
+          return acc
+
+getAllDBPins :: Server [CID]
```
dholms, line 90: Should we be basing this off of Events in the DB so that we get historical pins as well? 
I don't think it's necessary since they aren't referenced anywhere, but just thought I'd mention it, because this is not actually ensuring the pinning of the _entire_ system

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
```diff
@@ -19,3 +39,71 @@ deleteByUsername userNameTxt =
...
+  dbCIDs      <- getAllDBPins
+  clusterURLs <- asks ipfsURLs
+  manager     <- asks ipfsHttpManager
+
+  forM_ clusterURLs \(IPFS.URL url) -> do
+    let clientManager = mkClientEnv manager url
+
+    liftIO (runClientM listPins $ mkClientEnv manager url) >>= \case
+      Left err ->
+        logError $ "Pin list reuqest failed: " <> displayShow err
```
dholms, line 56: "reuqest" (spelling)

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
```diff
@@ -19,3 +39,71 @@ deleteByUsername userNameTxt =
       User.getByUsername uname >>= \case
         Just (Entity userId _) -> runDB $ deleteCascade userId
         Nothing                -> error "User doesn't exist"
+
+---
+
+ensureAllPinned :: Server ()
```
dholms, line 45: Awesome :raised_hands: :raised_hands:

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -397,15 +337,51 @@ instance IPFS.MonadLocalIPFS Server where
...
+              Left errMsg ->
+                return . Left . ConnectionError . toException $ GenericError errMsg
+
+              Right vals ->
+                case List.lastMaybe vals of
+                  Nothing    -> return . Left . ConnectionError . toException $ NotFound @PinStatus
+                  Just final -> return $ Right final
+
+      asyncIdleTimeout (Seconds (Unity (120 :: Natural))) asyncRef latestVar
+      return (asyncRef, resultChan)
```
expede, line 384: This is fairly quick & dirty, but we're not doing any "real" stream processing at the moment, so pushing over TChans as a way to get live updates if needed is here for convenience.

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -325,53 +327,24 @@ instance IPFS.MonadLocalIPFS Server where
...
+      clientManager = mkClientEnv manager url
 
-    logDebug @Text "Running IPFS request across cluster"
-    requests <- forM clusterURLs \(IPFS.URL url) ->
-      Process.asyncFor (Seconds (Unity secs)) do
-        runClientM query $ mkClientEnv manager url
+    liftIO $ runClientM query clientManager
 
-    liftIO $ untilDone requests
+instance MonadIPFSCluster Server where
```
dholms, line 350: Do we want to choose a different name for this other than "Cluster" since we're not actually using ipfs-cluster? I can see two possible future confusions:
- we end up using `ipfs-cluster` for something down the line & we now want to delineate between actual interactions with `ipfs-cluster` and interactions with our "group of ipfs nodes"
- someone new hops in the repo and is confused by why the `IPFSCluster` monad doesn't interact with an actual cluster

Probably not a big deal, and both of those could be resolved fairly easily, but just thought I'd mention :man_shrugging:

fission-core/library/Fission/Web/Async.hs
dholms: :raised_hands:

fission-web-server/library/Fission/Web/Server/IPFS/Streaming/Pin/Types.hs
dholms: :+1: totally

fission-web-server/library/Fission/Web/Server/Internal/Production.hs
dholms: ah okay :ok_hand: you answered my question from another comment

fission-web-server/library/Fission/Web/Server/IPFS/Cluster.hs
dholms: :+1: this seems like it also gets us like 80% of the way to streaming progress back to the client

expede: Yeah, it does. The only thing that we're really missing for that is it changes the contract, so we need new endpoints to do streaming.

When we do that, we may want to explore proper streaming as well (e.g. [streamly](https://hackage.haskell.org/package/streamly))

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
expede: @dholms can you expand on that? I'm not sure I follow.

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
expede: hahhahaha tahnks

fission-web-server/library/Fission/Web/Server/Types.hs
expede: Happy to change the name. It is technically a cluster, but yeah, it's not talking to `ipfs-cluster`. Any name suggestions?

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
dholms: We never unpin CIDs on, for instance, app updates or user data root updates.

So if we treated the original remote node's pin list as the "canonical" pin list, it would include some CIDs that are not included in the DB queries here. But those CIDs would be included in the Event Tables. 

However, we don't need that content for anything (and recent FS updates contain pointers to `previous` versions of the FS), so I think it's okay that we don't pin those. I just wanted to note the discrepancy.

Does that make sense?

fission-web-server/library/Fission/Web/Server/Types.hs
dholms: Heh yeah I really do like cluster the best, was hoping you'd have an idea :stuck_out_tongue: 

Nothing I can think of is descriptive enough. "swarm" "network" etc all refer to something else, and "group" or "suite", etc aren't as descriptive

So yeah maybe stick with cluster. Idt think the possible future confusion is very severe

dholms: Screenshot from 2021-02-22 12-05-13
We could also call it `TheWholeIPFSKitAndCaboodleMonad` :thinking:

PR #461, opened 2021-02-23
Don't print an "Unable to parse config" warning

matheus23:
The CLI used to print an error when fission wasn't set up on the machine yet.

Now the error is converted into a `logDebug`.

### DIFF
--- a/fission-cli/library/Fission/CLI/Release.hs
+++ b/fission-cli/library/Fission/CLI/Release.hs
@@ -8,7 +8,6 @@ import           Fission.Prelude
 
 import           Fission.Error.NotFound.Types
 
-import qualified Fission.CLI.Display.Error    as CLI.Error
 import           Fission.CLI.Display.Text
 import           Fission.CLI.Environment      as Env
 import qualified Fission.CLI.Meta             as Meta
@@ -27,12 +26,11 @@ checkLatestRelease ::
   , MonadEnvironment m
   , m `Raises` YAML.ParseException
   , m `Raises` NotFound FilePath
-  , Show (OpenUnion (Errors m))
   )
   => m ()
 checkLatestRelease = do
   attempt Env.get >>= \case
-    Left  err ->
+    Left  _ ->
       -- Also happens when the CLI isn't setup yet
       logDebug @Text "Unable to parse config"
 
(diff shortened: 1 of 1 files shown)


fission-cli/library/Fission/CLI/Release.hs
```diff
@@ -33,7 +33,8 @@ checkLatestRelease ::
 checkLatestRelease = do
   attempt Env.get >>= \case
     Left  err ->
-      CLI.Error.put err "Unable to parse config"
+      -- Also happens when the CLI isn't setup yet
+      logDebug @Text "Unable to parse config"
```
expede, line 37: We should also re-surface the error so that the caller can be alerted that this jappened

```haskell
Left err -> do
  logDebug @Text "Unable to parse config"
  return $ Left err
```

expede: (so that they can react and let the user know contextually)

matheus23: Ah, yeah right. It *does* swallow the errors! I was thinking it didn't do that :sweat_smile: The types misguided me, but yeah, we talked about that and `rescue`.

expede: [Discussed in Discord, but for those playing along at home]

Yeah, it's clearer in 0.3.x

```haskell
attempt :: innerM a -> outerM (Either (ErrorCase innerM) a) 
```

...where

expede: > I'm pretty sure it's intentional that the errors get swallowed. Otherwise the dispatch function doesn't continue with it's job.

👀

expede: Hmm maybe we want to push this down into the handlers 🤔 Anyhow, right now we can tell users to throw a `--verbose` on it to diagnose if they report issues (which we already do)

expede: Which is to say: you're right! Swallowing the error makes sense for now. Please make a quick tracking Issue and we can decide later 👍

matheus23: My thinking was this: If the version check fails, it's not so bad, so ignoring errors is fine.
If something else more critical needs the environment, it'll try to read the config file again anyway, and react to the errors accordingly.

expede (approved):

fission-cli/library/Fission/CLI/Release.hs
expede: Right right 👍 Whiiiich is probably why it's the way it is. Got it 😅

PR #464, opened 2021-03-03
Meaningful IPFS cluster timeout ⏱

expede:
Well, this is what I get for having written most of the original code at like 2am 🤷‍♀️ 

Will test on staging before merging 🧪 

Big thanks to Boris for reporting, and I swear this wasn't an all evening thing. Both what the problem likely was and the solution literally came to me while I was brushing my teeth 🦷🧹

### DIFF
--- a/fission-core/library/Fission/Error/GenericError/Types.hs
+++ b/fission-core/library/Fission/Error/GenericError/Types.hs
@@ -4,4 +4,5 @@ module Fission.Error.GenericError.Types (GenericError (..)) where
 import           Fission.Prelude
 
 newtype GenericError = GenericError String
-  deriving (Eq, Show, Exception)
+  deriving stock    (Eq, Show)
+  deriving anyclass (Exception)
--- a/fission-core/library/Fission/Web/Async.hs
+++ b/fission-core/library/Fission/Web/Async.hs
@@ -53,7 +53,7 @@ asyncIdleTimeout ::
   , FromPrefixed prefix Double
   )
   => Seconds prefix n
-  -> Async (Either err a)
+  -> Async a
   -> TVar  (Maybe a)
   -> m (Async (Either Process.TimedOut ()))
 asyncIdleTimeout idleTimeout asyncRef latestVar =
--- a/fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Class.hs
+++ b/fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Class.hs
@@ -22,4 +22,4 @@ class MonadIO m => MonadIPFSCluster m a where
 
   streamCluster ::
        Stream.ClientM (SourceIO a)
-    -> m (NonEmpty ((Async (Either ClientError a)), TChan a))
+    -> m (NonEmpty ((Async (Either ClientError a)), TChan (Either ClientError a)))
--- a/fission-web-server/library/Fission/Web/Server/Types.hs
+++ b/fission-web-server/library/Fission/Web/Server/Types.hs
@@ -8,7 +8,7 @@ import           Control.Monad.Catch                         hiding (finally)
 import           Control.Monad.Except
 
 import qualified RIO.ByteString.Lazy                         as Lazy
-import qualified RIO.List                                    as List
+-- import qualified RIO.List                                    as List
 import           RIO.NonEmpty                                as NonEmpty
 import           RIO.NonEmpty.Partial                        as NonEmpty.Partial
 import qualified RIO.Text                                    as Text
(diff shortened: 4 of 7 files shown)


fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -373,25 +373,44 @@ instance (Eq a, Display a) => MonadIPFSCluster Server a where
...
+                    atomically do
+                      latestVar  `writeTVar`  Just (Left err)
+                      resultChan `writeTChan` Left err
+
+                withVal x =
+                  runServer cfg do
+                    logDebug $ "🐙📥 Cluster node " <> display url <> " streamed value: " <> display x
+                    atomically do
+                      latestVar  `writeTVar`  Just (Right x)
+                      resultChan `writeTChan` Right x
```
expede, line 397: The core problem was probably this ☝️ We didn't actually push anything into the channels, so these would hit the timeout.

expede:
UPDATE: appears to work, though the original issue is intermittent. Was able to upload the Arch Linux ISO no problem, and it appears to actually be there.

IPFS locally seems to have a much harder time with lots of small files

fission-core/library/Fission/Web/Async.hs
```diff
@@ -53,7 +53,7 @@ asyncIdleTimeout ::
   , FromPrefixed prefix Double
   )
   => Seconds prefix n
-  -> Async (Either err a)
+  -> Async a
   -> TVar  (Maybe a)
   -> m (Async (Either Process.TimedOut ()))
 asyncIdleTimeout idleTimeout asyncRef latestVar =
```
expede, line 59: `Async a` will get inferred to `Async (Either ClientError value)` at the call site. Thanks to the type changing on the class, this function doesn't need to know as much about the innards

fission-web-server/library/Fission/Web/Server/IPFS/Cluster/Class.hs
```diff
@@ -22,4 +22,4 @@ class MonadIO m => MonadIPFSCluster m a where
 
   streamCluster ::
        Stream.ClientM (SourceIO a)
-    -> m (NonEmpty ((Async (Either ClientError a)), TChan a))
+    -> m (NonEmpty ((Async (Either ClientError a)), TChan (Either ClientError a)))
```
expede, line 25: The change mentioned above

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -373,25 +372,44 @@ instance (Eq a, Display a) => MonadIPFSCluster Server a where
 
       asyncRef <- liftIO $ async do
         Stream.withClientM streamQuery (mkClientEnv ipfsHttpManager url) \event ->
-          runServer cfg $
```
expede, line 376: For reference: `runServer :: Config -> Fission.Server a -> IO a`. So it turns "stuff that knows rich info about our server setup" into "hey it's running some generic IO for reals" (which is _always_ the pattern).

This got pushed down into each branch, mostly because `foreach` needs to return `IO ()` for the success and failure conditions (we could have left it here and done a bunch more `liftIO`, but it was ugly). Remember that this is happening in a stream off of a real world HLS connection, which means that `IO` is happening in here in general (they _could_ have made it a `MonadIO m` but didn't).

We totally _could_ make this work directly in `Fission.Server`, but we're already in `IO` because of that `async` on line 373 forking a process, so this is easier.

fission-web-server/library/Fission/Web/Server/Types.hs
expede: @walkah lemme know if you'd like a refresher on `TVar`/`TChan`

fission-core/library/Fission/Error/GenericError/Types.hs
```diff
@@ -4,4 +4,5 @@ module Fission.Error.GenericError.Types (GenericError (..)) where
 import           Fission.Prelude
 
 newtype GenericError = GenericError String
-  deriving (Eq, Show, Exception)
+  deriving stock    (Eq, Show)
+  deriving anyclass (Exception)
```
expede, line 8: This is just a compiler warning. It said that it was inferring stock & anyclass here, so I just made it explicit.

fission-web-server/library/Fission/Web/Server/Types.hs
walkah: @expede yeah, I actually remember these from the relay.

nix/commands.nix
```diff
@@ -82,16 +84,38 @@
       in
         [helper] ++ packages;
 
+    server-path = ~/.local/bin/server;
+    server-port = 10235;
+
+    # This will be much better when we have a nix-build
+    server-install = cmd "Install the Fission Server" "${stack} install fission-web-server:server";
+    server-start   = cmd "Run the currently installed Fission Server"   "DEBUG=true nohup ${server-path} &";
+    server-debug   = cmd "Run the Fission Server in debug verbose mode" "DEBUG=true ${server-path}";
```
walkah, line 92: 🙌

walkah (approved):
tenor-80811088

PR #466, opened 2021-03-09
Update whoami error message

bgins:
## Summary

Update error message to recommend the user to delete `~/.config/fission` instead of `~/.ssh/fission`.

## After Merge


walkah (approved):
tenor-94618203

Good catch! I'm probably looked at this 100 times and never recognized that the path is wrong 😊

fission-cli/library/Fission/CLI/Handler/User/Whoami.hs
```diff
@@ -67,5 +67,5 @@ whoami = do
               DecodeFailure _ _ -> "Trouble decoding the registration response."
               _                 -> "Invalid content type."
 
-      UTF8.putText "Please contact Fission support at https://fission.codes or delete `~/.ssh/fission` and try again."
+      UTF8.putText "Please contact Fission support at https://fission.codes or delete `~/.config/fission` and try again."
```
expede, line 70: This is great 👏 And to echo James — great catch!

You know, we could go one step further and hook it into our standard paths and display that:

https://github.com/fission-suite/fission/blob/5d0a808adc3e3cffb1c4eae17ab1f5194bc9814e/fission-cli/library/Fission/CLI/Environment/Class.hs#L7

We keep the standard paths inside the config in [`Fission.CLI.Environment.Path`](https://github.com/fission-suite/fission/blob/5d0a808adc3e3cffb1c4eae17ab1f5194bc9814e/fission-cli/library/Fission/CLI/Environment/Path.hs) module

bgins: Yeah, makes sense to pull this from a standard path. 

I'll give this a try. Thanks for the hints. :slightly_smiling_face:

bgins: Alright, a standard path for `globalDir` is added and we are using it in the error message. 

Right now, it will display the absolute path to the directory. Should consider displaying a relative path off of `~` instead?

expede: I agree that the `~` version would be prettier, but I'm not really sure how we would detect that and replace it, especially across different OS layouts (e.g. `/Users/expede` vs `/home/expede`). Having it update automatically is more important _for now_, so I think ship it with absolute paths and we can revisit if needed 🚀 

Thanks for the quick turn around on that 👏

bgins: Sounds good! :+1:

fission-cli/library/Fission/CLI/Environment/Path.hs
```diff
@@ -14,6 +15,11 @@ import           Fission.Prelude
 
 import           Fission.CLI.Environment.Class
 
+globalDir :: MonadEnvironment m => m FilePath
+globalDir = do
+  global <- getGlobalPath
+  return global
```
expede, line 21: Two questions:

1. Why this function at all? Naming? If so, should we rename the typeclass method?
2. This is equivalent to `globalDir = getGlobalPath`

bgins: Good point. The naming of `getGlobalPath` makes sense.

My initial thought was that we wanted to import all paths from `Fission.CLI.Environment.Path`, but less text to read does seem better here. :+1: 

Will push a change to update this to use `getGlobalPath` directly.

bgins: Updated this to reexport the typeclass method. This gives us the benefit of importing it off `Fission.CLI.Environment.Path` without adding an extra function.

expede (approved):
Looking good

PR #468, opened 2021-03-11
Allow underscores

dholms:
## Problem
Today, users are restricted to non-underscored usernames and app names (subdomains).

## Solution
- Allow underscores in the validator
- disallow usernames that _start_ with underscores so that we don't have conflicts with `_did`, `_dataroot`, `_dnslink`, etc

### DIFF
--- a/fission-core/library/Fission/URL/Validation.hs
+++ b/fission-core/library/Fission/URL/Validation.hs
@@ -23,7 +23,7 @@ import qualified Fission.Security as Security
 -- True
 --
 -- >>> isValid "under_score"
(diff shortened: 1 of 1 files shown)


dholms:
Should we allow usernames that end with an underscore? I don't immediately see a reason why not :thinking:

expede:
> Should we allow usernames that end with an underscore? I don't immediately see a reason why not 🤔

The username ends up as a subdomain (`boris.files.fission.name`), so it's the same problem in the end

expede (approved):

PR #469, opened 2021-03-12
Update Niv sources

expede:


shell.nix
```diff
@@ -40,7 +40,6 @@ let
     ];
 
     haskell = [
-      unstable.haskell-language-server
```
walkah, line 43: `LSP :: The following servers support current file but do not have automatic installation configuration: lsp-haskell` 

but was it working for you?

expede: Weird. Yeah, I have `lsp-haskell` via the `doom-emacs` module.

expede: From the `lsp-haskell` [docs](https://github.com/emacs-lsp/lsp-haskell#emacs-configuration):

> lsp-mode automatically requires the lsp-haskell package , so you do not need to require lsp-haskell unless you like being explicit.

expede: 🤦‍♀️ ZOMG no, I have it in my `home-manager`

expede: Also, landing soon:

> haskell-language-server: 0.9.0 -> 1.0.0 

https://github.com/NixOS/nixpkgs/commit/3c81fd41eec6e285c1e0e3a17fa2df39d74e911c

expede:
Step 1: Put HLS back in as a dependency
Step 2: Watch it compile forever
Step 3: Profit?

I really do wonder if it's just easier to manually grab from the releases section for now:

Screen Shot 2021-03-12 at 4 10 03 PM

expede:
> /home/runner/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_3.2.1.0_ghc-8.10.4: startProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

<Internal screaming intensifies>

PR #472, opened 2021-03-15
Set up haskell.nix

expede:
Attempting to build on EC2 and it's taking a _shocking_ amount of time and resources. Will be worthwhile if we can get it to make MUSL binaries, but not great for the base case.

UPDATE: increased the EC2 stats to something closer to what my laptop has. Is moving along much better now.

Finished building fission-core 👇 

Screen Shot 2021-03-15 at 12 22 57 AM

...and _in theory_ this is the CLI cross compiled for RPi:

Screen Shot 2021-03-15 at 12 57 52 AM

I've made a bunch of changes on the EC2 instance to make this work, but **ONCE YOU KNOW WHAT TO DO** it's pretty straightforward.


expede:
Superceded

expede:
Replaced by #484

PR #473, opened 2021-03-15
Fix typo "Fisson" in shell.nix

JonathanLorimer:
## Summary

This PR fixes a small typo that was bothering me when in a nix-shell


expede:
Thanks @JonathanLorimer 🙌 Good catch

@bmann will appreciate that typo. When we originally set up the company, there was a typo on the form so we were "Fisson" 😛

PR #476, opened 2021-03-27
UCAN: alg: "Ed25519" -> alg: "EdDSA"

expede:
Use the `alg: EdDSA` in Ed25519 UCANs (per JWT spec). Still accept Ed25519 for historical UCANs.

### DIFF
--- a/fission-core/library/Fission/Key/Asymmetric/Algorithm/Types.hs
+++ b/fission-core/library/Fission/Key/Asymmetric/Algorithm/Types.hs
@@ -21,8 +21,7 @@ instance Arbitrary Algorithm where
 instance ToJSON Algorithm where
   toJSON = String . \case
     RSA2048 -> "RS256" -- Per the JWT Spec (RFC 7519)
-    Ed25519 -> "Ed25519"
-    -- TODO Ed25519 -> "EdDSA"
+    Ed25519 -> "EdDSA"
 
 instance FromJSON Algorithm where
   parseJSON = withText "JWT.Algorithm" \case
--- a/fission-cli/package.yaml
+++ b/fission-cli/package.yaml
@@ -1,5 +1,5 @@
 name: fission-cli
-version: '2.10.0.7'
+version: '2.10.1.0'
 category: CLI
 author:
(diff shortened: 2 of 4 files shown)


fission-core/library/Fission/Key/Asymmetric/Algorithm/Types.hs
```diff
@@ -21,8 +21,7 @@ instance Arbitrary Algorithm where
 instance ToJSON Algorithm where
   toJSON = String . \case
     RSA2048 -> "RS256" -- Per the JWT Spec (RFC 7519)
-    Ed25519 -> "Ed25519"
-    -- TODO Ed25519 -> "EdDSA"
+    Ed25519 -> "EdDSA"
```
expede, line 24: @walkah this is mostly us failing to read the updated JWT spec a year ago

expede: You'll notice that on line 30 we maintain backwards compat when reading these in, but not when creating new UCANs.

icidasset:
Wait, `EdDSA` is the same as `Ed25519`? I thought they used different curves?

expede:
> Wait, `EdDSA` is the same as `Ed25519`? I thought they used different curves?

From Wikipedia:

> Ed25519 is the EdDSA signature scheme using SHA-512 (SHA-2) and Curve25519

I do need to double check that we're `SHA-512`ing the signature, but yeah it's Ed25519.

expede:
> Wait, `EdDSA` is the same as `Ed25519`? I thought they used different curves?

Oh, you may be thinking of ECDSA, which is different, and available in the WebCrypto API. That's different yeah.

icidasset:
Ohhh okay, thanks! Didn't know there was both eCdsa and eDdsa, the one letter difference tripped me up 😅

expede:
> I do need to double check that we're SHA-512ing the signature, but yeah it's Ed25519.

Yeah, it's `SHA-512`ing the content first 👍

walkah (approved):
🚢 👍  looks good to me!

PR #479, opened 2021-03-30
Implement CLI linking UI feedback

### LINKED ISSUE #478: CLI Linking feedback
Verbose mode, spelling of bearer token: `Decrypted still-serialized brearer token:`

When prompted with `Do you have an existing account? [Y/n]`, I believe convention is that capitalized item is selected if you hit return.

After asking for username, there is no prompt to the user that says "open auth.fission.codes to confirm codes and link your account". Here's where we could attempt to pop a browser directly, but that's a future feature. For now, have to put a message in for sure. The one time I didn't have the browser window already open, it also didn't work (non verbose mode, so no idea what was going on ... I force quit and ran again).

Installing dependencies in non verbose mode takes about 2 minutes. Most of this time is waiting for the IPFS node to boot up for the first time (it can take a while on publish as well). Should do messages to the user like:
* Installing dependencies
* IPFS server downloaded
* Setting config for IPFS server
* Starting IPFS server for the first time

Or, just keep adding to the dots ...............

Confirmation code and linking was very quick and felt like it completed instantly. Woot!

expede:
Closes #478 

* [x] Installing dependencies
* [x] IPFS server downloaded
* [x] Setting config for IPFS server
~~* [ ] Starting IPFS server for the first time~~ Not required, no longer start the daemon in setup

### DIFF
--- a/.github/workflows/artifacts-nix.yaml
+++ b/.github/workflows/artifacts-nix.yaml
@@ -31,7 +31,7 @@ jobs:
       name: 🗄️  Cache ~/.stack
       with:
         path: ~/.stack
-        key: ${{ matrix.os }}-stack-nix
+        key: ${{ matrix.os }}-stack-nix-2
 
     - name: 🏗️  Build
       run: |
--- a/.github/workflows/ci-nix.yaml
+++ b/.github/workflows/ci-nix.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os:
-          - macos-10.15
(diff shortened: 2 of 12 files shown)


expede:
Screen Shot 2021-03-29 at 11 26 19 PM
cc @bmann

bmann:
@expede lovely emoji. Also, opening the site (which will show you your username) and then entering username is a nice flow tweak.

star-trek-tng-worf-appeal-of-this-program

PR #485, opened 2021-04-08
Enforce utf-8 encoded output

matheus23:
This gets rid of errors like these for me:
```
$ fission setup -v
2021-04-08 18:49:42.069349: [debug] fission: <stderr>: hPutChar: invalid argument (invalid character)
```

Maybe this even gets rid of similar errors when running in some nix-shell environments (I remember us having to workaround those a little), but I haven't tested that.


expede (approved):
Amazing 👍  

For posterity:

> GHC creates new file handles with the encoding taken from getLocaleEncoding. All we have to do is call setLocaleEncoding utf8! Now all files opened in text mode will automatically use UTF-8
[Source](https://serokell.io/blog/haskell-with-utf8)

expede:
@matheus23 also please do a minor version number bump for the CLI 📈 🙏

PR #486, opened 2021-04-09
Ed25519-based DID Serialization Update

expede:
This is largely motivated by getting EdDSA support in the browser. The existing EdDSA DID format is a bit of a pain to decode, _and_ it turns out that it's deprecated. This PR includes the following changes:

* [x] Updated DID EdDSA format (all base58 rather than base64-in-base58)
* [x] Backwards compatibility with previous encoding on ingestion
* [x] The ability to track string encoding in code at the type-level
* [x] Bump Stackage to LTS-17.9


fission-cli/library/Fission/CLI/Environment.hs
```diff
@@ -189,7 +189,7 @@ fetchServerDID fissionURL = do
     Right (didTxt : _) ->
       case eitherDecodeStrict ("\"" <> didTxt <> "\"") of
         Left errs -> do
-          CLI.Error.put errs "Unable to find Fission's ID online"
+          CLI.Error.put errs "Unable parse Fission's DID from DNS"
```
expede, line 192: Distinguish between these cases more easily

fission-cli/library/Fission/CLI/Internal/ServerDID.hs
```diff
@@ -8,4 +8,4 @@ import           Fission.User.DID.Types
 
 fallback :: DID
 fallback =
-  fromJust $ decode "\"did:key:zStEZpzSMtTt9k2vszgvCwF4fLQQSyA15W5AQ4z3AR6Bx4eFJ5crJFbuGxKmbma4\""
+  fromJust $ decode "\"did:key:z2DSW536bcWxPGuz7ZMnXdju64pBoWrybTyzTqWYWa7EjsB\""
```
expede, line 11: Switching tests to use the new encoding

fission-core/library/Fission/Internal/RSA2048/Pair/Types.hs
```diff
@@ -21,6 +21,7 @@ data Pair = Pair
 instance Arbitrary Pair where
   arbitrary = elements pregenerated
 
+-- FIXME probably drop this
```
expede, line 24: Need to look at this still

.gitignore
```diff
@@ -3,6 +3,7 @@
 # Edit at https://www.gitignore.io/?templates=linux,macos,haskell,windows
 
 nohup.out
+result
```
expede, line 6: Result of building CLI with Nix expression

fission-core/library/Fission/Text/Encoded/Types.hs
```diff
@@ -0,0 +1,35 @@
...
+import           Fission.Text.Encoding.Types
+
+-- | Tag some text representation with the base it's encoded with
+--   NOTE that this is by no means a perfect method, and is mostly to make
+--   tracing encoding through code less painful. Character encoding can always
+--   be forgotten, mislabelled, and so on.
+newtype (enc :: Encoding) `Encoded` (carrierText :: Type)
+  = Encoded { encoded :: carrierText }
+  deriving stock   (Eq, Ord, Show)
+  deriving newtype (Semigroup, Monoid)
```
expede, line 18: Wait, `carrierText :: Type`? There's some light type-level stuff happening here. This is all to keep the checks at compile-time, prevent accidentally concatenating text in different bases, and so on.

If there's interest in learning about type-level programming, we can have a lunch & learn. In the meantime, just can absolutely use it as a magic interface that looks like:

```haskell
foo :: 'Base58_BTC `Encoded` Text
foo = Encoded "abcdef"

bar :: 'UTF8 `Encoded` Text
bar = Encoded "12345"

foo <> foo -- Encoded "abcdefabcdef"
foo <> bar -- Compiler error, Because 'UTF8 is not 'Base58_BTC
```

fission-web-server/env.yaml.example
```diff
@@ -22,7 +22,7 @@ storage:
     database: web_api
 
 auth:
-  fission_did: did:key:zStEZpzSMtTt9k2vszgvCwF4fLQQSyA15W5AQ4z3AR6Bx4eFJ5crJFbuGxKmbma4
+  fission_did: did:key:z2DSW536bcWxPGuz7ZMnXdju64pBoWrybTyzTqWYWa7EjsB
```
expede, line 25: Same key but in updated format

fission-core/library/Fission/Text/Encoded/Class.hs
```diff
@@ -0,0 +1,29 @@
+module Fission.Text.Encoded.Class (ToEncoded (..)) where
+
+import           RIO
+import qualified RIO.ByteString.Lazy as Lazy
+
+import           Fission.Text.Encoded.Types
+import           Fission.Text.Encoding.Types
+
+-- | A helper for beinging some native text into the Encoded type and tagging
```
matheus23, line 9: Typo "beinging/bringing" but :man_shrugging:

fission-core/library/Fission/Text/Encoded/Class.hs
```diff
@@ -0,0 +1,29 @@
...
+class ToEncoded txt where
+  type NativeEncoding txt :: Encoding
+  trackEncoding :: txt -> NativeEncoding txt `Encoded` txt
+
+instance ToEncoded Text where
+  type NativeEncoding Text = 'UTF8
+  trackEncoding txt = Encoded txt
+
+instance ToEncoded String where
+  type NativeEncoding String = 'Octal
```
matheus23, line 20: I find it weird to think of native encoding here. Is it really octal?

I think of encodings like restrictions on what bytes ranges or byte sequences can appear in a byte string. E.g. binary means bytes can only ever be "0" or "1" (in ascii?), and base64 means only A-Za-z0-9 and some other ascii bytes.

And then by extension wouldn't octal mean that only 0-7 ascii would be accepted bytes in the encoding?

I'm thinking there should be something akin to "all bytes can appear" encoding type.

I might be misunderstanding the interpretation of `Encoded` though, so please let me know :)

fission-core/test/Fission/Test/Internal/UTF8.hs
```diff
@@ -13,16 +13,16 @@ spec =
   describe "Fission.Internal.UTF8" do
     describe "toBase58Text" do
       it "converts regular text" $
-        toBase58Text "hello world" `shouldBe` "StV1DL6CwTryKyV"
+        toBase58Text "hello world" `shouldBe` Encoded @'Base58_BTC "StV1DL6CwTryKyV"
```
matheus23, line 16: Oh it reads so nicely :star_struck:

fission-web-server/env.yaml.example
matheus23: I'll have to update this in my pizza instance, right?

matheus23 (approved):

fission-core/library/Fission/Text/Encoded/Class.hs
expede: Ah, you're right — I've seen c2w around, but it does truncate. I'll change that to `Unicode` 👍

expede: > I might be misunderstanding the interpretation of Encoded though, so please let me know :)

Yes, possibly. We're saying that "despite the ability to express whatever native characters, you're only using the (e.g.) ASCII character set". We need a way to lift an untagged string type to a tagged version, and show its native encoding. `ByteStrings` are not UTF8 safe, but `Text` is, for instance. Both of them can be used to represent ASCII. Some encodings don't map cleanly onto each other, like Base58-BTC, or Base64 and Base64-URL.

expede:
@matheus23 I can't reply direct for... reasons? So reply here.

> I'll have to update this in my pizza instance, right?

It's probably a good idea. This code will take the new or legacy encoded DIDs just fine, but will only output new style DIDs. I'd recommend updating the one on your `.pizza` instance after the CLI changes fro this PR are published. We're going to hold off on production for a week or two, to give folks a change to upgrade their CLI (since `2.10.2` can't read the new style DIDs yet)

fission-core/library/Fission/Text/Encoded/Class.hs
expede: hahhahaa and I'm the native English speaker! I fat finger a _lot_ of typos these days. Thanks for catching that!

fission-core/library/Fission/Text/Encoded/Class.hs
expede: > A chinese character:      汉
> it's unicode value:       U+6C49
> convert 6C49 to binary:   01101100 01001001
> embed 6C49 as UTF-8:      11100110 10110001 10001001

😫

expede:
```haskell
data Text = Text
    {-# UNPACK #-} !A.Array          -- payload (Word16 elements)
    {-# UNPACK #-} !Int              -- offset (units of Word16, not Char)
    {-# UNPACK #-} !Int              -- length (units of Word16, not Char)
    deriving (Typeable)
```

Cool cool cool cool cool

expede: https://guide.aelve.com/haskell/strings-o62hqc69

> Text
> [...]
> Uses UTF-16

> String
> [...]
> Doesn't support Unicode perfectly (if you do something like map toUpper, for instance)

expede: Hmm okay, I think I was also letting my Elixir show a bit with `type NativeEncoding ByteString = 'Octal`. That's true of an _unpacked_ `ByteString` (i.e. `[Word8]`) in Haskell, but `ByteString` as a text representation is really ASCII by default. We don't actually use binary, hex, or octal, so I've dropped them in favour of ASCII, Unicode, and so on.

expede:
Screenshot of the quick & dirty UCAN validator working (first one valid, next several I intentionally broke). We could make this way nicer, but works well for now 😉 

Screen Shot 2021-04-09 at 7 30 10 PM

NOTE! The content type MUST be `text/plain;charset=utf-8`, with the `utf-8`. (On topic, given the rest of this PR).

fission-core/library/Fission/Text/Encoded/Class.hs
matheus23: Oh wow I totally missed this train of thought when scrolling through here! :D

Anyway, I like the changed version better :)

fission-core/library/Fission/Internal/Fixture/Bearer/Nested.hs
```diff
@@ -37,7 +37,7 @@ instance Proof.Resolver InTimeBounds where
 instance ServerDID InTimeBounds where
   getServerDID = pure did
     where
-      Right did = eitherDecode "\"did:key:zStEZpzSMtTt9k2vszgvCwF4fLQQSyA15W5AQ4z3AR6Bx4eFJ5crJFbuGxKmbma4\""
+      Right did = eitherDecode "\"did:key:z2DSW536bcWxPGuz7ZMnXdju64pBoWrybTyzTqWYWa7EjsB\""
```
dholms, line 40: Could we move this & the hardcoded DID in `CLI/Internal/ServerDID` to a shared location so that we only have to update it once in future scenarios? Maybe `Authorization/ServerDID`?

fission-core/test/Fission/Test/Internal/UTF8.hs
dholms: :point_up_2:

dholms (approved):
this all looks good to me :raised_hands: 

Just the one point about hardcoding the server did in only 1 location

Also I just want to make sure I remembering & interpreting the multicodecs repo correctly: since ed25519 (and in my case, bls12-381) keys are fixed size, we don't need to do any middle-endian funny business right? Just z + prefix + key bytes

fission-core/library/Fission/Internal/Fixture/Bearer/Nested.hs
expede: @dholms this isn't where we get the _actual_ value — that comes from the `env.yaml`. `InTimeBounds` is a test and development monad. This never gets called at production runtime.

expede:
> we don't need to do any middle-endian funny business right? Just z + prefix + key bytes

I _don't think so_... I'm pretty sure that's all in the RSA key, which is unchanged in this PR, and not added to the Multicodec spec. (Maybe we should add that, btw 😅 I'm serious)

I looked at the current WIP W3C DID spec, and it's, yeah, confusing:

https://w3c-ccg.github.io/did-method-key/#ed25519-x25519

> did-key-format := did:key:MULTIBASE(base58-btc, MULTICODEC(public-key-type, raw-public-key-bytes)) 

Okay, sure. And in the multicodec table, Ed25519 is `0xed`. So let's look at the W3C test vectors:

```
did:key:z6MkiTBz1ymuepAQ4HEHYSF1H8quG5GLVVQR3djdX3mDooWp
did:key:z6MkjchhfUsD6mmvni8mCdXHw216Xrm9bQe2mBH1P5RDjVJG
did:key:z6MknGc3ocHs3zdPiJbnaaqDi58NGb4pk1Sp9WxWufuXSdxf
```

When we deserialize those:

```
ed013b6a27bcceb6a42d62a3a8d02a6f0d73653215771de243a63ac048a18b59da29
ed014cb5abf6ad79fbf5abbccafcc269d85cd2651ed4b885b5869f241aedf0a5ba29
ed017422b9887598068e32c4448a949adb290d0f4e35b9e01b0ee5f1a1e600fe2674
```

Well, that `01` is back (after the `ed`)! Where did it come from? I have no idea! It's not in the Multiformats chart, and not in their format (above). We did have it before, so maybe it's listed elsewhere? 🤷‍♀️ 

Anyhow, our keys match theirs in format now. I'm going to do some further reading to double check about that `01`, but I _think_ this is right now. I dunno, the docs are pretty bad 😛

fission-core/library/Fission/Internal/Fixture/Bearer/Nested.hs
dholms: ah got it got it :ok_hand:

dholms:
> But then they merged just 0xed. In any case, we want to be did:key compatible, which adds that 01. The current spec and implementations from (e.g.) Digital Bazaar all include it. So that's what we're also going to have 

Wait.... doesn't this show that our keys _do not_ match them in format now? The w3c documentation says "These DID always start with `z6Mk`", which ours do not :thinking: 

So some quick bit calculations: `0xed` when varint encoded is `0xed01`. Which is confusing because the encoding in this case just tacks on a `0x01`. 

So I think the table is listing the non-varint encoded prefixes, but when added to `did:key` they need to be varint encoded. Which seems.... silly. and prone to error but :man_shrugging: 

We merged in just `0xed` as the prefix, but after looking at this again, it seems that we should be prefixing with `0xed01`

expede:
@dholms 

> We merged in just 0xed as the prefix

We did not. For example [here](https://github.com/fission-suite/fission/pull/486/files#diff-fb96f9a336f1e109568335dcbbabb204145bf6e9a8fc3cd42590668c7c5499b5R116) it is `0xed : 0x01 : ...`

> "These DID always start with z6Mk", which ours do not 

Ours _do_ match this; there were commits since the last time you looked at the PR

> when varint encoded is 0xed01. Which is confusing because the encoding in this case just tacks on a 0x01.

Ooooh right this is the whole bit shuffling stuff we dealt with last time

> So I think the table is listing the non-varint encoded prefixes

Yeah, that table really should just include both versions. Might be a good PR

PR #487, opened 2021-04-09
remove `user register` command

walkah:
`user register` isn't terribly useful as a standalone command. User registration is handled during `setup` and won't work without running `setup` first, so removing it to avoid confusion.


expede (approved):
Looks awesome 👍 👏

PR #488, opened 2021-04-09
Fix DELETE /app/:id requests

matheus23:
It seems like at some point the AWS API must have changed. Before I was getting these errors from amazonka:
```
Invalid request: Expected exactly one 
[ AliasTarget
, all of
    [ TTL
    , and ResourceRecords
    ]
, or TrafficPolicyInstanceId
]
, but found none in Change with
    [ Action=DELETE
    , Name=matheus23-web-test.fission.pizza
    , Type=TXT
    , SetIdentifier=null
    ]
```

The [docs](https://docs.aws.amazon.com/Route53/latest/APIReference/API_ChangeResourceRecordSets.html#API_ChangeResourceRecordSets_Errors) say:
> To delete a resource record set, you must specify all the same values that you specified when you created it.  

Hence this change.


expede (approved):
LGTM! 👍

expede:
@matheus23 also please bump the minor version number

matheus23:
For posterity. This is it working on the pizza instance:
image

matheus23:
That one Artifacts build seems to consistently fail for some reason. Should I / can I just merge anyway?

expede:
https://github.com/fission-suite/fission/pull/488#issuecomment-817798674

Yeah, go wild. I don't know why, but sometimes macOS GH actions fail, and even cache busting doesn't fix. But then later they work again 🤷‍♀️ 

I'm on a mac, works on my machine™️

Issue #489, opened 2021-04-11
WNFS data root updates can take > 60s with _some accounts._


expede:
### Reproduced By
* Brooke
* Philipp

### Environment
OSes: Arch, macOS
Browsers: Firefox, Safari

### Ruled Out

* [ ] IPFS Cluster
* [ ] Web API
* [ ] DNS
* [ ] SDK
* [ ] Application code

expede:
Issue occurs across all browsers and OSes, but only _some accounts._ It's consistent with those ones.

This _looks like_ something happening on the server-side, though we still need to validate that the request is _actually_ sent and not blocked by a service worker or similar.

---

From conversation, we had a hypothesis that it could be the UCAN validation when it's deeply nested. This doesn't appear to be the case from benchmarks, so we can rule that out:

```
    RSA-based UCAN checks
      One   Level:     OK (1.77s)
         53 μs ± 4.1 μs
      Two   Levels:    OK (4.70s)
         72 μs ± 3.8 μs
      Three Levels:    OK (3.93s)
        118 μs ± 3.8 μs
```

matheus23:
Hypothesis: This is related to general file system size.

I have this hunch, because there *should* be lots of history entries, due to lots of `publish` calls (which were automatically called in flatmate when any change occurred).

Not sure how to test this best. Here are a couple of ideas:
1. Create a new account and spam publish calls and changes and log the time PATCH requests take. Hypothesis should be confirmed when we see a steady increase in PATCH request times.
2. Log ipfs request performance server-side
3. Copy over the matheus23-drive file system (by grabbing the DNSLink CID) & read key and test another account to see if it still happens. Reasoning for this is: This is very likely something happening on the server. And the only account we can reproduce this with at the moment lives on production and we want to reproduce this somewhere else ideally. Also it double-checks that this is nothing related to the particular account.
4. Try pulling the account data and looking at the actual size? We have no idea how big it *actually* is: IPFS only gives us the inflated 533 Petabytes number.
5. Override the user's file system with an empty one (Ideally we keep a backup of the old one locally)

---

> This looks like something happening on the server-side, though we still need to validate that the request is actually sent and not blocked by a service worker or similar.

I just tested that quickly by bypassing all server-worker related things in chromium, no changes. It seems unrelated to service workers.

[...]

PR #490, opened 2021-04-11
Add RSA-2048 / UCAN benchmarks

expede:


expede:
Wrote some benchmarks to support investigation into https://github.com/fission-suite/fission/issues/489

matheus23 (approved):

Issue #491, opened 2021-04-12
Discussion: How we handle 'impure' exceptions in the server's route handlers


matheus23:
Something I noticed when writing #488.

This issue is a call for discussion on how and where to best implement a 'safety-net' for impure exception that might occur 'during' route handlers.

### What are impure exceptions?

I'm not 100% sure how to best call them, so I'll explain what I mean shortly:

A.k.a. "IO Exceptions", because they can be caught in the IO. However, they can be triggered everywhere, also outside IO, e.g. `1 / 0` will trigger an impure exception - or any other partial function.

### How do we handle them?

Catching impure exceptions in haskell is tricky because of laziness.

Something like this won't actually catch the exception:

```hs
> let x = return (1 `div` 0) `catch` handler in ()
```
Similarly, if you only use the value beyond your 'catch' safety-net, then the handler won't run.
More info: https://stackoverflow.com/questions/11149876/how-do-laziness-and-exceptions-work-together-in-haskell

In our case this means that if we try to catch exceptions in handlers outside where response bodies are forced, we might not actually catch them all (no pun intended).

### What happens when an impure exception isn't caught?

If a request generates an exception, it'll be (at the very latest) caught in [warp's exception handler](https://hackage.haskell.org/package/warp-3.2.23/docs/Network-Wai-Handler-Warp.html#v:setOnException).
However, this will bypass all our middlewares. So that's where our CORS issues stem from: We get internal server errors -> warp generates a basic response for them ([this one](https://hackage.haskell.org/package/warp-3.2.23/docs/Network-Wai-Handler-Warp.html#v:defaultOnExceptionResponse), to be exact), but it doesn't include any CORS headers, so it'll fail to appear on the user side.

### How should we catch impure exceptions?

You might say "just don't ever throw impure exceptions! Never write partial functions or divide by zero!".
Well, I think that's really hard, given you don't have control over all code. E.g. `amazonka` can throw exceptions. What we really want is a safety-net.

So there's two way I can think of to fix this:
1. We force response bodies in a middleware and try to catch everything there is (using `deepseq`).
2. Adapt warp's `setOnException` setting to include CORS headers. But this might bite us in the long run, if we expect our middlewares to always run.

I'm interested about your thoughts.

expede:
Fixed by #492

PR #492, opened 2021-04-12
Add CORS header to generic internal server error response

matheus23:
This is implementing option 2 from #491

I don't expect this to be merged, this is just me trying to see what this solution looks like.

This gets rid of unexpected CORS errors on the front-end when there's an internal server error. So instead of a CORS error it now looks like this in front-ends:

image

Another benefit is that anything that's creating a bad request will now also return with CORS headers.

I'm pretty sure having an accept-anything CORS header is fine here, as the response will never reveal any information, only a status code + a "Bad Request" or "Something went wrong" body.


fission-web-server/library/Fission/Web/Server/Internal/Production.hs
```diff
@@ -215,6 +216,10 @@ mkSettings logger (Server.Port port) =
   defaultSettings
     |> setPort port
     |> setLogger (Web.Log.fromLogFunc logger)
+    |> setOnExceptionResponse
+      (mapResponseHeaders (("Access-Control-Allow-Origin", "*") :) 
+        . defaultOnExceptionResponse
+      )
```
expede, line 222: I also have a tendency for point-free, but outside of a few cases, adding the lambda makes this stuff much more legible. A named function is even better:

```haskell
-- Why isn't this in the WAI package? Seems useful
addResponseHeader :: HeaderName -> ByteString -> Response -> Response
addResponseHeader field val = mapResponseHeaders \headers -> (field, val) : headers

-- ...

|> setLogger (Web.Log.fromLogFunc logger)
|> setOnExceptionResponse \err -> addResponseHeader $ defaultOnExceptionResponse err
|> setTimeout serverTimeout
```

expede (approved):
Just the one suggestion, but otherwise LGTM 👍

expede:
Aaaaaaand as usual, please bump the version number before merging

fission-web-server/library/Fission/Web/Server/Internal/Production.hs
matheus23: oooooh I always forget that we have BlockArguments! That actually looks super clean!

matheus23: Ahh, that actually doesn't work :slightly_frowning_face: Apparently, it'll put the next `|>` into the lambda.

matheus23:
Closes #491

PR #500, opened 2021-04-19
Don't create an app on user creation

matheus23:
## Summary

Don't create a "sample" app on account creation.

At the moment this feature is not really that useful and mostly confusing to the users. "I never created an app, why is there one?"

## Test plan (required)

I set up the auth lobby on my pizza instance with this branch's code, created a new account in the lobby and ran this in the console:

image


expede (approved):
LGTM 👍

PR #501, opened 2021-04-19
Don't create an app if the domain name to associate with is already taken

### LINKED ISSUE #493: Creating an app with a domain name that was already taken creates a database row
## Problem

When I `POST https://runfission.com/app?subdomain=web-uploaded-app` (as an example) and it returns a 409 (Conflict). It will create a database entry for that app.

image

However, an app is actually created, as can be seen by the response from the app index request:

image

(See the app with ID 1610 that has no associated domains)

### Impact

I *think* we don't want this. The way we changed the `webnative.apps` API recently, these apps are basically dead. Apps with no domains associated are [filtered out on `webnative.apps.index()` calls](https://github.com/fission-suite/webnative/blob/bd1266d481dc90e8c42214fd17b963c86f11733c/src/apps/index.ts#L43).

## Solution

We "clean up" when there's an error, i.e. remove the app database row we just created. See this code snippet:

```haskell
-- code at the moment
appId <- runDB (App.createDB ownerId cid size now)

runDB (App.Domain.associateWithFallback ownerId appId maySubdomain now) >>= \case
  Left err ->
    -- remove the app by id?
    return $ Error.relaxedLeft err
-- ...
```

matheus23:
Fixes #493

### DIFF
--- a/fission-web-server/library/Fission/Web/Server/Types.hs
+++ b/fission-web-server/library/Fission/Web/Server/Types.hs
@@ -628,7 +628,8 @@ instance App.Creator Server where
         appId <- runDB (App.createDB ownerId cid size now)
 
         runDB (App.Domain.associateWithFallback ownerId appId maySubdomain now) >>= \case
-          Left err ->
+          Left err -> do
+            runDB (App.destroy ownerId appId now)
             return $ Error.relaxedLeft err
 
           Right subdomain -> do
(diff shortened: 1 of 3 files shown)


expede (approved):
Oh yeah, this totally makes sense.

Thanks!





Just rambling now, but this kind of cleanup may make sense to move this cleanup logic into a `MonadCleanup.onRaise`/`MonadMask.onError`, as the closest we can get to a DB transaction.

PR #503, opened 2021-04-23
🐛 Fix ipfsignore

expede:
# Changes

1. Typo in the `FromJSON` field name 😅 
2. Include the app-specific ignore in that one important scenario

# Proof of Fix

Screen Shot 2021-04-22 at 6 54 18 PM
Screen Shot 2021-04-22 at 6 54 32 PM
Screen Shot 2021-04-22 at 6 54 52 PM
Screen Shot 2021-04-22 at 6 54 59 PM

### DIFF
--- a/fission-cli/library/Fission/CLI/Environment/Types.hs
+++ b/fission-cli/library/Fission/CLI/Environment/Types.hs
@@ -42,7 +42,7 @@ instance ToJSON Env where
 instance FromJSON Env where
   parseJSON = withObject "Env" \obj -> do
     peers          <- obj .:  "peers"
-    ignored        <- obj .:? "ignored" .!= []
+    ignored        <- obj .:? "ignore" .!= []
     serverDID      <- obj .:  "server_did"
     signingKeyPath <- obj .:  "signing_key_path"
     username       <- obj .:  "username"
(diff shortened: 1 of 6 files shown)


fission-cli/library/Fission/CLI/Environment/Types.hs
```diff
@@ -42,7 +42,7 @@ instance ToJSON Env where
 instance FromJSON Env where
   parseJSON = withObject "Env" \obj -> do
     peers          <- obj .:  "peers"
-    ignored        <- obj .:? "ignored" .!= []
+    ignored        <- obj .:? "ignore" .!= []
```
expede, line 45: Typo 😅 Where are my dependent types at?

fission-cli/library/Fission/CLI/Parser/Config/IPFS.hs
```diff
@@ -49,5 +49,5 @@ timeoutParser =
     , long "ipfs-timeout"
     , metavar "SECONDS"
     ----------
-    , value 300
+    , value 1800
```
expede, line 52: Ran into this while testing the above with a large file. We have a more sophisticated timeout mechanism on the server now, so letting the CLI have a longer default timeout.

matheus23 (approved):
:bug: :no_entry_sign:

PR #504, opened 2021-04-25
Let user know when their email is invalid

### LINKED ISSUE #384: Let CLI user know when their email is invalid
NB: Feature requests will only be considered if they solve a pain

# Summary

## Problem

At time of writing, if email verification cannot be sent (e.g. malformed email address), we return a generic `HTTP 500`. It would be good to return a `422` and message about being unable to send the verification email. Also should probably revert the account, too.

### Impact

* Harder to debug
* Confusing for the user when it fails

## Solution

* 500 -> 422
* Descriptive message

erlandsona:
closes #384

### DIFF
--- a/fission-web-server/library/Fission/Web/Server/Email/Error.hs
+++ b/fission-web-server/library/Fission/Web/Server/Email/Error.hs
@@ -13,8 +13,8 @@ newtype CouldNotSend = CouldNotSend ClientError
            )
 
 instance Display CouldNotSend where
-  display (CouldNotSend _) = "An error occured while trying to send an email"
+  display (CouldNotSend _) = "We couldn't send the verification email, please check you've provided a valid email address."
 
 
 instance ToServerError CouldNotSend where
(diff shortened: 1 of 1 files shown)


expede (approved):
This is much simpler than what I had in my head 😆  I love it!

Pending CI:

expede:
Hmm don't worry about Cachix failing. We can figure that out separately 🤔

erlandsona:
Yeh sounds good... and no worries. It was actually a god confidence booster as a first issue to just find my way around the code base and knock something out. Granted, if you want what was in your head feel free to elaborate and I'll do my best to accommodate. 
As in the other PR, I didn't actually know how to test this but given the requirements it seemed straightforward enough :man_shrugging:

expede:
> Granted, if you want what was in your head feel free to elaborate and I'll do my best to accommodate.

This PR is great as-is. Being pragmatic is a virtue 👍

PR #506, opened 2021-04-25
Expose app creation/update on API

### LINKED ISSUE #385: Make additional app data from web API
@nathanmalishev is building a dashboard. Additional data about an app would be useful:

- creation date
- date of last update

erlandsona:
## Test plan (required)

Demonstrate the code is solid. Example: The exact commands you ran and their output, screenshots / videos if the pull request changes UI.

## Closing issues
closes #385 

## After Merge
* [x] Does this change invalidate any docs or tutorials? _If so ensure the changes needed are either made or recorded_

### DIFF
--- a/fission-core/library/Fission/App/Types.hs
+++ b/fission-core/library/Fission/App/Types.hs
@@ -1,6 +1,6 @@
 module Fission.App.Types
   ( module Fission.App.Name
-  , App(..)
+  , Payload(..)
   ) where
 
 import           Data.Swagger     hiding (URL, url)
@@ -11,7 +11,7 @@ import           Fission.App.Name
 import           Fission.URL
 
 
-data App = App
+data Payload = Payload
   { urls       :: [URL]
   , insertedAt :: UTCTime
   , modifiedAt :: UTCTime
@@ -22,22 +22,22 @@ data App = App
 
 
 
-instance FromJSON App where
-  parseJSON = withObject "App" \obj -> do
+instance FromJSON Payload where
+  parseJSON = withObject "Payload" \obj -> do
     urls       <- obj .: "urls"
     insertedAt <- obj .: "insertedAt"
     modifiedAt <- obj .: "modifiedAt"
 
-    return App {..}
+    return Payload {..}
 
-instance ToJSON App where
-  toJSON App {..} = object
+instance ToJSON Payload where
+  toJSON Payload {..} = object
     [ "urls"       .= urls
     , "insertedAt" .= insertedAt
     , "modifiedAt" .= modifiedAt
     ]
 
-instance ToSchema App where
+instance ToSchema Payload where
   declareNamedSchema _ = do
     urls'       <- declareSchemaRef (Proxy :: Proxy [URL])
     insertedAt' <- declareSchemaRef $ Proxy @UTCTime
@@ -52,7 +52,7 @@ instance ToSchema App where
            ]
       |> required .~ ["username", "email"]
       |> description ?~ "Properties for a registered application"
-      |> example ?~ toJSON App
+      |> example ?~ toJSON Payload
         { urls       = [ URL
                             (DomainName "fission.codes")
                             (Just $ Subdomain "my-cool-subdomain")
@@ -60,5 +60,5 @@ instance ToSchema App where
         , insertedAt = fromSeconds 0
         , modifiedAt = fromSeconds 20
         }
-      |> NamedSchema (Just "App")
+      |> NamedSchema (Just "Payload")
       |> pure
--- a/fission-core/library/Fission/App/Types.hs
+++ b/fission-core/library/Fission/App/Types.hs
@@ -3,16 +3,16 @@ module Fission.App.Types
   , App(..)
   ) where
 
-import Data.Swagger hiding (URL, url)
+import           Data.Swagger     hiding (URL, url)
 
-import Fission.Prelude
+import           Fission.Prelude
 
-import Fission.App.Name
-import Fission.URL
+import           Fission.App.Name
+import           Fission.URL
 
 
 data App = App
-  { urls :: [URL]
+  { urls       :: [URL]
   , insertedAt :: UTCTime
   , modifiedAt :: UTCTime
   }
@@ -24,7 +24,7 @@ data App = App
 
 instance FromJSON App where
   parseJSON = withObject "App" \obj -> do
-    urls        <- obj .: "urls"
+    urls       <- obj .: "urls"
     insertedAt <- obj .: "insertedAt"
     modifiedAt <- obj .: "modifiedAt"
 
@@ -32,7 +32,7 @@ instance FromJSON App where
 
 instance ToJSON App where
   toJSON App {..} = object
-    [ "urls"        .= urls
+    [ "urls"       .= urls
     , "insertedAt" .= insertedAt
     , "modifiedAt" .= modifiedAt
     ]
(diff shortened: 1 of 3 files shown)


fission-core/library/Fission/App/Types.hs
```diff
@@ -0,0 +1,64 @@
...
+  }
+  -- What about deriving Json of normal data?
+  -- deriving (Eq, Generic, Show, ToJSON)
+  deriving (Eq, Show)
+
+
+
+instance FromJSON App where
+  parseJSON = withObject "App" \obj -> do
+    urls       <- obj .: "urls"
```
erlandsona, line 27: Pretty sure this is not accounting for serializing a `[URL]` type from json but went through a few blogs and couldn't find a `Decode.list obj "urls"` sort of thing, curious what the Aeson equivalent is?

fission-core/library/Fission/App/Types.hs
```diff
@@ -0,0 +1,64 @@
...
+instance ToJSON App where
+  toJSON App {..} = object
+    [ "urls"       .= urls
+    , "insertedAt" .= insertedAt
+    , "modifiedAt" .= modifiedAt
+    ]
+
+instance ToSchema App where
+  declareNamedSchema _ = do
+    urls'       <- declareSchemaRef (Proxy :: Proxy [URL])
```
erlandsona, line 42: Do type applications work for HKT's? like `Proxy @[URL]`?

fission-core/library/Fission/App/Types.hs
expede: Aeson has a polymorphic `instance (FromJSON a) => FromJSON [a]` ([source file](https://www.stackage.org/haddock/lts-17.10/aeson-1.5.6.0/src/Data.Aeson.Types.FromJSON.html)), so it'll deserialize lists automatically with no intervention required.

fission-core/library/Fission/App/Types.hs
expede: Yup! We had a bunch of `Proxy :: Proxy a`s kicking around initially, and switched to TYpeApplications a while ago, Guess we missed a few! Good catch. So yeah, this can be rewritten `Proxy @[URL]` 👍

fission-core/library/Fission/App/Types.hs
```diff
@@ -0,0 +1,64 @@
...
+
+
+data App = App
+  { urls       :: [URL]
+  , insertedAt :: UTCTime
+  , modifiedAt :: UTCTime
+  }
+  -- What about deriving Json of normal data?
+  -- deriving (Eq, Generic, Show, ToJSON)
+  deriving (Eq, Show)
```
expede, line 21: We may serialize the app in a bunch of different ways, because we're decoupling the inner representation from the serialized view. In Rails or Phoenix terms, these are the "model" and "view" layers.

I'd maybe put this in `Fission.Web.API.App.Detail.Payload.Types` or similar, since it's related to the specific endpoint, not all places we need `App`s.

expede: Oh, also something that you would have no way of knowing from just looking at the code: we're going to be reworking the app model in the near future (maybe even you'll get to it!) to have one canonical Fission-subdomain, and several "follow"ers (there's some IPFS magic in here that we get into later). If you can grab the first URL. Something like one of these two:

```haskell
data Payload = Payload
  { url :: URL
  , insertedAt :: UTCTime
  , modifiedAt :: UTCTime
  }
```

```haskell
data Payload = Payload
  { subdomain :: Subdomain
  , host :: Host
  , insertedAt :: UTCTime
  , modifiedAt :: UTCTime
  }
```

fission-web-server/library/Fission/Web/Server/Handler/App/Index.hs
```diff
@@ -23,10 +24,10 @@ index Authorization {about = Entity userId _} = runDB do
   let normalized = (\(k, v) -> (fromIntegral (fromSqlKey k), v)) <$> appXDomains
   return (Map.fromList normalized)
   where
-    findDomains :: App.Domain.Retriever m => Entity App -> m (AppId, [URL])
-    findDomains (Entity appId _) = do
+    findDomains :: App.Domain.Retriever m => Entity Models.App -> m (AppId, Core.App)
```
expede, line 27: See comment above about `App.Payload` vs `App`; it should also make this read better 👍

expede:
You were asking how to test this. I'd do a few things:

## 1. Test the Serializer

Most of that definition is declarative, so there's very little to actually test. Aeson will ensure that it's valid JSON. I would check that `eitherDecode . encode == Right`, since we've been bitten a few times by typos on field names.

## Test the Interface

Servant is extremely testable, but also so much of it is pushed into the type layer that you have proofs that it's going to at least conform to the API spec. Here's some docs from the Servant team on testing Servant in general: https://docs.servant.dev/en/release-0.15/cookbook/testing/Testing.html

## Test the Handler

The Servant handler must conform to the API spec, so we know that interface is correct if the API spec is. So, we want to test that the handler actually grabs what we expect and whatnot. We assume that the typeclasses do what they say (don't test dependencies, though they should be tested separately more than they currently are). You can then either write a monad that provides the correct class instances with mock values (i.e. don't actually hit the DB), or use the `Mock` type (which carries a bunch of info about which effects have been run, but this is often overkill).

So for example, on `Fission.Web.Server.Handler.App.Index`, we have:

```haskell
handler :: (MonadDB t m, App.Retriever t, App.Domain.Retriever t) => ServerT API.App.Index m

-- which is the same as

handler :: (MonadDB t m, App.Retriever t, App.Domain.Retriever t) => Authorization -> m (Map Natural [URL])
```

You can write a pure wrapper around that, where `runDB = id` `findDomains _ = mockResult` and so on.

---

For some existing tests, you can see a simple example at `Fission.Test.Web.Server.Ping`, and one with the effect check at `Fission.Test.Web.Server.Auth` (see `setup :: IO ...` starting on line 55). We're horrendously undertested at the moment, but do enough at the type level that portions are quite safe. More tests are always better though 🙏

erlandsona:
Okay tryin' to run the test suite and it's either hanging on the property tests cause I can't pass a sane default for --smallcheck-depth or --quickcheck-tests to make them run in a finite amount of time on my machine...

ERGO: what's the magic sauce to run the test suite locally or just how long is it supposed to take to complete?

expede:
> Okay tryin' to run the test suite and it's either hanging on the property tests cause I can't pass a sane default for --smallcheck-depth or --quickcheck-tests to make them run in a finite amount of time on my machine...
> 
> ERGO: what's the magic sauce to run the test suite locally or just how long is it supposed to take to complete?

It's expected to take around 10 minutes to run without adjusting anything. With Tasty, you can have it only run specific tests, and/or crank down the quickchecking. See the options here: https://hackage.haskell.org/package/tasty

That said, maybe we should write a `test-short` command! I'll give that a look.

It looks like it's all passing in the GitHub Actions tests.

expede:
Ah right, @matheus23 helped out with this in Discord. Copy/pasting some here for posterity:

> erlandsona — Yesterday at 07:27
> Well that definitely worked...
> --quickcheck-tests=1: passed: 3sec
> --quickcheck-tests=2: passed: 6sec
> --quickcheck-tests=3: passed: 13sec
> --quickcheck-tests=4: passed: 20sec
> In any case, I can run the suite now :thumbsup: Thanks!

erlandsona:
@expede as far as the Servant API is concerned, yeh I think think this is maybe mergable? 
I haven't gotten around to writing any actual tests for it yet tho.

expede:
@erlandsona looks good; thanks!

PR #511, opened 2021-05-10
Add EKG to server

expede:
[EKG is a tool for metrics](http://maxgabriel.github.io/ekg-yesod/). There's a bunch of packages for integrating with ElasticSearch and whatnot. I figured that it may be good for us to have some of these stats even crudely in panels for checking the production environment to find bottlenecks.

Use case: as we're bumping up the number of IPFS client HTTP connections originating from the server, I'd like to see if it's actually unblocking that previous bottleneck that Philipp seemed to run into with a large number of concurrent IPFS-related requests.

As with any instrumentation, this adds some overhead to our server. According to the docs, it's generally in the range of 200-600µs.

### DIFF
--- a/fission-web-api/library/Fission/Web/API/Auth/Types.hs
+++ b/fission-web-api/library/Fission/Web/API/Auth/Types.hs
@@ -12,8 +12,8 @@ import           Fission.Web.Auth.Token.Types
 
 import           Fission.Web.API.Prelude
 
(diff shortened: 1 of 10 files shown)


walkah (approved):
Zach Galifianakis Yes - lGmKzynHWftFm

🚀  works in my testing!

PR #514, opened 2021-05-17
Bump Stack to 2.7.1

expede:
@walkah rejoice for `In Nix shell but reExecL is False` is no more!

In other words, we don't _need_ to use all of the custom commands from inside Nix now. You can just `stack build`.

* Bump Stack to 2.7.1 (updated Nix packages in general)
* Enable Nix integration by default

---

NOTE: remember that you'll need to `lorri shell` to get the updated Stack

### DIFF
--- a/nix/commands.nix
+++ b/nix/commands.nix
@@ -20,7 +20,6 @@
           pkgs.writeScriptBin name ''
             #!${bash}
             echo "⚙️  Running ${name}..."
-            unset STACK_IN_NIX_SHELL
             ${script}
           '';
 
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,7 +1,5 @@
 resolver: lts-17.11
 
-  # allow-newer: true
-
 packages:
   - fission-cli
   - fission-core
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,6 +1,6 @@
-resolver: lts-17.10
+resolver: lts-17.11
 
-allow-newer: true
+  # allow-newer: true
 
 packages:
   - fission-cli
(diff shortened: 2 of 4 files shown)


nix/commands.nix
```diff
@@ -20,7 +20,6 @@
           pkgs.writeScriptBin name ''
             #!${bash}
             echo "⚙️  Running ${name}..."
-            unset STACK_IN_NIX_SHELL
```
expede, line 23: No longer needed

stack.yaml
```diff
@@ -27,6 +27,6 @@ ghc-options:
   "$everything": -haddock
 
 nix:
-  enable:     false
```
expede, line 30: Nix now enabled by default

stack.yaml
```diff
@@ -1,6 +1,6 @@
-resolver: lts-17.10
```
expede, line 1: Bumped resolver

stack.yaml
```diff
@@ -1,6 +1,6 @@
...
+resolver: lts-17.11
 
-allow-newer: true
+  # allow-newer: true
```
expede, line 3: Just waiting to see if it will actually build without this flag

walkah (approved):
Assuming the build still works... 🙌 

Jimmy Fallon Yes - ebQVo2Cl4Fo1YxoCHD

expede:
@walkah seems to work! I'll just let CI pass (new Stack, so may take a while). In the meantime, I'll fix the server DID stuff (this was on my working branch when I `cd`ed into the directory)

PR #515, opened 2021-05-17
Use oldstyle server DID when setting DNS _did

expede:
Still need to write a couple tests to make sure that the format is _actually_ what we expect, but in theory this is done

### DIFF
--- a/.github/workflows/artifacts-nix.yaml
+++ b/.github/workflows/artifacts-nix.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os:
-          - macos-latest
+          # - macos-latest
           - ubuntu-latest
 
     steps:
--- a/.github/workflows/artifacts-nix.yaml
+++ b/.github/workflows/artifacts-nix.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os:
-          # - macos-latest
+          - macos-latest
           - ubuntu-latest
 
     steps:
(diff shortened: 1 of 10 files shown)


fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -468,7 +470,7 @@ instance Server.DID.Publicize Server where
       ourURL         = URL (URL.DomainName . Text.pack $ baseUrlHost host) Nothing
       txtRecordURL   = URL.prefix' (URL.Subdomain "_did") ourURL
 
-      txtRecordValue = textDisplay did
+      txtRecordValue = textDisplay $ DID.Oldstyle did
```
expede, line 473: This is the main event

.github/workflows/artifacts-nix.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os:
-          - macos-latest
+          # - macos-latest
```
expede, line 16: Just fighting with GH Actions right now

stack.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: lts-17.11
+resolver: lts-17.12
```
expede, line 1: Damnit Stackage, stop releasing new things 🤣

walkah (approved):
Seth Meyers Thank You - 26gsjCZpPolPr3sBy
Will Ferrell Yes - 3ohzdIuqJoo8QdKlnW

PR #516, opened 2021-05-18
CLI: add --with-key option to setup

walkah:
This PR adds a `--with-key` option to `fission setup` that allows the CLI to bootstrap with a given key file (i.e. `machine_id.ed25519`). This is useful if you just back up your key file to move to a new machine or restore. It is also required to upgrade the https://github.com/fission-suite/publish-action to the latest CLI version.

This is related to #496 but an interim step, mostly to clean up the publish action in a backwards compatible way (right now the publish action takes advantage of old CLI versions that would check if there was a machine key in place already). 

Example: 
Screen Shot 2021-05-18 at 9 38 40 AM

And if the key file is bad or wrong (e.g. my key from staging used against production):
Screen Shot 2021-05-18 at 9 39 05 AM

### DIFF
--- a/fission-cli/library/Fission/CLI/Handler/Setup.hs
+++ b/fission-cli/library/Fission/CLI/Handler/Setup.hs
@@ -1,47 +1,53 @@
 module Fission.CLI.Handler.Setup (setup) where
 
 import           Network.IPFS
-import qualified Network.IPFS.Process.Error        as IPFS.Process
+import qualified Network.IPFS.Process.Error                  as IPFS.Process
+import           RIO.FilePath
 
-import           Network.DNS                       as DNS
+import           Network.DNS                                 as DNS
 import           Servant.Client
 
 import           Fission.Prelude
 
 import           Fission.Error
-import           Fission.Key.Error                 as Key
+import           Fission.Key                                 as Key
 
 import           Fission.User.DID.Types
 import           Fission.User.Email.Types
-import qualified Fission.User.Username.Error       as Username
+import qualified Fission.User.Username.Error                 as Username
 import           Fission.User.Username.Types
 
-import           Fission.Web.Client                as Client
+import           Fission.Web.Auth.Token.JWT.Types
+import           Fission.Web.Client                          as Client
 import           Fission.Web.Client.HTTP.Class
+import qualified Fission.Web.Client.User                     as User
 
-import qualified Fission.CLI.Environment.OS        as OS
-import           Fission.CLI.Environment.Types
+import           Fission.CLI.Display.Error                   as CLI.Error
+
+import           Fission.CLI.Environment                     as Env
+import qualified Fission.CLI.Environment.OS                  as OS
 import           Fission.CLI.Remote
 
-import qualified Fission.CLI.User                  as User
+import qualified Fission.CLI.User                            as User
 
-import qualified Fission.CLI.Display.Success       as Display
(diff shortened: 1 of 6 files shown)


fission-cli/library/Fission/CLI/Handler/Setup.hs
```diff
@@ -72,19 +79,33 @@ setup maybeOS maybeUsername maybeEmail = do
...
+              False ->
+                User.register maybeUsername maybeEmail
+
+              True -> do
+                logUser @Text "🔗 Please open auth.fission.codes on a signed-in device"
+                signingSK <- Key.Store.fetch $ Proxy @SigningKey
+                rootURL   <- getRemoteBaseUrl
+                Login.consume signingSK rootURL {baseUrlPath = "/user/link"} maybeUsername
+
+          Display.putOk $ "Done! Welcome to Fission, " <> textDisplay username <> " ✨"
```
expede, line 111: Purely aesthetic suggestion:

This is readable, but I'd suggest breaking the `maybeKeyFile` cases into named helper functions. It just feels a bit long.

fission-cli/library/Fission/CLI/Key/Store.hs
```diff
@@ -67,6 +68,26 @@ persist keyRole key = do
...
+  -> m ()
+fromFile keyRole keyFile = do
+  scrubbed <- doesFileExist keyFile >>= \case
+    False ->
+      raise Key.DoesNotExist
+    True -> do
+      bs <- readFileBinary keyFile
+      return $ B64.Scrubbed.scrub bs
+  secretKey <- ensureM $ parse keyRole scrubbed
+  persist keyRole secretKey
```
expede, line 89: No need to change, but another option that flows in one direction (`<-`, `->` and `>>=` together is fine, but sometimes less load to just read one lines and arrows one way. Plus it subordinates detail)

```haskell
fromFile keyRole keyFile = do
  scrubbed  <- getScrubbed keyFile
  secretKey <- ensureM $ parse keyRole scrubbed
  persist keyRole secretKey
  where
    getScrubbed keyFile =
      doesFileExist keyFile >>= \case
        False ->
          raise Key.DoesNotExist

        True -> do
          bs <- readFileBinary keyFile
          return $ B64.Scrubbed.scrub bs
```

fission-cli/library/Fission/CLI/Parser/Command/Setup.hs
```diff
@@ -62,6 +72,14 @@ email = do
     ""   -> Nothing
     mail -> Just mail
 
+keyfile :: ReadM (Maybe FilePath)
+keyfile = do
+  raw <- str
+  pure case raw of
+    ""   -> Nothing
+    path -> Just path
```
expede, line 80: Nothing to change in this PR, but I wonder if we should start collecting these in their own module to share between parsers 🤔

expede (approved):
This looks 🔥🔥🔥 🎉  Just a couple of suggestions, but they're very much optional.

Thanks @walkah LGTM 👍

expede:
Oh, also please bump the version number

fission-cli/library/Fission/CLI/Key/Store.hs
walkah: done and learned a new thing (`forall`) in the process! 🙏

PR #517, opened 2021-05-20
Update readme with correct fission guide links

therealjeffg:
The install and getting started links needed to be updated.


walkah (approved):
Amazing Season 6 - cnuNz0fTBIUGnx4F9T

Good catch! 🙏

PR #522, opened 2021-06-03
Remove STACK_IN_NIX_SHELL from README

expede:
This was fixed in Stack 2.7. You can just `stack build` now, and it defaults to building in a pure Nix environment. You can still pass it `--no-nix` for non-Nix builds.


walkah (approved):
Happy Despicable Me - YFIn0ICJFwGNa

PR #529, opened 2021-06-17
Account Recovery

matheus23:
* Route for sending a recovery email
* Route for updating DID using a challenge obtained from the former

Draft for now. Expect this to change, but feel free to give feedback.

To-do:
* [x] Add a new Email template to SIB for account recovery and add config options for that in env.yaml
* [x] Clean up some stuff. (E.g. web api DID.Types.hs file. Do we want to 'overload' the `/user/did` route?)

See the swagger docs live at https://matheus23.fission.pizza/docs/ (in case I've got the pizza instance running just now lol).


fission-web-server/library/Fission/Web/Server/RecoveryChallenge.hs
```diff
@@ -0,0 +1,20 @@
...
+
+import           Fission.Challenge.Types
+
+import           Fission.Web.Server.RecoveryChallenge.Creator.Class
+import           Fission.Web.Server.RecoveryChallenge.Retriever.Class
+
+
+recoveryLink :: Text -> Challenge -> Text
+recoveryLink recoveryAppUrl challenge =
+  recoveryAppUrl <> "?challenge=" <> toUrlPiece challenge
```
expede, line 20: Could we generate this with `mkLink`?

matheus23: I've [tried that before](https://github.com/fission-suite/fission/pull/529/commits/3d9351bb8053d7a4110c244c33495fabaae525ed#diff-f1e24b239e27892463f95beea5b05887f5d6be9fe542ba2e4f48ade5f3e47f16R18-R25), but it felt kind of clunky, because it's not part of the actual server API, but an 'apps API'. Also, it separates the baseUrl from the route, and the route itself is hardcoded, so instead of `recoveryAppUrl` being configurable as e.g. `https://dashboard.fission.codes/recover/`, the `recover/` part would have to be hardcoded in a servant route type.
Aaand I didn't know how to generate a link with a trailing slash (which I need unfortunately) using servant's `safeLink`. (Maybe `recover :> "" :> QueryParam ...` would've worked?)

If there's a more structural way to create http urls somehow, I'd love to use them (something like [`elm/url`](https://package.elm-lang.org/packages/elm/url/latest/Url)?)

expede: Oh, I see, this is the link to the dashboard app. Yeah makes sense, unless we start having a bunch of these.

> with a trailing slash (which I need unfortunately) 

Why? Is this something that Elm expects? Or is it an implicit `index.html`?

> If there's a more structural way to create http urls somehow, I'd love to use them (something like elm/url?)

Yeah, there's definitely options for these:

1. Create a `data TrailingSlash`, and give it the required Servant instances
2. Use [`URI`](https://www.stackage.org/haddock/lts-18.0/network-uri-2.6.4.1/Network-URI.html#v:uri)
  * Can then plug into `linkURI :: Link -> URI`, so something like `withTrailing = show . addSlash . linkURI . mkLink`

matheus23: > Why? Is this something that Elm expects? Or is it an implicit index.html?

Yeah, there's an index.html file at https://matheus23-dashboard.fission.pizza/recover/index.html

---

Thanks for the pointers to URI/linkURI!

fission-web-api/library/Fission/Web/API/User/DID/Types.hs
```diff
@@ -1,18 +1,30 @@
...
   :> Description "Set currently authenticated user's root public key to another one"
   --
   :> ReqBody '[JSON] Key.Public
   --
   :> Auth.HigherOrder
   :> PutNoContent
+
+type SetViaChallenge
+  =  Summary "Update Public Key via Email Challenge"
+  :> Description "Set given username's public key to something else via providing a challenge obtained from /user/email/recover"
```
expede, line 24: Optional, but "an idea" — we could use the [`MkLink`](https://www.stackage.org/haddock/lts-18.1/servant-0.18.3/Servant-API.html#t:MkLink) interface to autogenerate the path. It would automatically change if we ever moved that (unlikely), but mainly it'll complain if it breaks.

fission-web-api/library/Fission/Web/API/User/DID/Types.hs
```diff
@@ -1,18 +1,30 @@
...
+  --
+  :> ReqBody '[JSON] Key.Public
+  :> Capture "Username" Username
+  :> QueryParam' '[Required, Strict] "challenge" Challenge
```
expede, line 28: Why as a query param in particular?

fission-web-server/library/Fission/Web/Server/Email/SendInBlue/TemplateOptions/Types.hs
```diff
@@ -1,18 +0,0 @@
...
-data TemplateOptions = TemplateOptions
-  { verifyLink :: BaseUrl
-  , username   :: Username
-  }
-
-instance ToJSON TemplateOptions where
-  toJSON TemplateOptions { verifyLink, username } =
-    Object [ ("VERIFY_LINK", toJSON verifyLink )
-           , ("USERNAME",    toJSON username )
-           ]
```
expede, line 18: I guess we need to change some stuff in SiB?

expede: Whoops, no it's below just renamed, gotcha. Odd that sometimes git can't figure this stuff out

fission-web-server/library/Fission/Web/Server/Handler/User/Email/Recover.hs
```diff
@@ -0,0 +1,40 @@
...
+  Entity userId User { userEmail } <- Web.Err.ensureMaybe couldntFindUser =<< getByUsername username
+  email <- Web.Err.ensureMaybe noAssociatedEmail userEmail
+  now       <- currentTime
+  challenge <- RecoveryChallenge.create userId now
+  Web.Err.ensureM $ sendRecoveryEmail (Recipient email username) challenge
+  return NoContent
+
+  where
+    couldntFindUser =
+      err422 { errBody = "Couldn't find a user with this username" }
```
expede, line 37: We have a thing for this! `throwM (NotFound @User)` will throw the correct error code and message (though you may want to add a `NotFound Username` for this case)

fission-web-server/library/Fission/Web/Server/Models.hs
```diff
@@ -99,6 +99,17 @@ UserChallenge
...
+--------------------------------------------------------------------------------
+
+UserRecoveryChallenge
+  userId     UserId
+  hash       Challenge
+  insertedAt UTCTime
+
+  UniqueRecoveryChallenge hash
+
+  deriving Show Eq
```
expede, line 111: Is this different from a `UserChallenge`?

fission-web-server/library/Fission/Web/Server/RecoveryChallenge/Destroyer/Class.hs
```diff
@@ -0,0 +1,16 @@
...
+import           Fission.Web.Server.Models        as Models
+import           Fission.Web.Server.MonadDB.Types
+
+class Monad m => Destroyer m where
+  destroyForUser :: UserId -> m ()
+
+instance MonadIO m => Destroyer (Transaction m) where
+  destroyForUser userId =
+    delete $ from \challenge ->
+      where_ $ challenge ^. UserRecoveryChallengeUserId ==. val userId
```
expede, line 16: Totally fine here, but if you use the Persistent interface directly:

```haskell
import Database.Persist as P
-- ...
destroyForUser = P.destroy
```

expede (approved):
Looking good! One question and a minor suggestion, but LGTM!

fission-web-api/library/Fission/Web/API/User/DID/Types.hs
matheus23: The `SetViaChallenge` route is only used from within the recovery app, not from somewhere in the server.
The server only sends the e-mail and in the email there's a link to the recovery app itself. The app then uses the `SetViaChallenge` route.

fission-web-api/library/Fission/Web/API/User/DID/Types.hs
matheus23: Uh, good question. No particular reason honestly. And looking at the other verification route, it looks like it'd be more consistent if I'd be using a path here instead. I'll change that :+1:

fission-web-server/library/Fission/Web/Server/Email/SendInBlue/TemplateOptions/Types.hs
matheus23: > I guess we need to change some stuff in SiB?

There is changes for some things in SIB, but not for the existing stuff: We needed to add the new recovery email template, but I've already done that as my pizza instance is hooked up to "production" SIB.

fission-web-server/library/Fission/Web/Server/Models.hs
matheus23: It's structurally not much different, it's only structural difference is an additional `insertedAt` field for tracking when the challenge expires.

But it's generally meant to be used for another purpose, and I think should therefore be its own table: The recovery process should run independently of the email verification process for accounts.
The biggest semantic difference is that recovery challenges are deemed expired after one hour. This is not the case for `UserChallenge`s.

matheus23:
I've changed the route to use capture instead of a QueryParam now. Let me know if this is good to go now :)

Also, there's a small checklist for getting this running on staging, the `env.yaml` config needs to be updated according to the `env.yaml.example` in this PR. Concretely, we need to set these fields:
```yaml
send_in_blue:
  api_key: XXX # No need to change
  base_url: https://api.sendinblue.com # No need to change as well
  verification_email_template_id: 17 # This setting's key changed
  recovery_email_template_id: 26 # This key is new
  recovery_app_url: https://fission-dashboard.fissionapp.net/recover/ # This will be the recovery app URL for **staging**
```

fission-web-server/library/Fission/Web/Server/Models.hs
expede: Ah, expiry, good point 👍

expede:
> Let me know if this is good to go now :)

Bump the minor server version and merge away!

PR #530, opened 2021-06-21
Hotfix: CLI watch flag causes hang

expede:
Turns out that `forever . lifIO` is not the same as `liftIO . forever` 😅  Which makes sense. _Which thing_ are you running forever? `liftIO :: IO a -> FissionCLI a` (i.e. embed IO action in a pure data structure), so we want the conversion on the _outside_.


walkah (approved):
Well, that looks better :)  🚢

PR #536, opened 2021-07-02
Varnish cache BAN on update

expede:
* [x] Purge on app modify
* [x] Purge on app destroy
* [x] Purge on user modify
* [x] Purge on user destroy
* [x] Confirm working on staging

# Server Logs

Screen Shot 2021-10-01 at 22 48 51
Screen Shot 2021-10-01 at 22 49 33

I'd show more, but it's kind of hard to capture in screenshots ("Here it an app before, and here it is after a push!"). 

# Setting Expectations

Because of the way Varnish bans work, it can often take about a minute to actually cache bust. The end result is a refreshed page, you just sometimes have to be patient.

If the cache purge fails, we log and continue. It doesn't seem worth reporting a 502 over, since our worst case is that a page will at worst appear to update slowly and we can always manually cache bust. We'll see these in Sentry if they start to pile up.


fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -673,15 +692,49 @@ instance App.Modifier Server where
...
+                            let
+                              urls :: [URL]
+                              urls = appDomains <&> \(Entity _ AppDomain {..}) ->
+                                        URL { domainName = appDomainDomainName
+                                            , subdomain  = appDomainSubdomain
+                                            }
+
+                            NGINX.purgeMany urls >>= \case
+                              Left err -> return $ openLeft err
+                              Right _  -> return $ Right appId
```
expede, line 716: This entire section will feel _much_ cleaner when we switch to Rescue

fission-web-server/library/Fission/Web/Server/Internal/NGINX/Purge/Types.hs
```diff
@@ -0,0 +1,11 @@
...
+
+module Fission.Web.Server.Internal.NGINX.Purge.Types (PURGE (..)) where
+
+import           Network.HTTP.Req
+
+data PURGE = PURGE
+
+instance HttpMethod PURGE where
+  type AllowsBody _ = 'NoBody
+  httpMethodName _pxy = "PURGE"
```
expede, line 11: Easy enough

expede:
Now that https://github.com/fission-suite/infrastructure/pull/73 is deployed, we can test that this works!

expede:
Rebased on top of #547

expede:
You can really tell when I'm debugging stupid typos on staging by number of close commits 😛

expede:
Talked it through with @walkah 

The performance hit on write makes us nervous. We're going to wait and see what comes out of improving our IPFS setup directly first. We still have this code around if we want to add our own caches for larger users with immutable data (which we're currently throwing paid AWS caches in front of)

expede:
Backported the important parts

PR #539, opened 2021-07-06
Fetch go-ipfs from GH instead of ipfs.io

### LINKED ISSUE #537: go-ipfs intermittently fails to download from ipfs.io
# Summary

## Problem

go-ipfs fails to download from ipfs.io intermittently

### Impact

Cannot setup the CLI and/or have to retry

## Solution

Download go-ipfs from GH releases

expede:
Closes #537 

# Changelog

* Grab `go-ipfs` from GitHub instead of `ipfs.io`
  * Remove "bootstrap" environment, since no longer fetched from an IPFS gateway
  * Have to unzip & untar it, since that's how they provide it
  * Comparable download performance to ipfs.io (surprising!), but it... you know... will always work
* Better interface for grabbing a specific `go-ipfs` build 
  * Extensible to more architectures (e.g. native M1), but leaving for now
* Switch from IPFS 0.7 -> 0.9
  * Still takes "a while" for the daemon to start 😭 
  * Push seems roughly the same speed 👍 
* Drop special-cased Nix environment (Linux works these days)
* Bump Stackage to 18.1

# Visual Evidence

Screen Shot 2021-07-06 at 17 19 49


fission-cli/library/Fission/CLI/GitHub/Class.hs
```diff
@@ -0,0 +1,7 @@
+module Fission.CLI.GitHub.Class (MonadGitHub (..)) where
+
+import           Fission.Prelude
+import           Servant.Client
+
+class Monad m => MonadGitHub m where
+  sendRequest :: ClientM a -> m (Either ClientError a)
```
expede, line 7: May be reusable; we'll see!

fission-cli/library/Fission/CLI/IPFS/Download.hs
```diff
@@ -0,0 +1,16 @@
...
+
+import           Servant.Client
+
+import           Fission.Prelude
+
+import           Fission.CLI.IPFS.Download.GitHub.Types
+import qualified Fission.CLI.IPFS.Release.Types         as IPFS
+
+getRelease :: IPFS.Release -> ClientM Lazy.ByteString
+getRelease release@IPFS.Release {..} = client (Proxy @GetRelease) version release
```
expede, line 16: Just convenience so we don't mismatch the version the parts of the paths segments (the version is used in two places because of their naming convention)

fission-cli/library/Fission/CLI/IPFS/Download/GitHub/Types.hs
```diff
@@ -0,0 +1,17 @@
...
+import qualified Fission.CLI.IPFS.Version.Types as IPFS
+
+type GetRelease
+  = "ipfs"
+  :> "go-ipfs"
+  :> "releases"
+  :> "download"
+  :> Capture "version" IPFS.Version
+  :> Capture "release" IPFS.Release
+  :> Get '[OctetStream] Lazy.ByteString
```
expede, line 17: `Lazy.ByteString` because the tar extractor expects that, so it's more convenient. In general we default to strict because it's easier to reason about.

fission-cli/library/Fission/CLI/IPFS/Executable.hs
```diff
@@ -59,17 +66,57 @@ place' ::
...
+  , MonadLogger      m
+  , MonadRescue      m
+  , m `Raises` IPFS.Error
+  )
+  => m ()
+configure = do
   logUser @Text "🎛️  Configuring managed IPFS"
-
   void IPFS.Config.init
   void IPFS.Config.enableRelay
```
expede, line 122: I broke the stages out here, mainly because there's more manual work now with the tarball

fission-cli/library/Fission/CLI/IPFS/Version/Types.hs
```diff
@@ -0,0 +1,26 @@
...
+instance Display Version where
+  display Version {..} =
+    mconcat
+      [ "v"
+      , display @Int $ fromIntegral major
+      , "."
+      , display @Int $ fromIntegral minor
+      , "."
+      , display @Int $ fromIntegral patch
+      ]
```
expede, line 23: We want this printed in a very specific format, so this is a lightweight way to enforce this to be correct. Also may be reusable.

fission-cli/library/Fission/CLI/Types.hs
```diff
@@ -93,16 +92,13 @@ import           Fission.Web.Auth.Token.Types
 
 import           Fission.Web.Client
 
-import           Fission.Internal.Orphanage.BaseUrl                ()
 import           Fission.Internal.Orphanage.CID                    ()
-import           Fission.Internal.Orphanage.DNS.DNSError           ()
-import           Fission.Internal.Orphanage.OpenUnion              ()
-
-import           Fission.Internal.Orphanage.ClientError            ()
```
expede, line 101: Warnings from the compiler about unused imports. Thanks GHC!

fission-cli/library/Fission/CLI/Types.hs
```diff
@@ -812,23 +824,6 @@ instance
...
-  , SomeException `IsMember` errs
-  , Exception (OpenUnion errs)
-  , Display   (OpenUnion errs)
-  , Contains errs errs
-  )
-  => IPFS.MonadRemoteIPFS (FissionCLI errs cfg) where
-  runRemote query = do
-    logDebug @Text "Running remote IPFS"
-    _ <- IPFS.Daemon.runDaemon
-    runBootstrapT $ runRemote query
```
expede, line 830: We no longer use the HTTP gateway for anything, so ✂️

walkah (approved):
Looks good! I think we've got an opportunity to refactor / consolidate some of our github api interactions, but otherwise:

this is awesome whitney port - xT5LMFM8Pek1GXPHpK

expede:
@walkah that's a good point. I'll extract that GH stuff out 👍

PR #541, opened 2021-07-08
Automigrate IPFS repo

expede:
IPFS 0.9 has a new `--migrate` flag on the daemon, which automigrates the repo on startup. This PR enables that flag.

I tested this manually by removing my existing repo, and setting up a new repo via my global IPFS 0.7 at the correct location (via `IPFS_PATH`), and then running `fission up`, which worked

### DIFF
--- a/fission-cli/library/Fission/CLI/Types.hs
+++ b/fission-cli/library/Fission/CLI/Types.hs
@@ -606,11 +606,11 @@ instance
 
         Turtle.export "IPFS_PATH" $ Text.pack ipfsRepo
 
-        process <- startProcess . fromString $ intercalate " "
+        process <- startProcess . fromString $ unwords
           [ "IPFS_PATH=" <> ipfsRepo
           , "2> /dev/null"
           , ipfsPath
-          , "daemon > /dev/null"
+          , "daemon --migrate > /dev/null"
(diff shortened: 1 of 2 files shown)


fission-cli/library/Fission/CLI/Types.hs
```diff
@@ -606,11 +606,11 @@ instance
 
         Turtle.export "IPFS_PATH" $ Text.pack ipfsRepo
 
-        process <- startProcess . fromString $ intercalate " "
+        process <- startProcess . fromString $ unwords
```
expede, line 609: `unwords = intercalate " "`

This was a suggestion from HLint

walkah: "unwords". naming is hard.

walkah (approved):
🎉  Ran a quick test ... works for me ™️ 

ship it - Hw8vYF4DNRCKY

fission-cli/library/Fission/CLI/Types.hs
expede: @walkah haha yeah. Some context here: `words` is a function that breaks a string by spaces into a list of words. This is the reverse operation.

PR #542, opened 2021-07-14
Correct version change

matheus23:
I accidentally bumped the wrong version in #529 . I should've incremented the minor version. PRing this to fix this!


expede:
Thanks for catching that!

PR #544, opened 2021-08-19
Fix Swagger required fields for App Index payload

expede:
This is on me for not catching it in code review originally


walkah (approved):
Much Better - ieh305yo7DKqC5Mrla

PR #545, opened 2021-08-20
WIP spiral ratchet

expede:
Quick and dirty spiral ratchet. This was mainly so I could play with it in a REPL to debug something for Brendan.


spiral-ratchet/LICENSE
```diff
@@ -0,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
expede, line 1: This package is MIT, plz fix

PR #546, opened 2021-08-27
Versioned REST API

expede:
# Currently Deployed

Today our API is unversioned:

```sh
»  curl https://runfission.net/ping
"pong"
```

This has lead to breaking changes downstream in JS when the return types changed to support a feature in the CLI

# This PR

* Add `/v2/api`
* Leave the unprefixed paths as-is (but deprecated) as a (long) grace period

```sh
»  curl https://runfission.net/v2/api/ping
"pong"
```

Why the version first? Because we're also going to having things like versioned docs (`/v2/docs`). I have no strong feelings on this; just seemed to make sense.

# While I Had the Hood Up 👩‍🔧 

We've had folks try to use 3rd party tools to check if our server is up. We have nothing at the root, so these didn't return a success code and show as failed. For example:

```
»  curl -i https://runfission.com/
HTTP/2 404
date: Sun, 29 Aug 2021 05:05:10 GMT
server: Warp/3.3.17
```

https://downforeveryoneorjustme.com/runfission.com

Screen Shot 2021-08-27 at 16 47 31

...except it's not down; it's just not returning a 200-level code!

So I added an empty path at the root that returns a 200 and no body:

```sh
»  curl -i https://runfission.net/
HTTP/2 200
date: Fri, 27 Aug 2021 23:40:40 GMT
content-type: application/json;charset=utf-8
server: Warp/3.3.17
```

Screen Shot 2021-08-27 at 16 48 29

# Docs

* Latest at root
* Versioned at `/v2/docs`
* No docs for the root pathed API, since it's deprecated

## Latest
Screen Shot 2021-08-28 at 16 35 40

## V2
(Currently the latest)
Screen Shot 2021-08-28 at 16 35 34


fission-web-server/library/Fission/Web/Server.hs
```diff
@@ -179,10 +181,20 @@ bizServer ::
...
+      :<|> DNS.handler
+      :<|> Auth.UCAN.Verify.handler
+
+    v_ =   IPFS.handler
+      :<|> App.handler
+      :<|> Heroku.handler
+      :<|> User.handler
+      :<|> Ping.handler
+      :<|> DNS.handler
+      :<|> Auth.UCAN.Verify.handler
```
expede, line 200: Duplicated because we expect the V2 version to change over time, but the unversioned one not to!

fission-web-api/library/Fission/Web/API/Auth/Types.hs
```diff
@@ -1,19 +1,25 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
expede, line 1: Where "orphan instance" here is because we're defining type aliases. Yes, you could _potentially_ have a conflicting instance elsewhere, but you shouldn't use it! Such a compiler error would be correct. We could create a custom type for these, but then that won't plug into the auth subsystem. I'm sure it's possible with enough effort, but this feels fine given the context.

fission-web-server/library/Fission/Web/Server.hs
```diff
@@ -59,8 +58,21 @@ import           Fission.Web.Server.Relay.Store.Class
...
 -- | Top level web API type. Handled by 'server'.
-type API    = Swagger.API :<|> Fission.API :<|> LinkWS
+type API
+  -- V2
+  =    Fission.V2
+  -- Docs
+  :<|> Swagger.Latest
+  :<|> Swagger.V2
+  -- API with omitted docs
+  :<|> Fission.V_
```
matheus23, line 71: Why isn't this using `Fission.API` anymore, which includes `Fission.V2` and `Fission.V_`? I assume the ordering makes a difference?

matheus23 (approved):

fission-web-server/library/Fission/Web/Server.hs
expede: The ordering does make a difference, yeah. We want the docs before the unversioned section, because it matches incoming routing in this sequence. For example, if you put the `/` matcher first, nothing else will ever get matched. (This isn't just a servant thing; it's pretty standard)

We have the `servant-swagger` in `fission-web-server` because it imports `servant-server`, which has extra runtime dependencies. EDIT: actually I'm not sure that's true. We don't want to pull in things like Postgres for the CLI, but that's not a dependency for servant-server. I'll take a look at that!

On this topic, I actually have a WIP branch from this weekend that uses [`Servant.API.Generic`](https://www.stackage.org/haddock/lts-18.8/servant-0.18.3/Servant-API-Generic.html), which we had been putting off due. to performance. It's in the main package now, and the compile-time overhead doesn't seem too bad (subjectively, I'm going to run the actual numbers). It cleans up a lot of the router code, but adds a couple potential gotchas (e.g. it fails if you don't use `deriving Generic`). I'll put that up before I merge this.

expede: WIP but you can see what this (roughly) looks like here https://github.com/fission-suite/fission/pull/547

PR #550, opened 2021-09-10
Updated RSA DID multiformat

expede:
`rsa-x509-pub` has [made it into Multicodec](https://github.com/multiformats/multicodec/pull/226). This PR updates the 

* [x] Update RSA format
* [ ] Test backwards compatibility
* [ ] ~~Exchange test cases with Qri team~~ Test against W3C vectors


fission-cli/library/Fission/CLI/IPFS/Peers.hs
```diff
@@ -29,4 +28,4 @@ getPeers ::
 getPeers = do
   logDebug @Text "🌌📥 Getting peers remotely"
   Cursor.withHidden $ CLI.Wait.waitFor "Retrieving Fission Peer List..." do
-    ensureM $ sendRequest IPFS.getPeers
+    ensureM $ sendRequest getIPFSPeers
```
matheus23, line 31: That's interesting! Why did you decide to switch the naming convention? I personally like the `Something.x` convention, so I'm curious :)

fission-core/library/Fission/User/DID/Types.hs
```diff
@@ -113,13 +116,11 @@ instance Display DID where -- NOTE `pk` here is base2, not base58
...
-                               {-   ^     ^     ^
-                                    |     |     |
-                                    |    "expect 373 Bytes", encoded in the mixed-endian format
-                                  "raw"
-                              -}
+          Ed25519PublicKey ed ->
+            0xed : 0x01 : BS.unpack (BA.convert ed)
+
+          RSAPublicKey rsa ->
+            0x12 : 0x05 : BS.unpack (Lazy.toStrict . Builder.toLazyByteString . getUtf8Builder $ display rsa)
```
matheus23, line 123: :raised_hands:

matheus23 (approved):

fission-cli/library/Fission/CLI/IPFS/Peers.hs
expede: I also prefer that convention. This is from the new Servant.Generic client, since it's easier to decompose them in one file rather than importing the outer level into separate files and exposing them there. The tradeoff is that it puts them in a flat namespace.

fission-core/library/Fission/Internal/Orphanage/RSA2048/Public.hs
```diff
@@ -32,9 +34,8 @@ instance Arbitrary RSA.PublicKey where
 
 instance Display RSA.PublicKey where
   textDisplay pk =
-    X509.PubKeyRSA pk
-      |> X509.pubKeyToPEM
-      |> PEM.pemWriteBS
+    [X509.PubKeyRSA pk]
+      |> X509.writePubKeyFileToMemory
```
expede, line 38: These end up _larger_ than before.

> `writePubKeyFile path = writePEMs path . map pubKeyToPEM`

Soooo, that's in PEM, not X.509. This is definitely wrong.

expede:
Rebased. Tests failing, and need to validate the format against what Qri has (and then contribute a test vector in the Multicodec repo)

expede:
Waiting for the dust to settle over here: https://github.com/multiformats/multicodec/issues/235

fission-core/library/Fission/User/DID/Types.hs
matheus23: Just randomly took a stroll through here, and now I know this is actually incorrect, the prefix needs to be `0x85 : 0x24 : Bs.unpack ...`
I've implemented it that way in ts-ucan and it tests fine with the test vectors from the DID spec.
https://github.com/ucan-wg/ts-ucan/pull/42/files#diff-109e954c94a80ae8fd78b2ce9734642de7a787f48926c9a5ba18ad1ec76225ffR17

The reason it's `0x85 0x24` is that that's the number `0x1205` encoded as unsigned varint.
Just like a couple of lines above, the multicodec for ed25519-pub `0xed` is encoded as unsigned varint into `0xed 0x01`.

matheus23:
Closing in favor of #573

PR #551, opened 2021-09-13
Fix whoami route typo

expede:
* [x] Fix broken `/user/whoami` route
* [x] Bump stack to 18.10 LTS
* [x] Download Swagger API JSON for old and new routers for future use
* [x] Manually compare docs
* [x] Manually test

# CLI

```
»  fission setup -R staging
🌱 Setting up environment
🪐 Downloading managed IPFS for macOS
🕐🎛️  Configuring managed IPFS
🔑 Setting up keys
🏠 Do you have an existing account? [Y/n] n
Username: expede-210914-staging
Email: be.zelenka@gmail.com
✅ Registration successful! Head over to your email to confirm your account.
🎛️  Initializing user config file
✅ Done! Welcome to Fission, expede-210914-staging ✨

»  fission whoami -R staging
💻 Currently logged in as: expede-210914-staging

»  fission app register -R staging
👷 Choose build directory (.):
✅ App initialized as little-wide-silk-elf.fissionapp.net
⏯️  Next run fission app publish [--open|--watch] to sync data
💁 It may take DNS time to propagate this initial setup globally. In this case, you can always view your app at https://ipfs.runfission.com/ipns/little-wide-silk-elf.fissionapp.net

»  fission up -R staging
🕗🛫 App publish local preflight
✈️  Pushing to remote
🚀 Now live on the network
📝 DNS updated! Check out your site at:
🔗 little-wide-silk-elf.fissionapp.net
```

# Browser (Signup & Drive)

Screen Shot 2021-09-14 at 20 47 07
Screen Shot 2021-09-14 at 20 47 26
Screen Shot 2021-09-14 at 20 48 01
Screen Shot 2021-09-14 at 20 48 07

# Linking

Screen Shot 2021-09-14 at 20 49 19


matheus23 (approved):
Can we somehow make sure that there's no other typos like this?
Maybe by comparing the swagger files?

I just searched and found [openapi-diff](https://github.com/OpenAPITools/openapi-diff). I'll give that a try!

matheus23:
The openapi-diff tool I linked above didn't really work, so I tried this [openapi-diff](https://www.npmjs.com/package/openapi-diff) tool from atlassian on npm instead.
Also small sidenote, I had to convert our OpenAPI v2 into OpenAPI v3, because the diff tools would only understand that. There's an endpoint at converter.swagger.io, so I used that.

So I tried this:
```sh
$ yarn global add openapi-diff
$ curl https://converter.swagger.io/api/convert?url=https://runfission.net/docs.json > docsv2.json
$ curl https://converter.swagger.io/api/convert?url=https://runfission.com/docs.json > docsv1.json
$ openapi-diff docsv1.json docsv2.json
No changes found between the two specifications
```

I assume that is because staging was rolled back to before the v2 route changes.

So I tried to build this PR locally and run the server to get the OpenAPI file, but I can't build this PR (and main) with the same error as the CI on ubuntu-latest:
```
error: a 'x86_64-darwin' with features {} is required to build '/nix/store/asi46dj02lxc5ccq8azjr2wkfw7rrmxv-build.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, nixos-test}
```

I'll leave it at that for now. Let me know if there's something I can do to help make this run on linux!

expede:
> ```
> error: a 'x86_64-darwin' with features {} is required to build '/nix/store/asi46dj02lxc5ccq8azjr2wkfw7rrmxv-build.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, nixos-test}
> ```

Yeah, I committed pointing the nix.shell at Rosetta in stack.yaml. Will fix.

https://github.com/fission-suite/fission/blob/main/stack.yaml#L38

> I just searched and found openapi-diff. I'll give that a try!

I think this is great, but also we don't do this often enough to automate the process. It's just trading one declarative spec for another. A replay test suite would be useful long-term because it also captures the responses from a real-world server request, but we should probably write more tests _in general_ 😅

expede:
😫 Oh god, I dropped the sections from the docs to make them easier to generate, and that breaks `openapi-diff`. Literally cases where the endpoint is there, but the swagger docs are confused because it's no longer nested.

This will be more useful from here on out, but I'm stuck doing it by hand for this PR 🤷‍♀️ 

I _will_ add `openapi-diff` to (soon-to-exist) debugging environment, though!

shell.nix
```diff
@@ -1,78 +1,79 @@
...
-  pkgs  = if darwin.stdenv.isDarwin then darwin else nixos;
-  tasks = commands {
-    inherit pkgs;
-    inherit unstable;
-    inherit server-path;
-    inherit server-port;
-  };
+    nixos    = import sources.nixos    overrides;
+    darwin   = import sources.darwin   overrides;
+    unstable = import sources.unstable overrides;
```
expede, line 10: This is a MUCH nicer solution than a copy/pasted rosetta.nix file

expede: Is does mean that Brian and I need to `nix-shell --arg rosetta true` and then run stack with `--no-nix`, but that's the same as what Stack does underneath, so 🤷‍♀️

fission-web-api/library/Fission/Web/API/Types.hs
```diff
@@ -39,20 +39,20 @@ data RoutesV2 mode = RoutesV2
...
-  { ipfs   :: mode :- "ipfs"                  :> ToServantApi IPFS.RoutesV_
-  , app    :: mode :- "app"                   :> ToServantApi App.RoutesV_
-  , user   :: mode :- "user"                  :> ToServantApi User.RoutesV_
-  , dns    :: mode :- "dns"                   :> ToServantApi DNS.Routes
-  , auth   :: mode :- "auth"   :> "ucan"      :> ToServantApi UCAN.Routes
+  { ipfs   :: mode :- "ipfs"           :> ToServantApi IPFS.RoutesV_
+  , app    :: mode :- "app"            :> ToServantApi App.RoutesV_
+  , user   :: mode :- "user"           :> ToServantApi User.RoutesV_
+  , dns    :: mode :- "dns"            :> ToServantApi DNS.Routes
+  , auth   :: mode :- "auth" :> "ucan" :> ToServantApi UCAN.Routes
```
expede, line 56: Whitespace!

expede:
Here's what I ended up doing:

1. Directly on staging server, set docs to generate `RoutesV_` docs, build & run
2. Open production docs in one tab
3. Open staging docs in another tab
4. Walk down the prod (unversioned) docs, and compare to the staging docs. Make sure that:
  * All routes are available
  * Return types are the same

Honestly, tabbing back and forth was great. Aside from the sidebar, the pages are identical, so it's pretty easy to spot where (e.g.) the updated doc generator version has additional information.

I'm going to run through Drive and the CLI and stuff, but I think this is good to go now

matheus23 (approved):

PR #568, opened 2022-01-14
Add ipfs package in hie.yaml

matheus23:
This resolves some errors when going into the ipfs subdirectory's hs files with haskell-language-server for me. :v:


expede (approved):

matheus23:
Closing because we're removing the `hie.yaml` file in #571

PR #570, opened 2022-01-18
Add new errors for a few failure cases

### LINKED ISSUE #538: `fission app register` fails silently if no config.yml exists in ~/.config/fission/
STR:

1. go to ~/.config/fission and move config.yml somewhere
2. go to the root directly of your fancy new project and run `fission app register`

It should instead produce some sort of error eg 'fission is installed but you are not logged in'

Bonus points saying 'please run `fission login` first'.

This is sort of an edge case but people would definitely run into it if they skipped auth during `fission setup` and forgot to link their account before running `app register`. It also affects me juggling different accounts quickly without having to run `fission setup` every time because the ipfs binary download is slooooooow.

### LINKED ISSUE #569: Output a clearer error message when publishing to an unregistered app
# Summary

When attempting to publish to an app that hasn't been registered, the error message should reflect this fact.

## Problem

While setting up a new project, I created `fission.yaml` by hand, and manually entered my desired url. When I published the app, I received the following output:

```bash
$ fission app publish
🕛🛫 App publish local preflight
🕑✈️  Pushing to remotePeer List...
🚫 Server unable to sync data
```

I thought that perhaps the app wasn't being registered automatically, so I tried to register it, and the output suggested that the registration did in fact exist.

```bash
$ fission app register --name some-name
App already set up at some-name.fission.app
```

Publishing again, in verbose mode, made it clear that the app was unregistered:

```bash
FailureResponse (Request {requestPath = (BaseUrl {baseUrlScheme = Https, baseUrlHost = "runfission.com", baseUrlPort = 443, baseUrlPath = ""},"..."), requestQueryString = fromList [("copy-data",Just "true")], requestBody = Nothing, requestAccept = fromList [application/json;charset=utf-8,application/json], requestHeaders = fromList [("Authorization","<REDACTED>")]), requestHttpVersion = HTTP/1.1, requestMethod = "PUT"} (Response {responseStatusCode = Status {statusCode = 404, statusMessage = "Not Found"}, responseHeaders = fromList [("Date","Tue, 18 Jan 2022 01:28:56 GMT"),("Transfer-Encoding","chunked"),("Connection","keep-alive"),("Server","Warp/3.3.17")], responseHttpVersion = HTTP/1.1, responseBody = "App not found"})
```

### Impact

The current error message suggested an issue with my IPFS configuration, so I deleted everything and tried again from scratch. My next guess was an availability issue with the remote peers, which might drive a new user to table Fission and assume it isn't ready for use.

Right now, there aren't many reasons to create a `fission.yaml` outside of the regular `fission app register` flow, but I can imagine this failure case becoming more common as the configuration options available grow in number and complexity. For example, someone may wish to copy their existing `fission.yaml` from another project, in order to copy environment config as described in #567. Similarly, you may begin to see templating tools that scaffold out Webnative applications, and this is a caveat they'd need to be aware of.

## Solution

Since the response from the server is already a 404, this just needs to be translated to a human readable error. Perhaps:

>🚫  The app has not yet been registered. Delete your fission.yaml and run: fission app register --name some-name

Requiring that the user delete their `fission.yaml` isn't a great experience, so it may also be worth supporting a `--force` flag for `fission app register`, that registers an app with the name part of the URL in `fission.yaml`. Then the error message could be adjusted to:

>🚫  The app has not yet been registered. Register it by running: fission app register --force

QuinnWilton:
## Summary
This PR resolves a few minor UX issues I ran into while using the tool today. In particular, it:
- Improves the error when running `app publish` with an unregistered app (closes #569)
- Adds a new error when running `app register` without a `config.yaml` (closes #538)
- Adds a new error when running `whoami` without a `config.yaml`
- Changes one pre-existing error message to have a consistent style with other similar error messages

Feel free to nitpick this as much as you'd like; I haven't worked with the repo before, and I'm out of practice with Haskell, so it's likely I'm missing some idioms!

In particular, I suspect some of these error strings should be pulled into `Fission.CLI.Error.Types`. Then I could probably define a helper somewhere that translates the `NotFound FilePath` into a `NotConfigured` error that more nicely encapsulates this concern.

I avoided doing that to keep these changes small and avoid imposing too large of a review burden onto anyone, but I'm happy to make any adjustments!

## Test plan (required)

`app publish` with an unregistered app:
```bash
[nix-shell:~/dev/fission]$ cat fission.yaml
ignore: []
url: does-not-exist.fission.app
build: ./assets

[nix-shell:~/dev/fission]$ stack run -- app publish
🕛🛫 App publish local preflight
🕑✈️  Pushing to remotePeer List...
🚫 The app has not yet been registered. Please delete your fission.yaml and run `fission app register --name some-name`

[nix-shell:~/dev/fission]$ rm fission.yaml

[nix-shell:~/dev/fission]$ stack run -- app register
👷 Choose build directory (.): assets
✅ App initialized as geriatric-yellow-centaur.fission.app
⏯️  Next run fission app publish [--open|--watch] to sync data
💁 It may take DNS time to propagate this initial setup globally. In this case, you can always view your app at https://ipfs.runfission.com/ipns/geriatric-yellow-centaur.fission.app

[nix-shell:~/dev/fission]$ stack run -- app publish
🕛🛫 App publish local preflight
🕒✈️  Pushing to remotePeer List...
🚀 Now live on the network
📝 DNS updated! Check out your site at: 
🔗 geriatric-yellow-centaur.fission.app
```

`app register` and `whoami` without being logged in:
```bash
[nix-shell:~/dev/fission]$ cat ~/.config/fission/config.yaml
cat: /Users/quinn/.config/fission/config.yaml: No such file or directory

[nix-shell:~/dev/fission]$ stack run -- app register
🚫 Fission is installed, but you are not logged in. Please run `fission login`

[nix-shell:~/dev/fission]$ stack run -- whoami      
🚫 Fission is installed, but you are not logged in. Please run `fission login`

[nix-shell:~/dev/fission]$ mv ~/.config/fission/config.yaml.old ~/.config/fission/config.yaml

[nix-shell:~/dev/fission]$ stack run -- app register
👷 Choose build directory (.): ^C

[nix-shell:~/dev/fission]$ stack run -- whoami
💻 Currently logged in as: quinn
```

### DIFF
--- a/fission-cli/library/Fission/CLI/App.hs
+++ b/fission-cli/library/Fission/CLI/App.hs
@@ -14,6 +14,7 @@ import           Fission.CLI.Connected                   as Connected
 
 import           Fission.CLI.App.Environment             as App.Env
 import qualified Fission.CLI.Display.Error               as CLI.Error
+import           Fission.CLI.Environment                 as CLI.Env hiding (Env)
 import qualified Fission.CLI.Handler                     as Handler
 import           Fission.CLI.Handler.Error.Types         (Errs)
 
@@ -44,21 +45,32 @@ interpret baseCfg cmd = do
         run' :: FissionCLI errs Connected.Config () -> FissionCLI errs Base.Config ()
         run' = ensureM . Connected.run baseCfg timeoutSeconds
 
-      attempt App.Env.read >>= \case
-        Right Env {appURL} -> do
-          UTF8.putTextLn $ "App already set up at " <> textDisplay appURL
-          logDebug . textDisplay $ AlreadyExists @URL
-          raise $ AlreadyExists @URL
-
-        Left errs -> do
+      attempt CLI.Env.get >>= \case
+        Left errs ->
           case openUnionMatch @(NotFound FilePath) errs of
-            Just _ -> do
-              logDebug @Text "Setting up new app"
-              run' $ Handler.appInit appDir buildDir mayAppName
-
-            Nothing -> do
-              logError @Text "Problem setting up new app"
+            Just err -> do
+              CLI.Error.put err "Fission is installed, but you are not logged in. Please run `fission login`"
               raise errs
+
+            Nothing ->
+              raise errs
+
+        Right _ ->
+          attempt App.Env.read >>= \case
+            Right Env {appURL} -> do
+              UTF8.putTextLn $ "App already set up at " <> textDisplay appURL
+              logDebug . textDisplay $ AlreadyExists @URL
+              raise $ AlreadyExists @URL
+
+            Left errs -> do
+              case openUnionMatch @(NotFound FilePath) errs of
+                Just _ -> do
+                  logDebug @Text "Setting up new app"
+                  run' $ Handler.appInit appDir buildDir mayAppName
+
+                Nothing -> do
+                  logError @Text "Problem setting up new app"
+                  raise errs
 
     Up App.Up.Options {open, watch, updateDNS, updateData, filePath, ipfsCfg = IPFS.Config {..}} -> do
       let
--- a/fission-cli/library/Fission/CLI/App.hs
+++ b/fission-cli/library/Fission/CLI/App.hs
@@ -15,6 +15,7 @@ import           Fission.CLI.Connected                   as Connected
 import           Fission.CLI.App.Environment             as App.Env
 import qualified Fission.CLI.Display.Error               as CLI.Error
 import           Fission.CLI.Environment                 as CLI.Env hiding (Env)
+import           Fission.CLI.Error.Types
 import qualified Fission.CLI.Handler                     as Handler
 import           Fission.CLI.Handler.Error.Types         (Errs)
 
@@ -45,9 +46,9 @@ interpret baseCfg cmd = do
         run' :: FissionCLI errs Connected.Config () -> FissionCLI errs Base.Config ()
         run' = ensureM . Connected.run baseCfg timeoutSeconds
 
-      attempt CLI.Env.get >>= \case
+      attempt CLI.Env.alreadySetup >>= \case
         Left errs ->
-          case openUnionMatch @(NotFound FilePath) errs of
+          case openUnionMatch @NotSetup errs of
             Just err -> do
               CLI.Error.put err "Fission is installed, but you are not logged in. Please run `fission login`"
               raise errs
--- a/fission-cli/library/Fission/CLI/Connected.hs
+++ b/fission-cli/library/Fission/CLI/Connected.hs
@@ -165,7 +165,7 @@ mkConnected inCfg ipfsTimeout = do
             proof <- getRootUserProof
             attempt (sendAuthedRequest proof whoAmI) >>= \case
               Left err -> do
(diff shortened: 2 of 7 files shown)


fission-cli/library/Fission/CLI/App.hs
```diff
@@ -44,22 +45,33 @@ interpret baseCfg cmd = do
...
-        Left errs -> do
+      attempt CLI.Env.get >>= \case
+        Left errs ->
           case openUnionMatch @(NotFound FilePath) errs of
-            Just _ -> do
-              logDebug @Text "Setting up new app"
-              run' $ Handler.appInit appDir buildDir mayAppName
+            Just err -> do
+              CLI.Error.put err "Fission is installed, but you are not logged in. Please run `fission login`"
+              raise errs
```
expede, line 53: Hmm we have this exact check in two places just in this PR (maybe more elsewhere?) I imagine that it will be fairly common as we add cases. We can also add some more semantic context to the re-raise (e.g. `ExpectedToBeSetup`, but like... maybe with a better name?)

We have a function elsewhere called `ensureNotLoggedIn` (which makes an HTTP request and whatnot). Perhaps we could extend that convention here. There's a bunch of approaches for this, but a guard may make sense since we're in a context equipped with Rescue:

```haskell
data NotSetup = NotSetup

guardIsSetup :: (MonadRescue m, m`Raises` NotSetup, MonadLogger m) => m ()
guardIsSetup = 
  attempt CLI.Env.get >>= \case
     Left errs ->
       case openUnionMatch @(NotFound FilePath) errs of
          Just err -> do
           CLI.Error.put err "ERROR TEXT HERE"
           raise NotSetup

     Right _ ->
       return ()

foo = do
  actionA
  guardIsSetup
  actionB

-- Or, a bit more flexible *and* handles error in one place --

alreadySetup :: (MonadRescue m, MonadLogger m) => m (Either NotSetup ())
alreadySetup = 
  attempt CLI.Env.get >>= \case
     Left errs ->
       case openUnionMatch @(NotFound FilePath) errs of
          Just err -> return $ Left NotSetup
          Nothing -> return $ Right ()

     Right _ ->
       return $ Right () -- We have an alias `ok = Right ()` that you can use as well

inner = do
  actionX
  ensureM alreadySetup -- ensureM is from Rescue
  actionY

outer = do
  actionA
  actionB
  attempt inner >>= \case -- Handle all of them in one place
    Right _ ->
      actionC
      actionD

    Left err ->
      case openUnionMatch @NotSetup errs of
        Just NotSetup -> CLI.Error.put err "ERROR TEXT HERE"
        Nothing -> return ()
```

The second version essentially pushes the checking and messaging to the caller (in our case, I believe that's `Fission.CLI.App.interpret`).

expede (approved):
This is awesome; thank you so much @QuinnWilton 🎉 

I have a suggestion on structuring the error handling (two related options for abstracting it out), but it's also optional

expede:
Closes #569

QuinnWilton:
This PR was discussed on my call with @expede today, and she said to go ahead and merge it once I had a change :)

Issue #561, opened 2021-11-10
Break out UCAN into own package (Haskell)


expede:
Pretty much what it says on the tin. Let's break out what's already there into its own package. For now we can keep this in the monorepo, but it will almost certainly move to a dedicated UCAN org in the near future.

We may need to disentangle a few parts from the rest of `fission-core`, but this is mostly a direct lift-and-shift job.

* [x] Create package
* [x] Move code
  * Do not move the `Authorization` type (it's `fission-web-server` specific)
* [x] Move tests
* [x] Consume in `fission-web-api` and others

PR #571, opened 2022-01-18
Extract hs-ucan from fission-core

matheus23:
This is work on #561 

To do this, some dependencies on fission-specific UCAN stuff had to be broken up. The main point here being the definition of `Fact` and the resource type `Scope Resource`. (In hindsight I think `Scope` actually still belongs to hs-ucan, but now I've already done it this way, and we'll have to change these parts of the code when supporting the new UCAN version anyway.)

That means that we can't directly run `JWT.check` and know how to check delegation validity in UCANs. So I've abstracted that logic out into a `ResourceSemantics` class:

```hs
class ResourceSemantics rsc where
  canDelegate :: rsc -> rsc -> Bool
```

(Which is structurally equivalent to something like a `PartialOrder` typeclass, but I like giving it a dedicated name.)

There are still some parts that need work.

1. [x] I need to still get the tests compiling (and then working?)
2. [x] I don't 100% like how the fission-core and hs-ucan packages separated. E.g. I don't think `Error.AlreadyExists` should be a hs-ucan module.
3. [x] Generally I'll need to look through hs-ucan's `.Internal` modules again, making sure that we really only have modules in there that make sense
4. [x] ~~I think I want to introduce a `Web.JWT` module, which reexports `Web.JWT.Types`.~~ My current thinking is: Let's introduce that file once we have something to put into it.
5. [x] I don't like how fission-core's newtype around the hs-ucan JWT works. Either both `newtype JWT` and `newtype Proof` or both are type aliases. But if I didn't type-alias `Proof`, I'd be taking non-newtyped proofs out of JWTs and just re-newtype them. I'm thinking that this might get obsolete when I start working on hs-ucan anyway. I have some recursion-schemes thoughts.
6. [x] I *really* don't like the module name `Web.JWT.*` for the hs-ucan stuff. **I'd love some input**. Maybe `Web.Ucan.*`?
7. [x] Oh yeah and I need to figure out a better way to do the `Proof.Resolver` class, which now has 3 type parameters instead of one.

And a final TODO item unrelated to hs-ucan or fission-core:
- [x] Get all other packages & tests compiling again


matheus23:
Don't you like it when your TODO list grows while you're chasing TODO items? :P
- [ ] Re-rename everything to use uppercase `UCAN` instead of `Ucan`
- [ ] Abstract over `Web.Ucan.Potency`

matheus23:
It seemed like the `cachix/install-nix-action@v13` was broken. I upgraded it to `v16`, and that seems to be running fine :+1:

hs-ucan/LICENSE
```diff
@@ -0,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
expede, line 1: Thanks for porting the license file! This library should be under Apache 2.

AGPL forces anyone that includes it to make their source available, which is a barrier to use, but good for us as e.g. Amazon can't just modify it and resell their tweaked version without making their changes also available. The "A" portion in particular is essentially a protection for binaries.

Since we're extracting the UCAN bits out in a hope that it will be widely used, including in closed source, so we want something more permissive. Apache 2 is the preferred choice, as essentially MIT updated with a patent covenant. Relicensing to a more permissive license is a thing that we can do.

fission-web-server/library/Fission/Web/Server/User/Destroyer/Class.hs
```diff
@@ -1,10 +1,9 @@
 module Fission.Web.Server.User.Destroyer.Class (Destroyer (..)) where
 
-import           Database.Esqueleto
+import           Database.Esqueleto.Legacy
```
expede, line 3: Thank you! 🙏

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -441,19 +441,14 @@ instance App.Domain.Initializer Server where
 instance App.Content.Initializer Server where
   placeholder = asks appPlaceholder
 
-instance JWT.Resolver Server where
+instance UCAN.Resolver Server where
   resolve cid =
-    IPFS.ipfsCat cid >>= \case
+    IPFS.ipfsCat cid <&> \case
```
expede, line 446: `<&>` has caused a lot of grief over the years. We standardized to `>>=`, since it's easy to miss the `<&>` and wonder why the typechecker is complaining. Plus GHC will do this rewrite at compile time for us.

Historically _I_ have used a lot of `<&>` in this exact scenario, but it's probably best avoided. I think maybe stylish-haskell recommends these eta reductions, and we should probably configure that out.

fission-web-server/library/Fission/Web/Server/Handler/User/Verify.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
expede, line 1: Do we need this here, or was it recommended by the GHC when the code was in an intermediate state?

fission-web-server/library/Fission/Web/Server/Authorization/Types.hs
```diff
@@ -27,7 +27,7 @@ data Heroku = Heroku
 data Authorization = Authorization
   { sender   :: Either Heroku DID
   , about    :: Entity User
-  , potency  :: Potency
+  , potency  :: Maybe Potency -- Nothing indicates "just used for signature checking"
```
expede, line 30: Rather than dealing with the `Maybe` for all possible call sites in the future, could we consider either abstracting over `Potency`, or parsing that case to a different data type? It's not usable if only checking the signature, no? This type is specialized to look at the database with `User`, so it's not a general use thing.

fission-core/test/Fission/Test/Web/Auth/Token/UCAN/Resource.hs
```diff
@@ -0,0 +1,29 @@
...
+  describe "Resource" do
+
+    describe "serialization" do
+      itsProp' "serialized is isomorphic to ADT" \(resource :: Resource) ->
+        JSON.eitherDecode (JSON.encode resource) `shouldBe` Right resource
+
+    describe "has partial order properties on DelegationSemantics" do
+
+      itsProp' "x canDelegate x"
+        (Proof.Properties.reflexive @(Scope Resource))
```
expede, line 23: You really dislike the `$` operator, eh?

fission-core/test/Fission/Test/Web/Auth/Token/Bearer.hs
```diff
@@ -66,13 +66,14 @@ spec =
             result `shouldBe` jwt
 
 isValidChar :: Word8 -> Bool
-isValidChar w8 = Lazy.elem w8 (" " <> validB64URLChars)
+isValidChar w8 = Lazy.elem w8 (" " <> validEncodedJWTChars)
 
-validB64URLChars :: Lazy.ByteString
-validB64URLChars = Lazy.Char8.pack chars
+validEncodedJWTChars :: Lazy.ByteString
+validEncodedJWTChars = Lazy.Char8.pack (base64URLChars <> ['.']) -- dot is used as a separator in JWTs
```
expede, line 72: RE breaking out the `.` from `base64URLChars` — Fair!

hs-ucan/test/Test/Web/UCAN/Example.hs
```diff
@@ -0,0 +1,67 @@
...
+  | Everything
+  deriving (Show, Eq, Ord)
+  deriving DelegationSemantics
+    via (GreaterDelegatesMore Resource)
+
+data Potency
+  = CanLook
+  | CanTouch
+  | SuperUser
+  deriving (Show, Eq)
```
expede, line 24: OMG these are so great, I may use them in the spec 😛

fission-web-server/library/Fission/Web/Server/Types.hs
matheus23: It's actually `hlint` usually suggesting things, but I've turned it off for me.
Yeah, I was thinking "we can get rid of two `return`s here", but I agree that's really not worth it and going for a monadic lambda makes sense generally! :+1:

fission-web-server/library/Fission/Web/Server/Handler/User/Verify.hs
matheus23: I swear I had GHC barking at me for this once, but it must've been a cascading effect from somewhere else. It's not needed anymore. Removing it :+1:

fission-core/test/Fission/Test/Web/Auth/Token/UCAN/Resource.hs
matheus23: :laughing: Lemme fix that, it just didn't occur to me (but also, yes I kind of do :P )

fission-web-server/library/Fission/Web/Server/Authorization/Types.hs
matheus23: > Let's make error cases as unrepresentable as possible, if possible!
The new `Potency` type lacks a variant the old one had before: `AuthNOnly`. So basically `Old.Potency` is isomorphic to `Maybe New.Potency`. So I don't think there should be any changes in terms of what's representable. I basically moved the comment for `Nothing` over from the definition of the old `Potency` type.

---

In any way, as soon as hs-ucan is updated to UCAN 0.8 and I'm incorporating that into the fission stuff, this will change.

fission-cli/library/Fission/CLI/Types.hs
```diff
@@ -812,17 +814,15 @@ instance
...
     _ <- IPFS.Daemon.runDaemon
     IPFS.runLocal ["cat"] (Lazy.fromStrict $ encodeUtf8 hash') <&> \case
       Left errMsg ->
         Left $ CannotResolve cid (ConnectionError $ toException errMsg)
 
       Right (Lazy.toStrict -> resolvedBS) ->
-        case eitherDecodeStrict resolvedBS of
-          Left  _   -> Left $ InvalidJWT resolvedBS
-          Right jwt -> Right (JWT.contentOf (decodeUtf8Lenient resolvedBS), jwt)
+          Right (UCAN.contentOf (decodeUtf8Lenient resolvedBS))
```
matheus23, line 825: Here you can see an effect of one of the changes.
I've changed the `UCAN.Resolver` class to only resolve to `UCAN.RawContent`, instead of also decoding the JWT.

fission-core/library/Fission/Internal/UTF8.hs
```diff
@@ -5,116 +5,37 @@ module Fission.Internal.UTF8
...
-  , stripQuotes
-  , stripQuotesBS
-  , stripQuotesLazyBS
-  , stripN
-  , stripNBS
-  , stripNewline
+  , toBase58Text
   , textShow
   , wrapIn
+  , module Web.UCAN.Internal.UTF8
```
matheus23, line 12: I wasn't quite sure what to do with these `.Internal` stuff that we've been depending on in fission-core.

So what I did was split it up, you can see me removing stuff from the fission-core `.Internal.UTF8` module here, and instead reexporting the `Web.UCAN.Internal.UTF8` module, where that stuff was moved to.

fission-core/library/Fission/Web/Auth/Token/UCAN/Potency/Types.hs
```diff
@@ -12,30 +16,30 @@ append-only rights, so
 
 -- | How much power allowed in an authorization
 data Potency
-  = AuthNOnly   -- ^ Read signature only -- just a proof. Cannot delegate further.
```
matheus23, line 15: I removed the `AuthNOnly` constructor here, in favor of using `Maybe Potency` instead of `Potency` everywhere.
The reason is that in `hs-ucan` we don't want to depend on `Potency` (because it's also fission-specific), but we *need* to decode to *some* potency in case the `ptc` key is missing completely.

fission-core/library/Fission/Web/Auth/Token/UCAN/Resource/Types.hs
```diff
@@ -50,5 +56,53 @@ instance FromJSON Resource where
...
+            False
+
+      (RegisteredDomain _, RegisteredDomain Complete) ->
+        True
+
+      (FissionApp _, FissionApp Complete) ->
+        True
+
+      _ ->
+        rsc == rscProof
```
matheus23, line 80: This is code adapted from a previous function `Fission.Web.Auth.Token.JWT.Proof.WNFS.compareSubsets`

hs-ucan/library/Web/UCAN/Resolver/Class.hs
```diff
@@ -0,0 +1,11 @@
...
+
+import           Network.IPFS.CID.Types
+
+import           RIO
+
+import qualified Web.UCAN.RawContent     as UCAN
+import           Web.UCAN.Resolver.Error as Resolver
+
+class Monad m => Resolver m where
+  resolve :: CID -> m (Either Resolver.Error UCAN.RawContent)
```
matheus23, line 11: Here's what I've been talking about (above?): The `Resolver` class now returns
`Either Resolver.Error UCAN.RawContent` instead of
`Either Resolver.Error (UCAN.RawContent, UCAN.UCAN fct ptc rsc)`,
which just makes it a little simpler by getting rid of some dependencies on what `fct`, `ptc` and `rsc` are.

fission-core/library/Fission/Internal/UTF8.hs
expede: Hmm good point! A bunch of this stuff is actually now available in other libraries. We could switch to those, or break out a `fission-utils` or something 🤔

fission-core/library/Fission/Web/Auth/Token/UCAN/Potency/Types.hs
expede: Yeah, for this version of UCAN totally.

Thinking out loud: as we bring things to up date, I imagine that we'll have something like...

```haskell
data Capabilities custom
  = None
  |  UserDefined custom
  | AllByParenthood -- e..g Linking
```

...but obviously we need to actually solidify that in the v0.8 spec before we do that.

matheus23:
I've changed the base branch to `ucan-upgrade` as per discussion with @expede: We'll keep that as another branch running alongside main while we're upgrading hs-ucan, so we can merge partially-typechecking PRs while stuff is in flux.
We keep main clean so we can hotfix stuff.
Eventually `ucan-upgrade` will get merged into main.

fission-cli/library/Fission/CLI/Types.hs
expede: Hmm I have some vague memories of also going down this path (several times?) and then putting the JWT back in 🤔 I can't remember what the scenario was — something about needing the decoded case in various code paths? — so not going to say to not do this, but just mentioning it

fission-core/library/Fission/Web/Auth/Token/UCAN/Resource/Scope/Types.hs
```diff
@@ -2,11 +2,19 @@ module Fission.Web.Auth.Token.UCAN.Resource.Scope.Types (Scope (..)) where
...
+
 data Scope subset
   = Complete
   | Subset subset
   deriving (Eq, Ord, Show)
 
+instance DelegationSemantics rsc => DelegationSemantics (Scope rsc) where
+  Complete          `canDelegate` _            = True
+  (Subset _)        `canDelegate` Complete     = False
+  (Subset rscProof) `canDelegate` (Subset rsc) = rscProof `canDelegate` rsc
```
expede, line 16: Nice

expede (approved):
@matheus23 this looks great 🎉 

It's a lot of _files_, but the changes are extremely straightforward 👍

fission-cli/library/Fission/CLI/Types.hs
matheus23: Okay. I just now realized that `RawContent` isn't the *whole* JWT encoded. It's only the part without the signature. That's what I get for not reading its docs!
I'll fix this in another PR. I'm thinking that the easiest way is to return either `ByteString` from `resolve`, or some kind of newtype?

PR #573, opened 2022-01-25
Upgrade DID RSA Key format

matheus23:
Ported over most changes from #550 but adapted to the format with the prefix bytes `0x85 0x24` which is the unsigned varint encoding of `0x1204`.
Also, instead of storing the ASN1 DER `SubjectPublicKeyInfo`, it stores ASN1 DER encoded `RSAPublicKey`s (which are part of `SubjectPublicKeyInfo`s with RSA public keys) as per the DID spec (https://github.com/w3c-ccg/did-method-key/pull/41).

Depends on #571. Currently that's the "branch this will merge into" for better diffs on github, but I'll change this to the branch that #571 will merge into once that's merged.


matheus23:
Branch shenanigans resolved!
This PR leaves the whole monorepo in compiling state, so we can merge it into `main`. The `ucan-upgrade` branch will be initiated later.

fission-core/library/Fission/Internal/Orphanage/RSA2048.hs
```diff
@@ -7,5 +7,7 @@ import qualified System.IO.Unsafe      as Unsafe
 
 import           Fission.Prelude
 
+
+-- Wait. This is weird. How did this end up in this module?
 instance Arbitrary Ed25519.SecretKey where
```
expede, line 12: Yup that is weird! It's probably me being confused, but the compiler warning is turned off because `{-# OPTIONS_GHC -fno-warn-orphans #-}`. All the more reason for us to just bite the bullet and newtype everything. The more that I think about it, the more I'm in favour of just doing that.

hs-ucan/library/Web/UCAN/Internal/Orphanage/RSA2048/Public.hs
```diff
@@ -35,15 +36,14 @@ instance Arbitrary RSA.PublicKey where
...
-      & PEM.pemWriteBS
-      & decodeUtf8Lenient
-      & Text.strip
-      & Text.dropPrefix pemHeader
-      & Text.dropSuffix pemFooter
-      & Text.filter (/= '\n')
+      & PEM.pemContent
+      & BS64.decodeLenient
+      & Builder.byteString
+      & Utf8Builder
```
expede, line 46: Oh yeah, that's way better 🎉

hs-ucan/test/Test/Prelude.hs
```diff
@@ -1,4 +1,4 @@
-module Test.Web.UCAN.Prelude
+module Test.Prelude
```
expede, line 1: I would scope this down. `Test.Prelude` is just so general that a namespace collision feels likely. I also see a lot of stuff like `Test.AWS.Prelude` in the wild (https://www.stackage.org/haddock/lts-18.23/amazonka-test-1.6.1/Test-AWS-Prelude.html)

hs-ucan/library/Web/UCAN/Internal/Orphanage/RSA2048/Public.hs
expede, line 46: Oh yeah, that's way better 🎉

hs-ucan/library/Crypto/Key/Asymmetric/Public.hs
```diff
@@ -36,3 +46,43 @@ genRSA2048 = do
...
+  case ASN1.toASN1 (X509.PubKeyRSA pk) [] of
+    [ ASN1.Start ASN1.Sequence
+      , ASN1.Start ASN1.Sequence
+      , ASN1.OID [1,2,840,113549,1,1,1]
+      , ASN1.Null
+      , ASN1.End ASN1.Sequence
+      , ASN1.BitString bitArray
+      , ASN1.End ASN1.Sequence
+      ] ->
+        ASN1.bitArrayGetData bitArray
```
expede, line 85: Not that it's a big problem, but kind of weird alignment here, no?

expede: (Lines 78-84)

nix/sources.json
```diff
@@ -5,10 +5,10 @@
         "homepage": "",
```
expede, line 5: I really appreciate the version bump, but this breaks on `darwin-aarch64` 😭 I needed to roll back the `nix/` directory to get one of the `ipfs` dependencies to build

shell.nix
```diff
@@ -43,8 +42,6 @@
     ];
 
     haskell = [
-      unstable.ghc
-      unstable.haskellPackages.implicit-hie
```
expede, line 47: 👍

fission-web-server/library/Fission/Web/Server/Handler/User/Verify.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE AllowAmbiguousTypes #-}
```
expede, line 1: I removed every instance of this pragma from the code on my copy (there were 3 of them) and everything compiled. Can you remove this one and the others? Thanks!

expede: I know that GHC suggests this in some circumstances, especially while code is in flux. It's not the worst extension in the world, but also we shouldn't have it if we're not using it.

hs-ucan/library/Web/UCAN/Internal/Orphanage/RSA2048/Public.hs
matheus23: Haha, you're complementing your own code in this case though! :P (I've copied this over from the other PR)

hs-ucan/library/Crypto/Key/Asymmetric/Public.hs
matheus23: Yes, I agree. But for some very weird reason, it doesn't parse with the correct alignment :(
image

nix/sources.json
matheus23: Sure! I can totally roll back no problem. I just noticed that there was a `niv update` in the other PR and thought it wasn't a bad idea. I don't think there was another reason for me to upgrade.

fission-web-server/library/Fission/Web/Server/Handler/User/Verify.hs
matheus23: :facepalm:  Yeah for some reason while code is in flux, I constantly need to add this extension to get the *next* compiler errors. But I frequently forget to remove it again. Will do!

matheus23:
@expede Github Actions continues making my life hard by being somewhat flakey:

> /home/runner/.stack/setup-exe-cache/x86_64-linux-nix/Cabal-simple_mPHDZzAJ_3.2.1.0_ghc-8.10.7: startProcess: exec: invalid argument (Bad file descriptor)

This only happened in the "🏺 Artifacts (Nix) / 🖥️ ubuntu-latest ❄️ Nix (pull_request)" action. All other actions are running through (including tests now!).

So I think this is ready for another round of review :+1: (no rush but just to make sure you don't think this blocked on me getting the tests working or something)

fission-core/library/Fission/Internal/Orphanage/RSA2048.hs
expede: Maybe to clarify: let's make this change in this PR. This line _should not_ be in this module!

expede (approved):
Just move that orphan, and then 🚀 🚀 🚀 🚀 🚀 🚀 🚀

matheus23:
:joy: Thanks for the enthousiasm.

I need to change/revert *one* think, which is the RSA DID *encoding* should be rolled back to the old format *for now*, because - as you've experienced - it breaks e.g. the AWAKE protocol, as long as the auth lobby doesn't understand the new DID format.

I'll merge it later today! (Let's not have another PR hanging like the last one :scream: )

PR #575, opened 2022-01-26
Put current `ucan-upgrade` state into `main`

matheus23:
I was a little too fast with changing the base branch of PR #571 . Since #517 was fully compiling & test running green & a pure refactor (no changes in terms of function), I think it should've been committed to main.

We only need the parallel `ucan-upgrade` branch as soon as we're in a state where the monorepo isn't in a compiling case, but we want to iterate on the hs-ucan package.


expede (approved):

expede:
Agreed!

PR #579, opened 2022-02-03
PowerDNS Support

### LINKED ISSUE #574: Server support for PowerDNS
# Summary

## Problem

AWS Route53 limits the number of records per zone which means there is an upper bound on users / apps available.

## Solution

To start, we'll add support for PowerDNS (which also has a [REST API](https://doc.powerdns.com/authoritative/http-api/index.html)). Support for other potential DNS providers / servers is possible, but we'll start here.

walkah:
- [x] Read pdns config from env.yaml
- [x] Update CNAME / TXT records
- [x] Implement PowerDNS.get / clear
- [x] `syncDNS` task to sync TXT/CNAME records from DB

Closes #574


fission-web-server/library/Fission/Web/Server/PowerDNS.hs
```diff
@@ -0,0 +1,14 @@
...
+import           Fission.Prelude
+import           Fission.Web.Server.PowerDNS.Class
+
+import           Data.List                         (head)
+import           PowerDNS.Client                   as PDNS
+
+getValuesFromRecords :: [SearchResult] -> Maybe (NonEmpty Text)
+getValuesFromRecords rrs = case rrs of
+  []     -> Nothing
+  result -> Just $ pure (PDNS.sr_content $ head result)
```
expede, line 14: I see the Schemer in you coming out 😁

`head` is generally avoided, because it's a partial function; if you run `head []`, it'll blow up. `NonEmpty.head :: NonEmpty a -> a` on the other hand is totally safe, because it guarantees that there's at least one value (look at the constructor to see why)

In this case, it's more common to use pattern matching:

```haskell
case rrs of
  [] -> Nothing
  (h : _) -> Just . pure $ PDNS.sr_content h
```

There's also a helper function for this: [`nonEmpty :: [a] -> Maybe (NonEmpty a)`](https://www.stackage.org/haddock/lts-18.24/base-4.14.3.0/Data-List-NonEmpty.html#v:nonEmpty). (Given that this is "get _values_ from records" (plural), do you only want the first value?)

getValuesFromRecords :: [SearchResult] -> Maybe (NonEmpty Text)
getValuesFromRecords rrs = PDNS.sr_content <$> nonEmpty rrs

-- Lots of ways to phrase this one, so one example:
getOneValueFromRecords :: [SearchResult] -> Maybe (NonEmpty Text)
getOneValueFromRecords rrs = do
  (x :| _) <- nonEmpty rrs
  return $ PDNS.sr_content x

fission-web-server/library/Fission/Web/Server/PowerDNS/Class.hs
```diff
@@ -0,0 +1,29 @@
...
+    -> m (Either ServerError NoContent)
+
+  get ::
+       URL
+    -> Text -- ZoneID
+    -> m (Either ServerError [SearchResult])
+
+  clear ::
+       URL
+    -> Text -- ZoneID
```
expede, line 28: If you write `-- ^ ZoneID` (with the `^`), it'll show up in the docs, too!

fission-web-server/library/Fission/Web/Server/PowerDNS/Types.hs
```diff
@@ -0,0 +1,27 @@
...
+
+newtype URL = URL { getUrl :: Text }
+  deriving newtype ( Eq
+                   , Show
+                   , FromJSON
+                   , IsString
+                   )
+
+instance Display URL where
+  textDisplay = getUrl
```
expede, line 16: 👩‍🏫 This is the same as adding `Display` to the `deriving newtype`

fission-web-server/library/Fission/Web/Server/PowerDNS/Types.hs
```diff
@@ -0,0 +1,27 @@
...
+
+newtype ApiKey = ApiKey { getApiKey :: Text }
+  deriving newtype ( Eq
+                   , Show
+                   , FromJSON
+                   , IsString
+                   )
+
+instance Display ApiKey where
+  textDisplay = getApiKey
```
expede, line 26: Same here!

fission-web-server/library/Fission/Web/Server/PowerDNS/Types.hs
```diff
@@ -0,0 +1,27 @@
+module Fission.Web.Server.PowerDNS.Types
+  ( URL (..)
+  , ApiKey (..)
+  ) where
+
+import           Fission.Prelude
+
+newtype URL = URL { getUrl :: Text }
```
expede, line 8: Yaaaas `newtype`!

expede: This instead of our existing `URL` type because it's less structured I guess?

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -140,6 +155,102 @@ instance MonadDB (Transaction Server) Server where
...
+    apiKey <- asks pdnsApiKey
+    uri <- parseBaseUrl $ Text.unpack (textDisplay apiUrl)
+    let
+      rawHTTPSettings =
+        case baseUrlScheme uri of
+          Http  -> defaultManagerSettings
+          Https -> tlsManagerSettings
+
+    mgr <- liftIO $ newManager rawHTTPSettings
+    return $ PDNS.applyXApiKey (textDisplay apiKey) (mkClientEnv mgr uri)
```
expede, line 170: We should move most of this to `Main.hs`, and thread the HTTP manager through in the `Config`. This is setting up a connection pool, just like with Postgres. So right now, every time you make a DNS call, we're doing all that work over again.

Having a `getClientEnv` method is totally fine, if you want that to be part of the public API, but change it to be something like `asks pdnsManager`. I would probably remove it from the public API, personally, and create a helper function in this module that gets called in the class methods.

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -140,6 +155,102 @@ instance MonadDB (Transaction Server) Server where
...
+      ]
+    clientEnv <- getClientEnv
+    let
+      rrset = RRSets [RRSet { rrset_name = Text.append (textDisplay url) "."
+                      , rrset_type = recType
+                      , rrset_ttl = ttl
+                      , rrset_changetype = Just Replace
+                      , rrset_records = Just $ formatRecords contents
+                      , rrset_comments = Nothing
+      }]
```
expede, line 189: We tend to left-align brackets in Haskell (weird, but pretty)

```haskell
let
  rreset = 
    RRSet { rreset_name = textDisplay url <> "." -- The `<>` here is equivalent 
          , rrset_type = recType
          , rrset_ttl = ttl
      --  , ...
          }

  rresets =
    RRset [ rrset ]
```

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -140,6 +155,102 @@ instance MonadDB (Transaction Server) Server where
...
+    resp <- liftIO $ runClientM (PDNS.updateRecords "localhost" zoneTxt rrset) clientEnv
```
expede, line 190: Hardcoded `localhost`?

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -140,6 +155,102 @@ instance MonadDB (Transaction Server) Server where
...
+                      , rrset_records = Nothing
+                      , rrset_comments = Nothing
+          }]
+        _ <- liftIO $ runClientM (PDNS.updateRecords "localhost" zoneID rrset) clientEnv
+        return ()
+
+      convertType :: Text -> PowerDNS.API.Zones.RecordType
+      convertType "CNAME" = CNAME
+      convertType "TXT"   = TXT
+      convertType _       = PowerDNS.API.Zones.A
```
expede, line 251: This is totally fine, but some background:

This is typically covered by the `Read` typeclass, but it looks like they didn't implement that. You can plug into their parser by leveraging the JSON instances:

```haskell
decode (toJSON txt) :: Maybe RecordType

-- With a fallback to an A record, though not totally sure that's what you want?

maybe Zones.A $ decode $ toJSON txt :: RecordType
```

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -272,25 +384,25 @@ instance MonadWNFS Server where
...
+        case PowerDNS.getValuesFromRecords results of
           Nothing   -> return . Left $ Web.Error.toServerError (404 :: Int)
           Just vals -> return $ Right $ extractCID vals
     where
       extractCID = IPFS.CID . Text.dropPrefix "\"dnslink=/ipfs/" . Text.dropSuffix "\"" . NonEmpty.head
 
 instance MonadDNSLink Server where
   set _userId url@URL {..} zoneID (IPFS.CID hash) = do
-    Route53.set Cname url zoneID (pure $ textDisplay gateway) 86400 >>= \case
+    PowerDNS.set CNAME url (textDisplay zoneID) (pure $ Text.append (textDisplay gateway) ".") 86400 >>= \case
```
expede, line 400: `Text.append` here can be replaced with the infix `<>`

expede (changes requested):
Looking pretty good! There's mostly the one request for change to set up the connection manager once, and then some (hopefully) helpful tips!

Also totally happy to walk through the code in our call later 👍 

Well done! 👏

fission-web-server/library/Fission/Web/Server/Types.hs
walkah: I think I've done this now.

fission-web-server/library/Fission/Web/Server/Types.hs
walkah: yeah, this is weird, technically you can use the pdns http API to control multiple pdns servers ... we're likely only going to ever control the local one. (i.e. this doesn't mean run anything against the server's localhost it's actually part of the url path in all REST calls).

expede: WUT

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
```diff
@@ -45,7 +54,71 @@ deleteByUsername userNameTxt =
         Nothing                -> error "User doesn't exist"
 
 ---
+syncDNS :: Server()
```
expede, line 57: 💅 Purely aesthetic:

In Haskell, `()` is a type on its own. We put a space between data types

```haskell
-- Looks super weird
Server()

-- Fails to parse
MaybeInt

-- Looks good
Server ()
Maybe Int
```

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
```diff
@@ -45,7 +54,71 @@ deleteByUsername userNameTxt =
...
+syncDNS = do
+  users <- getUserData
+  apps <- getAppData
+  mapM_ setUserDNS users
+  mapM_ setAppDNS apps
```
expede, line 62: This totally works! No change required _at all_, but I find that `forM_` reads nicely in these situations:

```haskell
-- Actually any traversable, not just lists
mapM :: Monad m => (a -> m ()) -> [m a] -> m ()
forM :: Monad m => [m a] -> (a -> m ()) -> m ()

syncDNS :: Server ()
syncDNS = do
  users <- getUserData
  apps  <- getAppData
  forM_ users userUserDNS
  forM_ apps  setAppDNS
```

In English: "for the apps, set app DNS"

Just for completeness, there are other phrasings...

```haskell
syncDNS :: Server ()
syncDNS = do
  forM_ users =<< getUserData
  forM_ apps  =<< getAppData
```

...but I think that your existing version where you run the queries first is clearer

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
expede: Perhaps to clarify the role of `()` 👇 

This is called "unit", and is has the type `() :: ()` (type-level punning makes the term and type read the same in this case). It has only the one value: `()`, in the same way that `Bool` has two values.

```haskell
data () = ()
data Bool = False | True
```

`()` carries very little information. As such, it can be used to stub out a type. For example, these types carry the same data (though they are counted as different by the compiler):

```haskell
Maybe a ~ Either () a
```

Since it carries no information of its own, we can always get a unit. For example, this would be totally valid (though stupid, but maybe good for building intuition):

```haskell
ignore :: a -> ()
ignore _ = ()

ignore 42
-- ()

ignore "this is a string"
-- ()

ignore ()
-- ()

ignore (Just (Just (Right "hi")))
-- ()
```

Being FP, we always need to return a value, so especially when running side effects with no return value of their own, we use `()` to represent that "nothing" was returned.

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
```diff
@@ -45,7 +54,71 @@ deleteByUsername userNameTxt =
...
+setUserDNS :: Entity User -> Server()
+setUserDNS (Entity userId User { userUsername, userPublicKey, userDataRoot }) = do
+  zoneID <- asks userZoneID
+  domainName <- asks userRootDomain
+  let
+    unSubDom = Subdomain $ textDisplay userUsername
+    url      = URL {domainName, subdomain = Just (Subdomain "_did") <> Just unSubDom}
+    segments = case userPublicKey of
+      Nothing   -> pure ""
+      Just pkey -> DNS.splitRecord $ textDisplay (DID Key pkey)
```
expede, line 87: This is SUCH a nit, but I've seen this in a few places in this PR and it makes my eye twitch 😛 

If you could put a blank line between an indented and the next outdented line, it's pretty idiomatic

fission-web-server/library/Fission/Web/Server/Internal/Tasks.hs
```diff
@@ -45,7 +54,71 @@ deleteByUsername userNameTxt =
...
+  let
+    url = URL
+      { domainName = appDomainDomainName
+      , subdomain = appDomainSubdomain
+      }
+  DNSLink.set appOwnerId url zoneID appCid >>= \case
+    Left _ -> return ()
+    Right _ ->
+      logDebug $ "Successfully updated " <> show appDomainSubdomain
+---
```
expede, line 121: Same here, feel free to add some space around the `---`. If you like C-style sections, you can also make pretty `--` boxes

```haskell
logDebug $ "Successfully updated " <> show appDomainSubdomain

---------------------
-- IPFS Migration --
---------------------

ensureAllPinned :: Server ()
```

(Though in many cases, that's a sign that you would want another module... but hey, this is helper scripts so _whatever_ 😛 )

fission-web-server/library/Fission/Web/Server/PowerDNS.hs
```diff
@@ -0,0 +1,12 @@
+module Fission.Web.Server.PowerDNS
+  ( module Fission.Web.Server.PowerDNS.Class
+  , getValuesFromRecords) where
```
expede, line 3: Is your formatter doing this? 🤔 

```haskell
module Fission.Web.Server.PowerDNS
  ( module Fission.Web.Server.PowerDNS.Class
  , getValuesFromRecords
  ) where
```

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -140,6 +155,106 @@ instance MonadDB (Transaction Server) Server where
...
+        RRSet { rrset_name = Text.append (textDisplay url) "."
+              , rrset_type = recType
+              , rrset_ttl = ttl
+              , rrset_changetype = Just Replace
+              , rrset_records = Just $ formatRecords contents
+              , rrset_comments = Nothing
+              }
+      rrsets = RRSets [rrset]
+    resp <- liftIO $ runClientM (PDNS.updateRecords "localhost" zoneTxt rrsets) clientEnv
+    case resp of
```
expede, line 180: Fine as-is, but another phrasing when you're going to use the result immediately:

```haskell
liftIO (runClientM (PDNS.updateRecords "localhost" zoneTxt rrsets) clientEnv) >>= \case
  Left err -> -- ...
```

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -140,6 +155,106 @@ instance MonadDB (Transaction Server) Server where
...
+      Left err -> do
+        logWarn @Text "PowerDNS.set failed"
+        return . Left $ Web.Error.toServerError err
+      Right good -> do
+        logDebug @Text "PowerDNS.set succeeded"
+        return $ Right good
+    where
+      formatRecords :: NonEmpty Text -> [Record]
+      formatRecords values =
+        fmap newRecord (toList values)
```
expede, line 190: 👩‍🏫 100% totally fine to write as you have it here. You can use infix backticks or the `<$> = fmap` alias  to break up the line instead of brackets:

```haskell
newRecord `fmap` toList values
newRecord <$> toList values
```

expede: The pun here is that `$` is infix application, so `<$>` is mapping (i.e. application to all elements in a container).

Same goes for `&` and `<&>`, which are the same but with arguments reversed

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -140,6 +155,106 @@ instance MonadDB (Transaction Server) Server where
...
+
+      newRecord :: Text -> Record
+      newRecord content =
+        Record { record_content = format recType content, record_disabled = False }
+
+      format :: PowerDNS.API.Zones.RecordType -> Text -> Text
+      format TXT = UTF8.wrapIn "\""
+      format _   = identity
```
expede, line 198: Just the TXT records? 🤔 I'll read the rest, but when I see `format` in Haskell, I immediately wonder if there's a class we could use, like `ToJSON` or `Display`. Probably overkill for now, but conceptually we may later want a `class ToDNS a`.

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -140,6 +155,106 @@ instance MonadDB (Transaction Server) Server where
...
+
+  clear url zoneID = do
+    logDebug $ "PowerDNS.clear " <> textDisplay url
+    results <- PowerDNS.get url zoneID
+    case results of
+      Left err -> do
+        return $ Left err
+      Right searchResults -> do
+        mapM_ removeResult searchResults
+        return $ Right ()
```
expede, line 220: Same comments as above RE empty lines and `>>= \case` (as an option!)

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -140,6 +155,106 @@ instance MonadDB (Transaction Server) Server where
...
+  apiKey <- asks pdnsApiKey
+  uri <- parseBaseUrl $ Text.unpack (textDisplay apiUrl)
+  let
+    rawHTTPSettings =
+      case baseUrlScheme uri of
+        Http  -> defaultManagerSettings
+        Https -> tlsManagerSettings
+
+  mgr <- liftIO $ newManager rawHTTPSettings
+  return $ PDNS.applyXApiKey (textDisplay apiKey) (mkClientEnv mgr uri)
```
expede, line 256: This function should get called once in `Main.hs`, and then the resulting `rawHttpSettings` get threaded through the code via `Server`'s internal `Config`. You _may_ be able to reuse on of these:

https://github.com/fission-suite/fission/blob/c7f5907867e94b76bdcbed3a499a4128263d7bfd/fission-web-server/library/Fission/Web/Server/Config/Types.hs#L35-L36

Which would look something like:

```haskell
do
  -- ...
  httpMgr  <- asks httpManager
  httpsMgr <- asks tlsManager
```

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -140,6 +155,106 @@ instance MonadDB (Transaction Server) Server where
...
+    logDebug $ "PowerDNS.clear " <> textDisplay url
+    results <- PowerDNS.get url zoneID
+    case results of
+      Left err -> do
+        return $ Left err
+      Right searchResults -> do
+        mapM_ removeResult searchResults
+        return $ Right ()
+    where
+      removeResult :: (MonadIO m, MonadReader Config m, MonadThrow m) => PDNS.SearchResult -> m ()
```
expede, line 222: Amazing that you're diving right into abstraction! 👏 

No need to change anything, but we're also working directly in `Server a`, which has instances for all of these. So _in this context,_ we don't need to generalize:

```haskell
removeResult :: Server ()
```

A lot of our other code is pretty general, because we want to be able to plug in other contexts, such as test environments that stub out DNS access and whatnot. Actually with this PDNS stuff, we can possibly think about making a lot of the code more concrete, and "just" using `Server a` everywhere since we can spin up DNS locally 🤔🤔🤔🤔🤔🤔🤔

(It _is_ nice to have the ability to easily stub out anything in a super clean, first class way, but also it has typeclass overhead, so worth thinking about after this PR makes it possible)

expede: > removeResult :: Server ()

Let me know if why you can do this here is confusing, and we can chat about it 👍

fission-web-server/library/Fission/Web/Server/Types.hs
```diff
@@ -455,39 +571,20 @@ instance ServerDID Server where
...
-          Right resp -> do
-            let status' = view crrsrsResponseStatus resp
-            if status' < 300
-              then ok
-              else Left $ Web.Error.toServerError status'
+    PowerDNS.set TXT txtRecordURL (textDisplay zoneID) (pure txtRecordValue) 10 <&> \case
+      Left err ->
+        Left err
+      Right _ -> do
+        ok
```
expede, line 587: It's _fine,_ but also you don't need the `do` here.

I'd also recommend against using `<&> \case` over `>>= \case` in monadic contexts. GHC will optimize this in either case, but right now you can't just drop in and do this:

```haskell
PowerDNS.set TXT txtRecordURL (textDisplay zoneID) (pure txtRecordValue) 10 <&> \case
  Left err -> do
    logError "BOOM
    return $ Left err

  Right -> do
    logInfo "Success"
    return ok
```

...because `<&>` is an inverse `fmap` with arguments reversed. To do more than one "action" here, you'd need to switch to `>>=`.

I used to be a heavy user of `<&> \case` because it's shorter, but it's bitten _me_ a bunch ver the years, yelling at the type checker not realizing that I'm literally 3 characters away from having it compile.

stack.yaml
```diff
@@ -32,6 +33,6 @@ ghc-options:
   "$everything": -haddock
 
 nix:
-  enable:     true
-  pure:       true
+  enable: true
+  pure: true
```
expede, line 37: haha okay FAIR: I do go a little overboard with the alignment

fission-web-server/library/Fission/Web/Server/Types.hs
expede: (`asks` pulls values out of the `Config` contained in `Server`. This is the `MonadReader Config Server` class instance AKA "it knows how to access the ambient configured values without explicitly threading them around everywhere)

fission-web-server/library/Fission/Web/Server/PowerDNS/Types.hs
walkah: Yeah, we don't ever parse or manipulate this one...

fission-web-server/library/Fission/Web/Server/Types.hs
walkah: 🤦  I added a `pdnsManager` bit to config and then when I refactored this function I forgot to actually *use* it. Fixed now, I think?

fission-web-server/library/Fission/Web/Server/Types.hs
expede: All the more reason to extract it!

...at some point when we're not running out of DNS space

expede (approved):
Looking good!

expede:
@walkah have we tested this exact code on staging?

expede:
> The Fission API (v2.17.0.0)

I guess I can look that up, and the answer is no 😛

expede:
Well crap, I meant to do that on a branch 🤔

stack.yaml
```diff
@@ -1,4 +1,4 @@
-resolver: lts-18.14
+resolver: lts-18.24
```
expede, line 1: "While we have the hood up", we may as well get a bunch of bugfixes

fission-web-server/library/Fission/Web/Server/PowerDNS.hs
```diff
@@ -0,0 +1,11 @@
...
+  ( module Fission.Web.Server.PowerDNS.Class
+  , getValuesFromRecords
+  ) where
+
+import           Fission.Prelude
+import           Fission.Web.Server.PowerDNS.Class
+import           PowerDNS.Client                   as PDNS
+
+getValuesFromRecords :: [SearchResult] -> Maybe (NonEmpty Text)
+getValuesFromRecords rrs = nonEmpty (PDNS.sr_content <$> rrs)
```
expede, line 11: @walkah let me know if you explicitly do not want the rest of the values that are in the `[SearchResult]`. This version keeps them available rather than tossing them. Being a lazy language, we'll only get the ones we consume.

PR #580, opened 2022-02-23
hs-ucan upgrade

matheus23:


matheus23:
Opened #586 in favor of this.

Github would not remove the "Extract hs-ucan from fission-core" commit from the history, even though the base branch already has it. Making the diff unreadable.
Opening exactly the same PR fixes the issue. :man_shrugging:

PR #582, opened 2022-03-03
Ensure github requests always use TLS

walkah:
Note: This issue only really impacts local development. 

We currently set up an `HTTP.Manager` globally based on the scheme of the fission server REST url. This means in local development (when using HTTP), the CLI will try to make a request to github via http. This PR adds a dedicated tlsManager that github requests will always use.


expede (approved):
Looks good to me 👏

PR #586, opened 2022-03-28
hs-ucan upgrade

matheus23:


expede:
In fact (and this is maybe controversial), we could potentially put old styles in this sum https://github.com/fission-suite/fission/pull/587/files#diff-0c1068947ea16611dfdbdf6707885117deab65d004a64590bf36c0b59bd437b3R77

matheus23:
Closing, because diffs render incorrectly in github when you force-push the base branch :(

PR #587, opened 2022-03-29
Upgrade DID multiformat

matheus23:
Split out from #586 

This upgrades the DID multiformat to the new version. We can't deploy this yet, because this PR will have the CLI generate new-style DIDs and browsers need to be upgraded to understand that format first.

What we *could* merge is support for *reading* the new DID RSA format, but *generating* the old RSA format.

However for now this is just some reviewable commits extracted from #586, so we can split reviewing into rounds.


expede:
> We can't deploy this yet, because this PR will have the CLI generate new-style DIDs and browsers need to be upgraded to understand that format first.

Could we break out the format into its own type as another `Oldstyle`, use it at the call sites, and merge the new style code, just to get it in `main`? The advantages are that we don't need to keep rebasing this branch, and flipping over to the new style DIDs would be like 12 LOC (just the call sites).

hs-ucan/library/Web/DID/Types.hs
```diff
@@ -67,29 +65,25 @@ More here: https://github.com/multiformats/unsigned-varint
...
-Right (DID {method = Key, publicKey = MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAnzyis1ZjfNB0bBgKFMSvvkTtwlvBsaJq7S5wA+kzeVOVpVWwkWdVha4s38XM/pa/yr47av7+z3VTmvDRyAHcaT92whREFpLv9cj5lTeJSibyr/Mrm/YtjCZVWgaOYIhwrXwKLqPr/11inWsAkfIytvHWTxZYEcXLgAXFuUuaS3uF9gEiNQwzGTU1v0FqkqTBr4B8nW3HCN47XUu0t8Y0e+lf4s4OxQawWD79J9/5d3Ry0vbV3Am1FtGJiJvOwRsIfVChDpYStTcHTCMqtvWbV6L11BWkpzGXSW4Hv43qa+GSYOD2QU68Mb59oSk2OB+BtOLpJofmbGEGgvmwyCI9MwIDAQAB})
+Right (DID.Key MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAnzyis1ZjfNB0bBgKFMSvvkTtwlvBsaJq7S5wA+kzeVOVpVWwkWdVha4s38XM/pa/yr47av7+z3VTmvDRyAHcaT92whREFpLv9cj5lTeJSibyr/Mrm/YtjCZVWgaOYIhwrXwKLqPr/11inWsAkfIytvHWTxZYEcXLgAXFuUuaS3uF9gEiNQwzGTU1v0FqkqTBr4B8nW3HCN47XUu0t8Y0e+lf4s4OxQawWD79J9/5d3Ry0vbV3Am1FtGJiJvOwRsIfVChDpYStTcHTCMqtvWbV6L11BWkpzGXSW4Hv43qa+GSYOD2QU68Mb59oSk2OB+BtOLpJofmbGEGgvmwyCI9MwIDAQAB)
 
 -}
-data DID = DID
-  { method    :: Method
-  , publicKey :: Key.Public
-  } deriving (Show, Eq)
+data DID
+  = Key Key.Public
```
expede, line 77: RE the comment about getting this mergable:

In fact (and this is maybe controversial), we could potentially put old styles in this sum

expede: Something like:

```haskell
data DID
  = Key Key.Public
  | OldstyleRSA Oldstyle
```

matheus23: :thinking: Yeah we could do that.

However, I think getting this mergable is easier by switching DID generation to the old style, and keeping the new style around in an unused function/comment.

matheus23: I guess, what's the advantage of having both old & newstyle available, if we only ever use one of the two styles in call sites? Then we can just change it in the, uh.. definition site? (What's the opposite of call site? :D )

expede: Having both available means that _old UCANs continue to work_ — we can decode them, even if we only generate new ones

matheus23: Oh, old UCANs already continue working even in this PR. We simply allow both formats and parse them to the same DID.

expede: Ah okay. Clearly I need to look at the code clearer! I'm just in a meeting; will look in more detail in a bit. Apologies 🙏

expede: (Just got excited :P )

hs-ucan/library/Web/DID/Types.hs
```diff
@@ -157,10 +146,16 @@ parseText txt =
...
                 CryptoPassed edPK -> return $ Ed25519PublicKey edPK
 
+        (0x85 : 0x24 : rsaASNPublicKeyW8s) ->
+          case Public.decodeASN1DERRSAPublicKey $ BS.pack rsaASNPublicKeyW8s of
+            Right pk -> pure $ RSAPublicKey pk
+            Left err -> fail err
+
+        -- Backwards compatibility
         (0x00 : 0xF5 : 0x02 : rsaKeyW8s) ->
           RSAPublicKey <$> parseKeyW8s (BS64.encode $ BS.pack rsaKeyW8s)
```
matheus23, line 156: This is what I'm talking about: We just add the ability to interpret another multiformat prefix.

hs-ucan/library/Web/DID/Types.hs
```diff
@@ -107,21 +101,16 @@ instance FromHttpApiData DID where
...
       multicodecW8 :: [Word8]
       multicodecW8 =
         case pk of
-          Ed25519PublicKey ed  -> 0xed : 0x01 : BS.unpack (BA.convert ed)
-          RSAPublicKey     rsa -> 0x00 : 0xF5 : 0x02 : BS.unpack (BS64.decodeLenient . encodeUtf8 $ textDisplay rsa)
-                               {-   ^     ^     ^
-                                    |     |     |
-                                    |    "expect 373 Bytes", encoded in the mixed-endian format
-                                  "raw"
-                              -}
```
matheus23, line 124: Buuuuut, we've removed the old format generation code. If we put this back in we can merge this into main I think.

expede: Awesome; let's put it back!

We actually did this with the Oldstyle DIDs last time, too. It worked well! IIRC we "just" wrapped calls to emitted DIDs in `DID.Oldstyle`, and it serialized in the old format. After the browser was able to handle the new format, we removed that wrapper, but kept the ability to deserialize old UCANs.

Does that make sense?

matheus23: Yup! I've taken a close look at `Oldstyle` recently, actually. It's still there and we're using it to announce the fission server's DID in old-style IIRC.
I think in that discussion we realized we can just remove the ... "old oldstyle" and replace it with something else, since webnative doesn't actually "read" the DID, it only takes it verbatim as the audience for its UCANs. But that's something for another PR - I'll put that into an issue.

matheus23:
Facepalm

I got hardcore deja-vu when reading through these changes.
For a good reason: There's already this: https://github.com/fission-suite/fission/pull/573

For some reason I thought these changes didn't make it into main yet, but this confusion all stems from git branch confusion. :woozy_face: 

Please ignore this @expede!

PR #588, opened 2022-03-30
Upgrade hs-ucan to spec v0.8.2

matheus23:


matheus23:
Ok, sorry for the notification & PR spam @expede, but this is finally the "correct" diff for this PR.

hs-ucan/library/Web/UCAN/Capabilities.hs
```diff
@@ -0,0 +1,304 @@
...
+--     filterResources = \case
+--       UCAN.CapResource resource ability -> [(resource, ability)]
+--       _                                 -> []
+
+
+checkDelegation :: UCAN fct res abl -> UCAN fct res abl -> Either Error ()
+checkDelegation UCAN.UCAN{ claims = from } UCAN.UCAN{ claims = to } = do
+  () <- senderReceiverMatch
+  () <- expirationBeforeNotBefore
+  () <- notBeforeAfterExpiration
```
expede, line 276: Why `() <-` and `return ()`? Is this not implied by the type system?

hs-ucan/library/Web/UCAN/Claims/Error.hs
```diff
@@ -2,20 +2,20 @@ module Web.UCAN.Claims.Error (Error (..)) where
...
-    Expired           -> "Expired"
-    TooEarly          -> "Use too early"
-    IncorrectSender   -> "Incorrect sender"
-    IncorrectReceiver -> "Incorrect receiver"
-    ProofError resErr -> "Proof error: " <> display resErr
+    Expired             -> "Expired"
+    TooEarly            -> "Use too early"
+    IncorrectSender     -> "Incorrect sender"
+    IncorrectReceiver   -> "Incorrect receiver"
+    WitnessError resErr -> "Witness error: " <> display resErr
```
expede, line 21: 🙏

hs-ucan/library/Web/UCAN/Capabilities.hs
```diff
@@ -0,0 +1,304 @@
...
+--       any (`canDelegate` (resource, ability)) (filterResources =<< UCAN.attenuation witnessClaims)
+--   where
+--     filterOwnedResources = \case
+--       UCAN.CapOwnedResources ownedResources -> [ownedResources]
+--       _                                     -> []
+
+--     filterResources = \case
+--       UCAN.CapResource resource ability -> [(resource, ability)]
+--       _                                 -> []
+
```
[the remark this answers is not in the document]
matheus23: Yeah, this is all WIP, sorry. The idea is to write a `checkProof` function which can take `Proof`s, and verify their integrity, mostly because this is useful for testing (but possibly for something else?). Maybe I'll remove that / move that to tests.

hs-ucan/library/Web/UCAN/Capabilities.hs
matheus23: This makes it easier to include this function in something that runs in the either monad.

Or, if the question was purely syntactical, I guess I could just remove all of the `()`s. I guess it felt clearer when I wrote it, but yeah, lemme remove that :+1:

matheus23:
Remaining things to-do:
* [ ] Get tests green (part of that: get ts-ucan v0.8 fixtures into tests)
* [ ] Finish `checkProof` functions (the commented out stuff from above) & get back the equivalent of `Validation.hs` stuff. Make sure we're checking all things around UCAN validity we need to
* [ ] Make sure we're returning sensible errors for any errors we can detect
* [ ] Split up the `*.UCAN.Types` module into parts
* [ ] (Figure out whether to) rename `Proof` into `DelegationChain` and `Wittness` back into `Proof`?
* [ ] Introduce a type class for all things capability: parse with/can field, etc.

Explicitly *not* to do today is rights amplification, i.e. another function in the `DelegationSemantics` type class & another delegation type in the `Proof` data type in `Capabilities.hs`.

hs-ucan/library/Web/UCAN/Types.hs
```diff
@@ -8,8 +9,12 @@ module Web.UCAN.Types
 
   -- * reexports
 
+  , module Web.UCAN.Capabilities.Types
+  , module Web.UCAN.Claims.Types
   , module Web.UCAN.Header.Types
+  , module Web.UCAN.Nonce.Types
   , module Web.UCAN.RawContent
+  , module Web.UCAN.Proof.Types
```
matheus23, line 17: This module go quite big, so I split it into multiple smaller ones.

hs-ucan/library/Web/UCAN/Types.hs
```diff
@@ -58,30 +65,49 @@ import qualified Web.UCAN.Internal.UTF8                        as UTF8
...
 -- | An RFC 7519 extended with support for Ed25519 keys,
 --     and some specifics (claims, etc) for Fission's use case
-data UCAN fct rsc ptc = UCAN
-  { header :: Header
-  , claims :: Claims fct rsc ptc
-  , sig    :: Signature.Signature
+data UCAN fct res abl = UCAN
+  { header     :: Header
+  , claims     :: Claims fct res abl
+  , signedData :: RawContent
```
matheus23, line 81: I'm storing the `signedData` inside the UCAN.
The downside of this is that the UCAN type can represent weird stuff, e.g. the signedData might not match what's in `claims`.
However, it turns out that we've basically had this before, where lots of functions would have taken both `RawContent` and `UCAN` as an argument or return both `RawContent` as well as `UCAN`.
I just noticed the cohesion and made it more coherent.

hs-ucan/library/Web/UCAN/Validation.hs
```diff
@@ -15,164 +13,112 @@ import qualified Crypto.PubKey.RSA.PKCS15       as Crypto.RSA.PKCS
...
 
 import           Web.UCAN                       as UCAN
 import           Web.UCAN.Error                 as UCAN
-import           Web.UCAN.Proof                 as UCAN.Proof
-
+import qualified Web.UCAN.Proof.Error           as Proof
 import qualified Web.UCAN.Signature.RS256.Types as RS256
 import           Web.UCAN.Signature.Types       as Signature
 
-check ::
```
matheus23, line 39: `check` doesn't quite work with the same interface with UCANs that can have multiple capabilities, so I didn't want to revive it.
I think we have a rough toolbox of functions that cover our needs. Today the interface isn't great, I have some ideas, but I want to start using the current tools to get a better understanding of what exactly would work well.

matheus23 (commented):
Alright. I think this is ready for review finally @expede :)

All tests in fission-core/etc. will obviously fail, so overall CI will report "failed".
If you look at the logs, then we can verify that the tests in `hs-ucan` actually all succeed.

I know this is a huge diff. I'm particularly sorry for the base64 diff stuff. It's actually reverting a library change I made before :upside_down_face: 

I left some comments to hopefully help the review.

hs-ucan/library/Web/UCAN/Capabilities.hs
```diff
@@ -0,0 +1,254 @@
...
+  , IsAbility abl
+  , Resolver m
+  )
+  => UCAN fct res abl
+  -> ListT m (Either Error (DelegationChain fct res abl))
+capabilitiesFromDelegations ucan@UCAN.UCAN{ claims = UCAN.Claims{..} } = do
+  (proofRef, proofIndex) <- ListT.fromFoldable (proofs `zip` [(0 :: Natural)..])
+  liftAttempt (lift (resolveToken proofRef)) \proofResolved ->
+    attempt (parseJSONString proofResolved) \proof ->
+      attempt (Validation.checkDelegation proof ucan) \() ->
```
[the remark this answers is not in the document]
expede: We chatted about this one in person at DevConnect. TL;DR to have it for posterity: I agree that the `ListT` stuff makes this feel weird. At the root, this is about effects, so folding the streams, error handling, and CID resolution into the `m` by breaking out some effect management (read: "probably type classes") would likely make this code read better. It also frees up a lot of niceties like being able to run UCAN cache functionality directly in the concrete monad stack without having to mention it in the constraints on `m`.

expede (approved):
Just a couple comments on a cleanup option using managed effects, but otherwise 🎉 🚢

hs-ucan/library/Web/UCAN/Validation.hs
expede: Also for posterity: Philipp and I discussed a strategy for this in person, too. It's not a blocker, but could be nice to have a top-level function at some stage.

hs-ucan/library/Web/UCAN/Types.hs
expede: Yes this totally makes sense. I had tried to break it up previously, but that required abstracting over a bunch of fields... which now _need_ to be abstracted because of higher generality, so 👍👍👍

matheus23:
How do you like that @expede? :D

https://github.com/fission-suite/fission/blob/0f6a41a8d76a6c5ed147e67e7a274cee218ee1f9/hs-ucan/library/Web/UCAN/DelegationChain/Class.hs#L16-L22

Usage in the delegation chain algorithms (much less indentation!):
https://github.com/fission-suite/fission/blob/0f6a41a8d76a6c5ed147e67e7a274cee218ee1f9/hs-ucan/library/Web/UCAN/DelegationChain.hs#L141-L160


Thanks for pushing me to do that thing. Much better now I think.

matheus23:
Turns out not that much! Brooke gave me great feedback. I'm prouder of this version now:

```hs
-- | A monad abstracting over iterating elements
class Alternative m => MonadWalk m where
  walk :: Foldable f => f a -> m a


newtype StreamWithErrors m a
  = StreamWithErrors { runStreamWithErrors :: ListT m (Either Error a) }
  deriving Functor


instance Monad m => MonadWalk (StreamWithErrors m) where
  walk = StreamWithErrors . fmap Right . ListT.fromFoldable


instance Monad m => MonadError Error (StreamWithErrors m) where
  throwError = StreamWithErrors . return . Left
  catchError (StreamWithErrors listt) handler =
    StreamWithErrors $ do
      listt >>= \case
        Right a  -> return $ Right a
        Left err -> runStreamWithErrors $ handler err
```

Essentially, I've split up the `MonadDelegationChain` into `MonadWalk` and `MonadError Error`.

Much nicer now! Also, I feel like `MonadWalk` *must* exist out there already, but I can't find it!
Anyway. :man_shrugging:

expede:
_Oh yeah,_ this looks awesome 🎉 I love this pattern, where you pull everything out into effects abstractions on `m`. This code is super clean, super legible, easy to follow ✨👏👏👏✨

> Also, I feel like MonadWalk must exist out there already, but I can't find it!

I haven't actually looked at the implementation in depth, but the description for [select](https://www.stackage.org/haddock/lts-19.4/list-transformer-1.0.7/src/List.Transformer.html#select) sounds maybe related?

> Convert any collection that implements [Foldable](https://www.stackage.org/haddock/lts-19.4/base-4.15.1.0/Data-Foldable.html#t:Foldable) to another collection that implements [Alternative](https://www.stackage.org/haddock/lts-19.4/list-transformer-1.0.7/List-Transformer.html#t:Alternative)
>
> For this library, the most common specialized type for [select](https://www.stackage.org/haddock/lts-19.4/list-transformer-1.0.7/List-Transformer.html#v:select) will be:
>
> `select :: [a] -> ListT IO a`

```haskell
select :: (Foldable f, Alternative m) => f a -> m a
select = Data.Foldable.foldr cons empty
  where
    cons x xs = pure x <|> xs
```

Despite the name, this looks about right or at least right for an after midnight quick glance 😅😛), especially considering what `fromFoldable` looks like:

```haskell
fromFoldable :: (Monad m, Foldable f) => f a -> ListT m a
fromFoldable = foldr cons mzero
```

That said, I think that the custom typeclass is super expressive to read, so I wouldn't change anything

matheus23:
Ooooh right. You don't need a custom type class if select follows from Alternative 🤔

expede:
> You don't need a custom type class if select follows from Alternative 🤔

Unless you want to abstract over the behaviour completely, in which case you'd possibly want to drop some of the constraints in the class to make it more general.

That said, if using Alternative+Foldable suffices here, then 🚀

PR #605, opened 2022-06-08
Use alternate pdns-recursor image

### LINKED ISSUE #596: pdns-recursor docker image fails on m1 macs
For local server development, we have a `docker-compose.yml` file to run PowerDNS locally. However, the `pdns-recursor` image fails to properly start (when running on M1 Macs):

```
STL Exception: std::bad_alloc
```

On a quick search, it looks like this has been an issue in previous versions... perhaps upgrading will be sufficient.

walkah:
Fixes #596 

I'm not sure why the official pdns-recursor image fails to run Docker for Mac (M1's at least) but I found an alternate image that seems to work okay. Works for me ™️

### DIFF
--- a/fission-web-server/docker-compose.yml
+++ b/fission-web-server/docker-compose.yml
@@ -20,9 +20,15 @@ services:
       default:
         ipv4_address: 172.25.0.6
   dns-recursor:
-    image: powerdns/pdns-recursor-44
-    command: --forward-zones=${BASE_DOMAIN}=172.25.0.6:5300,${USER_DOMAIN}=172.25.0.6:5300,${APP_DOMAIN}=172.25.0.6:5300 --forward-zones-recurse=.=1.1.1.1 --local-address=0.0.0.0 --local-port=53
-    env_file: .env
+    image: pschiffe/pdns-recursor
+    depends_on:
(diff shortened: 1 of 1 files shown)


bgins (approved):
Tested this out and PowerDNS runs without the exception we saw in #596.

Also, `dig runfission.test` returns the expected response:

```
runfission.test.	2380	IN	SOA	a.misconfigured.dns.server.invalid. hostmaster.runfission.test. 0 10800 3600 604800 3600
```

Maybe a separate issue, but `runfission.test` doesn't resolve with `curl` or in the browser. Server is running, and I've included `127.0.0.1` in my DNS hosts.

walkah:
> Maybe a separate issue, but `runfission.test` doesn't resolve with `curl` or in the browser. Server is running, and I've included `127.0.0.1` in my DNS hosts.

Did you add an A record for `runfission.test` ? The README instructions to just create the zone doesn't actually add any records, so this is probably fine.

bgins:
> Did you add an A record for `runfission.test` ? The README instructions to just create the zone doesn't actually add any records, so this is probably fine.

I did not! That seems like the most likely explanation. 👍

PR #608, opened 2022-06-10
Append endpoint DAG surgery

icidasset:
This PR implements most of the functionality needed for the HTTP endpoint that appends a file to the DAG of an app.

**Based on the `append-endpoint` branch.**

- Adds various remote IPFS functions for working with the MFS
- Adds a function to add data to the DAG of an app, and another that calls it, and sets the new CID in the database + DNSLink
- Adds some additional info the server readme

This doesn't check for the correct resources in the provided UCAN, that'll be a separate PR.

```http
PUT "http://runfission.test:1337/v2/api/append/app-name/file.name"

Content-Type: application/octet-stream
Authorization: Bearer UCAN

204 No Content
```

### DIFF
--- a/fission-web-api/library/Fission/Web/API/Append/Types.hs
+++ b/fission-web-api/library/Fission/Web/API/Append/Types.hs
@@ -9,7 +9,7 @@ import qualified Fission.Web.API.Auth.Types              as Auth
 import           Fission.FileSystem.DirectoryName.Types  as DirectoryName
 import           Fission.FileSystem.FileName.Types       as FileName
 
-data RoutesV2 mode = RoutesV2
+newtype RoutesV2 mode = RoutesV2
   { append ::
       mode
       :- Summary     "Append a file"
--- a/fission-web-api/library/Fission/Web/API/Append/Types.hs
+++ b/fission-web-api/library/Fission/Web/API/Append/Types.hs
@@ -1,6 +1,7 @@
 module Fission.Web.API.Append.Types (RoutesV2 (..)) where
 
 import           Network.IPFS.File.Types                 as File
+import           Network.IPFS.CID.Types
 
 import           Fission.Web.API.Prelude
 
@@ -21,6 +22,6 @@ newtype RoutesV2 mode = RoutesV2
       :> ReqBody '[OctetStream] Serialized
       --
       :> Auth.HigherOrder
-      :> PutNoContent
+      :> PutAccepted '[JSON] CID
   }
   deriving Generic
--- a/fission-web-server/library/Fission/Web/Server/App/Modifier.hs
+++ b/fission-web-server/library/Fission/Web/Server/App/Modifier.hs
@@ -13,7 +13,6 @@ import qualified Network.IPFS.Files                                 as IPFS.File
 import qualified Network.IPFS.Pin                                   as IPFS.Pin
 import qualified Network.IPFS.Stat                                  as IPFS.Stat
 
-import           Network.IPFS.Local.Class (MonadLocalIPFS)
 import           Network.IPFS.Remote.Class (MonadRemoteIPFS)
 
 import qualified RIO.ByteString.Lazy                                as Lazy
@@ -26,11 +25,15 @@ import           Fission.FileSystem.DirectoryName
 import           Fission.FileSystem.FileName
 import           Fission.URL
 
+import qualified Fission.Web.Server.App.Domain.Retriever            as App.Domain
 import           Fission.Web.Server.App.Modifier.Class
 import           Fission.Web.Server.App.Retriever.Class (Retriever)
 import qualified Fission.Web.Server.App.Retriever.Class             as App.Retriever
 import           Fission.Web.Server.Config.Types
+import qualified Fission.Web.Server.Domain                          as Domain
+import qualified Fission.Web.Server.Domain.Retriever.Class          as Domain.Retriever
 import           Fission.Web.Server.Error.ActionNotAuthorized.Types
+import qualified Fission.Web.Server.IPFS.DNSLink                    as DNSLink
 import           Fission.Web.Server.Models
 import           Fission.Web.Server.MonadDB.Class (MonadDB(..))
 import           Fission.Web.Server.MonadDB.Types (Transaction)
@@ -45,7 +48,10 @@ Notes:
 
 -}
 addFile ::
-  ( MonadDB t m
+  ( App.Domain.Retriever m
+  , Domain.Retriever.Retriever m
+  , DNSLink.MonadDNSLink m
+  , MonadDB t m
   , MonadLogger m
   , MonadReader Config m
   , MonadRemoteIPFS m
@@ -62,9 +68,10 @@ addFile userId appName@(DirectoryName appNameText) fileName rawData = do
   now           <- currentTime
   domainName    <- asks baseAppDomain
 
-  mayApp        <- App.Retriever.byURL
-                    userId
-                    (URL domainName $ Just $ Subdomain appNameText)
+  let subdomain = Subdomain appNameText
+  let url = URL domainName (Just subdomain)
+
+  mayApp        <- App.Retriever.byURL userId url
 
   case mayApp of
     Left _ -> do
@@ -72,7 +79,7 @@ addFile userId appName@(DirectoryName appNameText) fileName rawData = do
 
     Right (Entity appId app@App {appCid}) ->
       if isOwnedBy userId app then do
-        -- Update DAG & DB
+        -- Update DAG, DNSLink & DB
         appendToDag appName fileName rawData appCid >>= \case
           Left err ->
             return $ Left err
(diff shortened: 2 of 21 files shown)


fission-web-server/library/Fission/Web/Server/App/Modifier.hs
```diff
@@ -8,18 +9,109 @@ import           Database.Persist                                   as Persist
...
+  , MonadLogger m
+  , MonadTime m
+  )
+  => UserId
+  -> DirectoryName
+  -> FileName
+  -> Lazy.ByteString
+  -> m (Either Errors' ())
+addFile userId (DirectoryName appName) (FileName fileName) rawData = do
+  let domainName = "fission.app"
```
icidasset, line 57: Need to get this from the environment.

fission-web-server/library/Fission/Web/Server/App/Modifier.hs
```diff
@@ -8,18 +9,109 @@ import           Database.Persist                                   as Persist
...
+Notes:
+* Overwrites existing file data
+* Uses IPFS MFS as a temporary storage to manipulate the DAG structure
+
+-}
+addFile ::
+  ( MonadRemoteIPFS m
+  , MonadLocalIPFS m
+  , MonadIO t
+  , MonadDB (Transaction t) m
```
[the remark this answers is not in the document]
expede: @icidasset I believe that should just be `MonadDB t m`, but lemme build it and double check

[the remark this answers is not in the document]
expede: Perhaps to clarify: the type class `MonadDB t m` says that there's a `t` that can be run in the environment `m`. So for example (just for intuition) `MonadDB PostgresTransaction IO` says that you'll run a PG transaction in IO.

ipfs/library/Network/IPFS/Client/Files/Copy/Types.hs
```diff
@@ -0,0 +1,12 @@
+module Network.IPFS.Client.Files.Copy.Types (API) where
+
+import           Servant.API
+import           Network.IPFS.Prelude
+
+
+type API
+  = "cp"
+    :> QueryParam' '[Required] "arg" Text -- from
+    :> QueryParam' '[Required] "arg" Text -- to
```
expede, line 10: I had to double check against the HTTP API, because this is correct, but seems _like a terrible design._ Most/many libraries will return an array of `arg`s, and you depend on them being sent in the right order 😬 We may want to open an Issue on go-ipfs suggesting they use `from` and `to`.

Thoughts?

matheus23: My hunch is they have some sort of auto-generated CLI + HTTP API from some kind of schema, and if they changed it to be parameterized as `from` and `to` in the HTTP API, then their CLI API would have to look like this: `ipfs files cp --from=<path> --to=<path>`, which would be weird for someone used to `cp` in unix.
But maybe that's what `cp` should've been like from the start? :P

fission-web-server/library/Fission/Web/Server/App/Modifier.hs
```diff
@@ -8,18 +9,135 @@ import           Database.Persist                                   as Persist
...
+  -> FileName
+  -> Lazy.ByteString
+  -> CID
+  -> m (Either Errors' CID)
+appendToDag domainName (DirectoryName appName) (FileName fileName) rawData appCid = do
+  let appCidText = unaddress appCid
+
+  let tmpDirPath = Text.concat [ "/", domainName, "/", appName, "/" ]
+  let distDirPath = Text.concat [ tmpDirPath, appCidText, "/" ]
+  let filePath = Text.concat [ distDirPath, fileName ]
```
bgins, line 125: We may want to write the files to an `uploads` subdirectory in the app. The original rationale is to have a dedicated location for files that are appended, but we should check if that's still what we want.

icidasset: Fixed 👍 I haven't allowed nested directories, but that should be as simple as allowing that to be passed to Servant (ie. now it doesn't allow multiple segments, could change that). I think we should probably allow that?

expede: @icidasset

No need to change, but FYI you can also do this with `<>`

```haskell
let filePath = distDirPath <> fileName
```

ipfs/library/Network/IPFS/Client/Files/Copy/Types.hs
expede: Ah, indeed this is where they do that:

https://github.com/ipfs/kubo/blob/d82dec32482bb4ece33d3e867c4f615eea6bb487/core/corehttp/commands.go#L126-L152

Which is probably less work to maintain when you're doing all of this manually without much help from the compiler

fission-web-api/library/Fission/Web/API/Append/Types.hs
```diff
@@ -9,7 +9,7 @@ import qualified Fission.Web.API.Auth.Types              as Auth
 import           Fission.FileSystem.DirectoryName.Types  as DirectoryName
 import           Fission.FileSystem.FileName.Types       as FileName
 
-data RoutesV2 mode = RoutesV2
+newtype RoutesV2 mode = RoutesV2
```
expede, line 12: Good call :)

ipfs/library/Network/IPFS/Client/Files/Write/Form/Types.hs
```diff
@@ -0,0 +1,23 @@
...
+
+import           RIO
+import qualified RIO.ByteString.Lazy     as Lazy
+
+import           Servant.Multipart
+import           Servant.Multipart.API
+
+import qualified Network.IPFS.File.Types as File
+
+data Form = Form
```
expede, line 11: Since this only has one field, this could also be a `newtype`

expede: (Which is really just a performance optimization. The compiler can't automatically force this because of how it interacts with laziness — newtypes are always strict)

ipfs/library/Network/IPFS/Files.hs
```diff
@@ -0,0 +1,155 @@
...
+        , parents = True
+        }
+  in
+  IPFS.mfsCopy args >>= \case
+    Right () ->
+      return $ Right ()
+
+    Left err -> do
+      formattedError <- parseClientError err
+      return <| Left formattedError
```
expede, line 55: No need to change, but just an FYI we're moving away from the `<|` operator since it's equivalent to `$` (which is used just above on line 51)

ipfs/library/Network/IPFS/Files.hs
```diff
@@ -0,0 +1,155 @@
...
+
+-}
+statCID :: (MonadLogger m, MonadRemoteIPFS m)
+  => FilePath
+  -> m (Either IPFS.Files.Error IPFS.CID)
+statCID path =
+  let
+    args = MfsStatArgs { path = Text.pack path }
+  in
+  IPFS.mfsStat args >>= \case
```
expede, line 91: This is syntactically correct, but if there's ever a `do` added above, I'm pretty sure that you'll be forced to indent this.

fission-web-server/library/Fission/Web/Server/App/Modifier.hs
icidasset: Changed 👍

ipfs/library/Network/IPFS/Client/Files.hs
```diff
@@ -0,0 +1,14 @@
+module Network.IPFS.Client.Files
+  ( API
+  , module Statistics
```
bgins, line 3: Seeing a warning that `module Statistics` exports nothing. It may be because `Network.IPFS.Client.Files.Statistics.Types` is imported qualified. In other places where we re-export, the import isn't qualified.

bgins (approved):
Looks good and working well over here! Great to have all the MFS APIs. ✨ 🎉

fission-web-server/library/Fission/Web/Server/App/Modifier.hs
```diff
@@ -8,18 +9,135 @@ import           Database.Persist                                   as Persist
...
+  , MonadRemoteIPFS m
+  , MonadTime m
+  , Modifier t
+  , Retriever m
+  )
+  => UserId
+  -> DirectoryName
+  -> FileName
+  -> Lazy.ByteString
+  -> m (Either Errors' ())
```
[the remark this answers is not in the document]
icidasset: I'll return the CID yeah, should be useful 👍 

The app CID could be helpful too, since DNS lags sometimes.
Depends on how this'll be used 🤔

ipfs/library/Network/IPFS/Client/Files.hs
icidasset: I'll remove it, don't think I need this anymore.

PR #610, opened 2022-06-16
Bugfix: `fromRawContent` can never succeed

expede:
@bgins and I ran into this problem while working on #606. The core thing is that it's impossible to actually construct a UCAN from a `RawContent` without also having a `Signature`, so `fromRawContent` could never succeed.

```haskell
UCAN.RawContent -> Either Resolver.Error (UCAN fct rsc ptc) -- 😾 How get signature?
```

I also bubbled some error content through while I had the hood up.

# TODO

The one thing that isn't covered here is if `UCAN.parse` should automatically add quotes to make the JSON parsing work internally. I think not, but perhaps it would be less surprising with them?

### DIFF
--- a/hs-ucan/library/Web/UCAN.hs
+++ b/hs-ucan/library/Web/UCAN.hs
@@ -1,25 +1,24 @@
 module Web.UCAN
-  ( fromRawContent
+  ( parse
   , module Web.UCAN.Types
   ) where
 
 import           Data.Aeson
+
 import           RIO
+import qualified RIO.Text as Text
 
 import           Web.UCAN.Resolver.Error as Resolver
 import           Web.UCAN.Types
 
-
-fromRawContent ::
+parse ::
   ( FromJSON fct
   , FromJSON rsc
   , FromJSON ptc
   )
-  => RawContent
+  => ByteString
   -> Either Resolver.Error (UCAN fct rsc ptc)
-fromRawContent rawContent =
-  case decodeStrict rawContentBS of
-    Nothing   -> Left (InvalidJWT rawContentBS)
-    Just ucan -> Right ucan
-  where
-    rawContentBS = encodeUtf8 (unRawContent rawContent)
+parse bs =
+  case eitherDecodeStrict bs of
+    Left reason -> Left $ InvalidJWT (Text.pack reason) bs
+    Right ucan  -> Right ucan
--- a/hs-ucan/library/Web/UCAN/Error.hs
+++ b/hs-ucan/library/Web/UCAN/Error.hs
@@ -8,7 +8,7 @@ import qualified Web.UCAN.Header.Error    as Header
 import qualified Web.UCAN.Signature.Error as Signature
 
 data Error
-  = ParseError
+  = ParseError     Text
   | HeaderError    Header.Error
   | ClaimsError    Claims.Error
   | SignatureError Signature.Error
@@ -22,7 +22,7 @@ instance ToJSON Error where
 
 instance Display Error where
   display = \case
-    ParseError         -> "Could not parse JWT"
+    ParseError     err -> "Could not parse JWT: " <> display err
     HeaderError    err -> "JWT header error: "    <> display err
     SignatureError err -> "JWT signature error: " <> display err
     ClaimsError    err -> "JWT claims error: "    <> display err
--- a/hs-ucan/library/Web/UCAN/Resolver/Class.hs
+++ b/hs-ucan/library/Web/UCAN/Resolver/Class.hs
@@ -4,8 +4,7 @@ import           Network.IPFS.CID.Types
 
 import           RIO
 
-import qualified Web.UCAN.RawContent     as UCAN
 import           Web.UCAN.Resolver.Error as Resolver
 
 class Monad m => Resolver m where
-  resolve :: CID -> m (Either Resolver.Error UCAN.RawContent)
+  resolve :: CID -> m (Either Resolver.Error ByteString)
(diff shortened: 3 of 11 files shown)


hs-ucan/library/Web/UCAN/Resolver/Error.hs
```diff
@@ -7,13 +7,13 @@ import           Web.UCAN.Internal.Orphanage.ClientError ()
 
 data Error
   = CannotResolve CID ClientError
-  | InvalidJWT ByteString
+  | InvalidJWT Text ByteString
```
expede, line 10: Bubbling the reason up

hs-ucan/library/Web/UCAN/Resolver/Class.hs
```diff
@@ -4,8 +4,7 @@ import           Network.IPFS.CID.Types
 
 import           RIO
 
-import qualified Web.UCAN.RawContent     as UCAN
 import           Web.UCAN.Resolver.Error as Resolver
 
 class Monad m => Resolver m where
-  resolve :: CID -> m (Either Resolver.Error UCAN.RawContent)
+  resolve :: CID -> m (Either Resolver.Error ByteString)
```
expede, line 10: There's a bunch of places that need the signature. We haven't resolved from CID before, so had never actually caught this. I'm 90% sure this is on me. Apologies 🙏

hs-ucan/library/Web/UCAN/Error.hs
```diff
@@ -8,7 +8,7 @@ import qualified Web.UCAN.Header.Error    as Header
 import qualified Web.UCAN.Signature.Error as Signature
 
 data Error
-  = ParseError
+  = ParseError     Text
```
expede, line 11: Context!

hs-ucan/library/Web/UCAN.hs
```diff
@@ -1,25 +1,24 @@
...
-fromRawContent rawContent =
-  case decodeStrict rawContentBS of
-    Nothing   -> Left (InvalidJWT rawContentBS)
-    Just ucan -> Right ucan
-  where
-    rawContentBS = encodeUtf8 (unRawContent rawContent)
+parse bs =
+  case eitherDecodeStrict bs of
+    Left reason -> Left $ InvalidJWT (Text.pack reason) bs
+    Right ucan  -> Right ucan
```
expede, line 24: This is the big change. The awkward thing about this function is that it's now basically `Aeson.decode`. We do embed the value in `InvalidJWT`, so I think it probably does warrant its own function, but we can also drop this function and do that at the call sites of preferred.

hs-ucan/library/Web/UCAN/Resolver/Class.hs
expede: I considered (and actually started on) making...

```haskell
newtype UnparsedUCAN = UnparsedUCAN Text
```

...just to tack the intention around in the code. Its' unparsed though, so `ByteString` kind of implies that we don't know what's in there already 🤷‍♀️ 

This could probably be replaced with good ol' IPFS.byCID (or whatever it's called), but the server and CLI call it differently so I kept it.

hs-ucan/library/Web/UCAN/Validation.hs
```diff
@@ -111,16 +111,21 @@ checkProof now ucan@UCAN {claims = Claims {proof}} =
...
+          return . Left . UCAN.ClaimsError . ProofError $ UCAN.Proof.ResolverError err
 
-        Right rawProof -> do
-          case UCAN.fromRawContent rawProof of
-            Left _          -> return $ Left UCAN.ParseError
+        Right proofBS -> do
+          case UCAN.parse proofBS of
             Right proofUCAN ->
-              check' rawProof proofUCAN now <&> \case
+              check' (UCAN.contentOf (decodeUtf8Lenient proofBS)) proofUCAN now <&> \case
```
expede, line 119: Turns out that the content conversion was only required in a handful of places

bgins (approved):
Confirmed this works over on the `cli-app-delegate-command` branch:

https://github.com/fission-suite/fission/blob/dcf5f5885425389d16f631c853cc65853cb3944f/fission-cli/library/Fission/CLI/Handler/App/Delegate.hs#L160-L173

For a valid token, that results with a `UCAN`.

Thanks for the fix! Much appreciated. 🙏

hs-ucan/library/Web/UCAN.hs
matheus23: I need to write up some thoughts I have on using type classes for en/decode functions and why I dislike them :D

I love having specific functions like these :)

expede: Yeah, I like having both, because of canonical. That said, being able to abstract over things with To/FromJSON instances is hugely useful in practice.

I'd love to know why you avoid them!

hs-ucan/library/Web/UCAN/Resolver/Class.hs
matheus23: I like this version.

The version before should be blamed on me! The type used to be `CID -> m (Either Resolver.Error (UCAN ..., RawContent))`. However that means you'd have to parse the UCAN at that point already, and that turned out difficult for reasons I don't quite remember anymore.

I misunderstood `RawContent` back then and assumed it was just a newtype wrapper for an "unparsed UCAN" like what the `ByteString` is now. But yeah, this is correct and I like it better :)

In fact, that's what I've changed it to in the `ucan-upgrade` branch:
https://github.com/fission-suite/fission/blob/c630610f53055a1d347b3672ffbbde3aaee6896f/hs-ucan/library/Web/UCAN/Resolver/Class.hs#L9-L10

Again, sorry for messing this up & not backporting it!

matheus23 (approved):
Yes!
I have this fixed in `ucan-upgrade`, and it'll lead to some conflicts there, but that's fine, they'll be easy to resolve.

On me! Should've backported my fixes.

bgins:
@expede is this ready to be merged?

PR #615, opened 2022-06-24
haskell.nix integration

walkah:
Another **VERY WIP EXPERIMENT** using [haskell.nix](https://github.com/input-output-hk/haskell.nix) to build proper nix packages for fission cli + server. 

Known issues:
- [ ] Builds a *lot* of stuff (there is *no* binary cache for `aarch64-darwin` at all, and minimal stuff on x86_64 platforms)
- [x] current stack-based build / install workflow doesn't work (*only* `nix build`)


tennox:
I forgot to check first if an existing PR is doing what I was doing, so I duplicated this: #623

But then, checking your checklist of known issues, I think I managed to not break the existing workflow - `stack build` still works fine in the devshell.
Maybe you check it out yourself? As I never actually used a stack-based haskell workflow and I'm not too sure about it.

walkah:
> I also managed to use the binary cache (`linux x86_64`), at least for the basics (like `ghc`) - the haskell packages themselves still took quite some time to build. But this is the same with `stack build` (for me)?

totally right! I was tinkering with two approaches in parallel ... I've rebased against the dev setup in `main` and it looks like it's all working pretty well. thanks for poking around! 🙏

PR #621, opened 2022-07-14
Validate UCAN resource

icidasset:
Checks the resource of the UCAN used with the append endpoint, and fixes the domain name used in the "delegated" UCAN resources.

## Test plan (required)

Using a UCAN with an invalid resource, ie. a `rsc` that doesn't match `{ "app": "DOMAIN_OF_APP" }`, should produce a `403` status.

### DIFF
--- a/fission-web-server/library/Fission/Web/Server/Handler/Append.hs
+++ b/fission-web-server/library/Fission/Web/Server/Handler/Append.hs
@@ -4,7 +4,6 @@ import           Fission.Web.Server.IPFS.DNSLink.Class
 import           Network.IPFS.File.Types                            as File
 import           Network.IPFS.Remote.Class (MonadRemoteIPFS)
 
-import           Servant
 import           Servant.Server.Generic
 
 import           Fission.Prelude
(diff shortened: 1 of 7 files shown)


fission-web-server/library/Fission/Web/Server/Handler/Append.hs
```diff
@@ -36,6 +35,6 @@ handlerV2 ::
   => Append.RoutesV2 (AsServerT m)
 handlerV2 = Append.RoutesV2 {append}
   where
-    append appName fileName (Serialized rawData) Authorization {about = Entity userId _} = do
-      cid <- Web.Err.ensureM $ App.addFile userId appName fileName rawData
+    append appName fileName (Serialized rawData) Authorization {about = Entity userId _, potency, resource} = do
```
bgins, line 38: Does this `userId` represent the root issuer in the UCAN? From testing, it seems like it does, but it's not clear to me where we are pulling the UCAN apart to get that.

bgins (approved):
Looks good! Checked a delegated case that should succeed and a couple that should not due to invalid app resources or expired UCANs.

fission-web-server/library/Fission/Web/Server/Handler/Append.hs
icidasset: It does. That happens here: https://github.com/fission-suite/fission/blob/fa3d2b52848fe3573714a9a6c3611555461c13bc/fission-web-server/library/Fission/Web/Server/Auth/Token/UCAN.hs#L55

bgins: Ah great! Thanks for the reference.

PR #622, opened 2022-07-15
remove nix pure from shell environment

walkah:
Using stack's nix integration with `--nix-pure` seems to be (sometimes?) creating segfaults in stack specifically on aarch64-darwin. This PR simply removes `--nix-pure` from the stack wrapper.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -27,7 +27,6 @@
               wrapProgram $out/bin/stack \
                 --add-flags "\
                   --nix \
-                  --nix-pure \
                   --nix-shell-file=nix/stack-integration.nix \
                 "
             '';


bgins:
Tried this out and `build`, `cli-install`, and `server-install` all work for me, but `quality` segfaults:

```
[nix-shell:~/code/fission/fission]$ quality
⚙️  Running quality...
/private/tmp/nix-shell-29850-0/rc: line 3: 30030 Segmentation fault: 11  '/nix/store/a6qyw1maymr5589nmw6lzqpanm2xr1ff-stack-2.7.5/bin/stack' $STACK_IN_NIX_EXTRA_ARGS '--internal-re-exec-version=2.7.5' '--nix' '--nix-shell-file=nix/stack-integration.nix' 'test' '--test-arguments=--color=always'
```

Hadn't noticed a segfault from `quality` before, but the `cli-install` and `server-install` ones have also been intermittent.

walkah:
> Hadn't noticed a segfault from `quality` before, but the `cli-install` and `server-install` ones have also been intermittent.

just to be super clear - you used a new / fresh shell (clean ENV) before trying your tests? `quality` works for me with this change.

icidasset:
`quality` doesn't work for me either. Interestingly if I run `stack build --nix --nix-shell-file=nix/stack-integration.nix`, I get the segfault too. Whereas running `build`, which is the same command, runs fine. How weird is that?

icidasset:
I do notice that the quality command has some additional single quotes in its command, whereas the others do not, might have something to do with it (because of the wrapper).

bgins:
> just to be super clear - you used a new / fresh shell (clean ENV) before trying your tests? quality works for me with this change.

Yep, this was in a clean environment.

walkah:
> Yep, this was in a clean environment.

sorry for the "did you plug it in" question 🙈  I am a little stumped why this is happening *sometimes*.

> I do notice that the quality command has some additional single quotes in its command, whereas the others do not, might have something to do with it (because of the wrapper).

yea, this is the stack wrapper script in entirety (from my machine):
```
#! /nix/store/lkw407y1x1v5bg6hc290c5ry1qaabbgl-bash-5.1-p16/bin/bash -e
exec -a "$0" "/nix/store/1aw7295bi6kxlvqb9a5kp9276aym4970-stack/bin/.stack-wrapped"      --nix     --nix-shell-file=nix/stack-integration.nix    "$@"
```
I don't think the single quotes should cause an issue here.

icidasset (approved):
Only `quality` doesn't work for me, fix later? 🤷 I can run the quality command by doing `.stack-wrapped test --nix --nix-shell-file=nix/stack-integration.nix --test-arguments='--color=always'` in `nix develop`.

Also, I would've thought `.stack-wrapped` would run stack with the additional flags defined the flake.nix file. But I guess it doesn't? Because if I leave the `--nix` flag from the command above, it starts to download GHC.

PR #625, opened 2022-07-25
Add secp256k1 key support for did:key

icidasset:
This PR adds secp256k1 key support for `did:key`, the type of key used by Ethereum, Bitcoin and various other blockchains. We intend to use this initially for wallet auth, where we look up the public key of an Ethereum account and then construct a DID from that client side. In addition to that, we will also sign UCANs with those DIDs, which this PR also adds support for.

## Public key types

This PR changes how we store public keys in the database. Before we stored the public key bytes formatted as base64. This is now wrapped in a JSON object, along with the type of the public key. That affects the `Update Public Key` and `Update Public Key via Email Challenge` API endpoints for which I created new V3 routes.

_New format:_
```json
{"key":"Hv+AVRD2WUjUFOsSNbsmrp9fokuwrUnjBcr92f0kxw4=","type":"Ed25519"}
{"key":"AvuqnqcYF49vlzMHKKAuaXjEOaqs/x9V1GnY+rJxx6Y2","type":"Secp256k1"}
```

### Migration

I've written a DB migration that converts the old string format into the json-object format.
This can be found at [fission-web-server/sql/1658241548_set_public_key_type.sql](https://github.com/fission-codes/fission/pull/625/files#diff-4d417e89171a0c33b4ab99a584e4ee2e5ebdc2d6cbce42e431adb61407a4038d)
You can run this migration multiple times, it detects both the old and new format.

## SECP256K1 UCANs

UCANs with secp256k1 signatures can be created by using the `ES256K` algorithm (`alg` field in the header). This `alg` value is specified here: https://datatracker.ietf.org/doc/html/rfc8812#section-3.2 Not sure that's the correct spec, but it's the only one I found.

Example: https://github.com/fission-codes/webnative-walletauth/blob/375c63881a300442ec3112bf6a7863fc22ecbb4a/src/webnative.ts#L155-L159

### Ethereum signatures

This is currently specifically tuned to Ethereum. Normally the content that one would sign for a UCAN signature would be `${UCAN_HEADER}.${UCAN_PAYLOAD}`. But here this content will be:
```js
const regularContentBytes =
  uint8arrays.fromString("${UCAN_HEADER}.${UCAN_PAYLOAD}", "utf8")

const actualContent = keccak_256(
  uint8arrays.concat([
    uint8arrays.fromString("\x19Ethereum Signed Message:\n${regularContentBytes.length}", "utf8"),
    regularContentBytes
  ]) 
)
```

I made a module that is responsible for "wrapping" this data at `Web.UCAN.Signature.Secp256k1.Ethereum`.

~~I guess to see if it's an Ethereum signatures we could detect if the recovery value `v` of the secp256k1 signature is bigger than or equal to 27 (specification: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md#specification). Although I guess Bitcoin has a `v` value of 27 as well. Thoughts?~~
Added an Ethereum check which looks for a `v` value of `>= 35`.


hs-ucan/library/Crypto/Key/Asymmetric/Public/Types.hs
```diff
@@ -1,70 +1,83 @@
...
 
 instance FromHttpApiData Public where
-  parseUrlPiece txt =
-    if "MII" `Text.isPrefixOf` txt
-      then RSAPublicKey     <$> parseUrlPiece txt
-      else Ed25519PublicKey <$> parseUrlPiece txt
+  parseUrlPiece text =
+    case (Json.decode . Lazy.fromStrict . Text.encodeUtf8 $ text) :: Maybe Public of
+      Just pub -> Right pub
+      Nothing -> Left "Failed to decode the JSON formatted public key"
```
icidasset, line 56: As far as I've seen we only use this in the fixtures.
Not sure if this has any other implications.

hs-ucan/library/Crypto/Key/Asymmetric/Public/Types.hs
```diff
@@ -1,70 +1,83 @@
...
-  parseJSON = withText "PublicKey" \txt ->
-    case parseUrlPiece txt of
-      Right pk -> return pk
-      Left msg -> fail $ Text.unpack msg
+  parseJSON =
+    withObject "PublicKey" \j ->
+      ((,) <$> j .: "type" <*> j .: "key") >>= \(t, k) ->
+      case fromConstructor t k of
+        Right pub -> return pub
+        Left err -> Json.parseFail (Text.unpack err)
```
icidasset, line 67: Same question here as with `parseUrlPiece`, any other implications?

hs-ucan/library/Crypto/Key/Asymmetric/Public/Types.hs
```diff
@@ -79,3 +94,19 @@ instance ToSchema Public where
...
+
+
+-- CONSTRUCTORS
+
+
+fromKeyType :: Text -> Text -> Either Text Public
+fromKeyType "Ed25519"   = fmap Ed25519PublicKey . parseUrlPiece
+fromKeyType "RSA"       = fmap RSAPublicKey . parseUrlPiece
+fromKeyType "Secp256k1" = fmap Secp256k1PublicKey . parseUrlPiece
+fromKeyType _           = \_ -> Left "Invalid Public.Key type"
```
icidasset, line 107: I tried to do this dynamically with the `Data.Data` module, but turned into a bit of hassle, so I kept it simple.

hs-ucan/library/Web/UCAN/Signature/Types.hs
```diff
@@ -9,26 +9,32 @@ import           RIO                                           hiding (length)
...
 import           Web.UCAN.Internal.Orphanage.Ed25519.Signature ()
 import qualified Web.UCAN.Signature.RS256.Types                as RS256
+import qualified Web.UCAN.Signature.Secp256k1.Types            as Secp256k1
 
 data Signature
-  = Ed25519 Ed25519.Signature
-  | RS256   RS256.Signature
+  = Ed25519   Ed25519.Signature
+  | RS256     RS256.Signature
+  | Secp256k1 Secp256k1.Signature
```
icidasset, line 17: Still need to write tests for the UCAN changes.

hs-ucan/library/Web/UCAN/Types.hs
```diff
@@ -314,3 +327,15 @@ signRS256 header claims sk =
...
+  , ToJSON rsc
+  , ToJSON ptc
+  )
+  => Header
+  -> Claims fct rsc ptc
+  -> Secp256k1.SecKey
+  -> Maybe Signature.Signature
+signSecp256k1 header claims sk =
+  -- Only works if the signed data is 32 bytes
+  Signature.Secp256k1 . Secp256k1.Signature . Secp256k1.exportRecoverableSignature <$> Secp256k1.ecdsaSignRecoverable sk (encodeUtf8 $ B64.URL.encodeJWT header claims)
```
icidasset, line 341: Need to take a closer look at this. The way we sign things with Metamask/Ethereum is not the same as this. This doesn't have the prefix and keccak256 hashing.

hs-ucan/library/Web/UCAN/Types.hs
```diff
@@ -80,28 +83,38 @@ instance
...
 
-      sig' = case sk of
-        Left rsaSK -> Unsafe.unsafePerformIO $ signRS256 header claims rsaSK
-        Right edSK -> Right $ signEd25519 header claims edSK
+        return (claims, signSecp256k1 header claims sk)
 
     case sig' of
-      Left _    -> error "Unable to sign UCAN"
-      Right sig -> return UCAN {..}
+      Nothing  -> error "Unable to sign UCAN"
```
icidasset, line 116: This is failing:
```
fission-web-server>           🔀 serialized is isomorphic to ADT:               FAIL (0.02s)
fission-web-server>             uncaught exception: ErrorCall
fission-web-server>             Unable to sign UCAN
fission-web-server>             CallStack (from HasCallStack):
fission-web-server>               error, called at library/Web/UCAN/Types.hs:116:19 in hs-ucan-0.0.2.0-IOFzaPf71q4EDXHnYnFimA:Web.UCAN.Types
```

icidasset:
@walkah Do I have to do anything special here for Nix and the Github actions? I added a C library https://github.com/bitcoin-core/secp256k1 using Nix (`pkgs.secp256k1`). Locally I had to add `pkgs.pkg-config` to get it to work, but I can't get it to work for Github actions. Any thoughts?

Resources:
https://nixos.wiki/wiki/FAQ/I_installed_a_library_but_my_compiler_is_not_finding_it._Why%3F

icidasset:
The error I'm talking about:
https://github.com/fission-codes/fission/runs/7582732721?check_suite_focus=true#step:5:12643

hs-ucan/library/Web/UCAN/Types.hs
matheus23: It does test them automatically. At least it generates UCANs with secp256k1 keys:
https://github.com/fission-codes/fission/blob/221d1f3189a90b8a60c9d1d42e35d62c312bb900/hs-ucan/library/Web/UCAN/Types.hs#L108-L114

matheus23: Having an `Arbitrary Algorithm` that adds `Secp256k1` as one of the options leads to `Arbitrary UCAN.Header` to sometimes generate with secp256k1 as the algorithm which in turn causes the code I linked above to generate secp256k1 signatures.

icidasset: Ah awesome 👏 I guess that plus the DID tests covers everything.

hs-ucan/library/Crypto/Key/Asymmetric/Public/Types.hs
matheus23: Hm. Is it possible that the route `/v2/api/user/did` uses this? The servant request type there is
```hs
type SetAuthenticated
  =  Summary "Update Public Key"
  :> Description "Set currently authenticated user's root public key to another one"
  --
  :> ReqBody '[JSON] Key.Public
  --
  :> Auth.HigherOrder
  :> PutNoContent
```
And I'm thinking the `ReqBody '[JSON] Key.Public` probably uses this instance for going JSON -> Public Key.

matheus23: Same goes for the `SetViaChallenge` route. That one is actually used in the recovery process in the dashboard (I'm not aware of any production usage of `SetAuthenticated`).

icidasset: Thanks! I think it makes sense here to use the new structure `{ type: ..., key: ... }`, for both of these. But I guess I could use the old logic as a fallback, in case just a string was passed? What do you think?

matheus23: I guess that'd work.
Here's two more options:
- Full blown: Create a `v3` route that has the new behavior, keep the old route with the old behavior by creating a `Key.Public.Oldstyle` type and implement the old `FromJSON` for it.
- No backwards compatibility at all: We figure out all places that use this function. If it turns out that's pretty much only the dashboard app's recovery process, then we just deploy a fix roughly at the same time as we're updating our server.

hs-ucan/package.yaml
```diff
@@ -158,6 +159,8 @@ dependencies:
 
 library:
   source-dirs: library
+  pkg-config-dependencies:
+    - libsecp256k1
```
icidasset, line 163: I added this line to get it to build on the Github actions, but doesn't seem to change anything. The `libsecp256k1` package has this too.

flake.nix
```diff
@@ -57,8 +56,16 @@
...
+              pkgs.secp256k1
               pkgs.stylish-haskell
               tasks
+
+              # https://nixos.wiki/wiki/FAQ/I_installed_a_library_but_my_compiler_is_not_finding_it._Why%3F
+              # Needed for binding secp256k1
+              pkgs.autoconf
+              pkgs.automake
+              pkgs.pkg-config
+              pkgs.libtool
```
walkah, line 68: I'm not sure we need these ... maybe try the above change and see if they're still needed?

hs-ucan/library/Crypto/Key/Asymmetric/Public/Types.hs
```diff
@@ -61,12 +61,25 @@ instance IsString (Either Text Public) where
...
-    withObject "PublicKey" \j ->
-      ((,) <$> j .: "type" <*> j .: "key") >>= \(t, k) ->
-      case fromKeyType t k of
-        Right pub -> return pub
-        Left err -> Json.parseFail (Text.unpack err)
+  parseJSON (Object j) =
+    ((,) <$> j .: "type" <*> j .: "key") >>= \(t, k) ->
+    case fromKeyType t k of
+      Right pub -> return pub
+      Left err -> Json.parseFail (Text.unpack err)
```
matheus23, line 68: Just saw this flying by and was thinking how about this?
```suggestion
  parseJSON (Object j) =
    (fromKeyType <$> j .: "type" <*> j .: "key") >>= \case
      Right pub -> return pub
      Left err -> Json.parseFail (Text.unpack err)
```

Or perhaps this is more readable:
```suggestion
  parseJSON (Object j) = do
    t <- j .: "type"
    k <- j .: "key"
    case fromKeyType t k of
      Right pub -> return pub
      Left err -> Json.parseFail (Text.unpack err)
```

icidasset: Thanks! Definitely more readable 😄

flake.nix
```diff
@@ -57,8 +56,16 @@
               haskellPackages.haskell-language-server
               pkgs.cachix
               pkgs.nixpkgs-fmt
+              pkgs.secp256k1
```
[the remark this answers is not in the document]
icidasset: Thanks! That indeed fixed the ubuntu nix build 🎉 I haven't gotten the native ubuntu action to build yet with the apt-get install, but we did get a step further.

hs-ucan/library/Crypto/Key/Asymmetric/Public/Types.hs
icidasset: I've implemented the V3 routes 👍

hs-ucan/library/Crypto/Key/Asymmetric/Public/Types.hs
icidasset: Created new v3 routes to handle this.

hs-ucan/library/Web/UCAN/Validation.hs
```diff
@@ -181,3 +187,37 @@ checkEd25519Signature (UCAN.RawContent raw) ucan@UCAN {..} =
...
+checkSecp256k1Signature ::
+     UCAN.RawContent
+  -> UCAN fct rsc ptc
+  -> Secp256k1.Signature
+  -> Either UCAN.Error (UCAN fct rsc ptc)
+checkSecp256k1Signature (UCAN.RawContent raw) ucan@UCAN {..} (Secp256k1.Signature innerSig) = do
+  case publicKey of
+    Secp256k1PublicKey pk ->
+      -- `innerSig` is actually a recoverable signature,
+      -- but since Ethereum (or Metamask?) often has a `v` value
```
avivash, line 200:
```suggestion
      -- but since Ethereum often has a `v` value
```

No need to mention Metamask here. Any Ethereum wallet(I would hope) would function in a consistent way 👍🏼  https://eips.ethereum.org/EIPS/eip-155

hs-ucan/library/Web/UCAN/Internal/Orphanage/Secp256k1/PublicKey.hs
```diff
@@ -0,0 +1,48 @@
...
+
+import           Data.Aeson
+import           RIO
+import           Test.QuickCheck
+
+instance Arbitrary Secp256k1.PubKeyXY where
+  arbitrary = Secp256k1.derivePubKey <$> arbitrary
+
+instance ByteArrayAccess Secp256k1.PubKeyXY where
+  -- Bool argument to `exportPubKeyXY` is wether or not to enable compression
```
avivash, line 24:
```suggestion
  -- Bool argument to `exportPubKeyXY` is whether or not to enable compression
```

small typo: `wether` -> `whether` 👀

hs-ucan/library/Web/UCAN/Validation.hs
icidasset: Thanks, forgot to remove this! Regarding this, see the last paragraph in the PR description, would love your feedback on that 🙏

avivash: > Thanks, forgot to remove this! Regarding this, see the last paragraph in the PR description, would love your feedback on that 🙏

@icidasset sure! I think [EIP-155](https://eips.ethereum.org/EIPS/eip-155) may be the key to distinguishing the two. as you mentioned, bitcoin and ethereum used to both have `v` values of 27(or sometimes 28?), but, looking at EIP-155, it says `v = CHAIN_ID * 2 + 35` or `v = CHAIN_ID * 2 + 36`, so maybe we can just check if `v >= 35`?

icidasset: Ah good point, I'll use `v >= 35`, thanks! Do you know of any other chains that prefix and/or hash the content before signing it?

icidasset:
Still can't figure out why data-root lookups are returning a `404` for me with my local instance. If anyone else is trying it out. You can test it with [walletauth](https://github.com/fission-codes/webnative-walletauth), by putting the following code in the `index.ts` file (replacing the existing `async` function):
```ts
;(async () => {
  wn.setup.endpoints({
    api: "http://runfission.test:1337",
    lobby: "http://localhost:8001",
    user: "fissionuser.test"
  })

  webnative.login()
})()
```

hs-ucan/library/Web/UCAN/Validation.hs
avivash: > Ah good point, I'll use `v >= 35`, thanks! Do you know of any other chains that prefix and/or hash the content before signing it?

hmm 🤔 I believe Tezos and Cosmos prefix them, but I'm not entirely sure how it works across most other chains(I've mostly worked in ethereum). I can do some digging though!

icidasset: That'd be great! Especially if we want to support other chains than Ethereum in the future 👀

avivash:
> Still can't figure out why data-root lookups are returning a `404` for me with my local instance. If anyone else is trying it out. You can test it with [walletauth](https://github.com/fission-codes/webnative-walletauth), by putting the following code in the `index.ts` file (replacing the existing `async` function):
> 
> ```ts
> ;(async () => {
>   wn.setup.endpoints({
>     api: "http://runfission.test:1337",
>     lobby: "http://localhost:8001",
>     user: "fissionuser.test"
>   })
> 
>   webnative.login()
> })()
> ```

I don't have a huge amount of context around this, but I just installed `nix` (i know i'm late to the `nix` party 😅), got  `stack` setup and i successfully ran `stack build` on your branch, but nothing seems to be running on port `8001`. Is there a specific package I need to run? This is my first time using haskell 🙈

bgins:
> I don't have a huge amount of context around this, but I just installed `nix` (i know i'm late to the `nix` party 😅), got `stack` setup and i successfully ran `stack build` on your branch, but nothing seems to be running on port `8001`. Is there a specific package I need to run? This is my first time using haskell 🙈

Hey @avivash! I can give you some pointers on project setup and running the server.

First, a question -- are you using the nix shell in the project? You can use it by running `nix-shell` at the root of the project. That will take a while to run because it will retrieve and compile a whole lot of Haskell.

Once that finishes and you are in the nix shell, you should be able to run `helpme` to get a help menu that looks like this:

```sh
$ helpme
  ____                                          _     
 / ___|___  _ __ ___  _ __ ___   __ _ _ __   __| |___ 
| |   / _ \| '_ ` _ \| '_ ` _ \ / _` | '_ \ / _` / __|
| |__| (_) | | | | | | | | | | | (_| | | | | (_| \__ \
 \____\___/|_| |_| |_|_| |_| |_|\__,_|_| |_|\__,_|___/
                                                      

build          | Build entire project
cli-install    | Install the Fission CLI
quality        | Run the complete test suite
repl           | Enter the project REPL
server-debug   | Run the Fission Server in debug verbose mode
server-install | Install the Fission Server
server-start   | Run the currently installed Fission Server
server-update  | Update & run the current server to the latest on the current branch
ssh-prod       | SSH into the production environment
ssh-staging    | SSH into the staging environment
watch          | Autobuild with file watcher
welcome        | Print the pretty welcome
```

The `build` command is a wrapper around `stack build` that builds all of the projects (CLI, server, etc). `server-install` will install the server binary where you can run it, and `server-debug` will run the server with some logging.

Before you run the server, there's a bit of docker, server and DNS configuration that is documented here: https://github.com/fission-codes/fission/tree/main/fission-web-server#using-docker.

I think that should be enough to get you up and running, but let me know if you'd like a sync chat to help out with any of it.

avivash:
> > I don't have a huge amount of context around this, but I just installed `nix` (i know i'm late to the `nix` party 😅), got `stack` setup and i successfully ran `stack build` on your branch, but nothing seems to be running on port `8001`. Is there a specific package I need to run? This is my first time using haskell 🙈
> 
> Hey @avivash! I can give you some pointers on project setup and running the server.
> 
> First, a question -- are you using the nix shell in the project? You can use it by running `nix-shell` at the root of the project. That will take a while to run because it will retrieve and compile a whole lot of Haskell.
> 
> Once that finishes and you are in the nix shell, you should be able to run `helpme` to get a help menu that looks like this:
> 
> ```shell
> $ helpme
>   ____                                          _     
>  / ___|___  _ __ ___  _ __ ___   __ _ _ __   __| |___ 
> | |   / _ \| '_ ` _ \| '_ ` _ \ / _` | '_ \ / _` / __|
> | |__| (_) | | | | | | | | | | | (_| | | | | (_| \__ \
>  \____\___/|_| |_| |_|_| |_| |_|\__,_|_| |_|\__,_|___/
>                                                       
> 
> build          | Build entire project
> cli-install    | Install the Fission CLI
> quality        | Run the complete test suite
> repl           | Enter the project REPL
> server-debug   | Run the Fission Server in debug verbose mode
> server-install | Install the Fission Server
> server-start   | Run the currently installed Fission Server
> server-update  | Update & run the current server to the latest on the current branch
> ssh-prod       | SSH into the production environment
> ssh-staging    | SSH into the staging environment
> watch          | Autobuild with file watcher
> welcome        | Print the pretty welcome
> ```
> 
> The `build` command is a wrapper around `stack build` that builds all of the projects (CLI, server, etc). `server-install` will install the server binary where you can run it, and `server-debug` will run the server with some logging.
> 
> Before you run the server, there's a bit of docker, server and DNS configuration that is documented here: https://github.com/fission-codes/fission/tree/main/fission-web-server#using-docker.
> 
> I think that should be enough to get you up and running, but let me know if you'd like a sync chat to help out with any of it.

Amazing! Thanks @bgins! I'll give that a try 🙏🏼

fission-web-server/library/Fission/Web/Server.hs
```diff
@@ -138,20 +138,32 @@ server ::
...
+
     v2 =
       genericServerT Fission.RoutesV2
         { api  = serverV2
         , docs = v2Docs
         }
 
+    serverV3 =
+      genericServerT Fission.V3
+        { user   = genericServerT User.handlerV3
```
bgins, line 164: What are your thoughts on versioning the other endpoints like `ipfs`, `app`, and `auth`? Doesn't seem right on this PR, but maybe we will want to version/alias them to have a consistent set of routes?

bgins (approved):
I'm missing some context on the Ethereum/secp256k1 bits of this PR, but I did a bit of testing and the code looks good to me.

I tested the following:

✅  The database migrates with no issues
✅ Registering a user using `walletauth` app. A user with an secp256k1 DID registers in the database.
✅ A user's public key can be updated on the update public key v3 route (tested with an Ed25519 PK).

One thing that I tried that failed was updating a user public key from Ed25519 to secp256k1. Not sure if that is a feature that we want.

A question about the old style public keys. Are those a compatibility layer while we migrate the database? Do we keep them around to support v2 routes?

There's more general testing I'd like to do, kicking the tires on other parts of the system overall, but I think that can happen when these changes are on staging.

bgins:
> Still can't figure out why data-root lookups are returning a `404` for me with my local instance. If anyone else is trying it out. You can test it with [walletauth](https://github.com/fission-codes/webnative-walletauth), by putting the following code in the `index.ts` file (replacing the existing `async` function):
> 
> ```ts
> ;(async () => {
>   wn.setup.endpoints({
>     api: "http://runfission.test:1337",
>     lobby: "http://localhost:8001",
>     user: "fissionuser.test"
>   })
> 
>   webnative.login()
> })()
> ```

I also ran into these issues, making calls from Insomnia. Seems to be down where we are calling on PowerDNS to resolve DNS records. Do we know if this is an issue in outside of this PR? If so, maybe we could test it on a pizza instance? Or move forward and test this part of the functionality on staging?

walkah:
> I also ran into these issues, making calls from Insomnia. Seems to be down where we are calling on PowerDNS to resolve DNS records. Do we know if this is an issue in outside of this PR? If so, maybe we could test it on a pizza instance? Or move forward and test this part of the functionality on staging?

If it's working locally - and tests are all passing - I'd vote to merge and deploy to staging for additional testing (and open subsequent issues if stuff comes up on staging).

fission-web-server/library/Fission/Web/Server.hs
icidasset: I don't know, maybe alias them to v1 or v0 sounds like a good idea.

bgins:
> > One thing that I tried that failed was updating a user public key from Ed25519 to secp256k1. Not sure if that is a feature that we want.
> 
> Hmm, that should be possible. How did it fail?

It fails with a 400 with this message in the response

```
Error in $: Unable to decode Ed25519 PK because: CryptoFailed CryptoError_PublicKeySizeInvalid / BDTfGgRGmbA1Vjob/f8GU2ZjvY3KYP6f2DMor5Z7Te33wWPD48ik4nQeD7Qez549ue78xK5EBohBFkTogZeJ7bo=
```

The original key that was to be replaced was an Ed25519 key.

icidasset:
Since updating public keys of the same type works I'm going to merge this in. I have to test transitioning between key types more extensively, but we can do that later (tracking in #629).

PR #628, opened 2022-08-09
Add migration in docker container instructions

### LINKED ISSUE #627: Update Postgres migration to include a docker one liner
Postgres migrations are here https://github.com/fission-codes/fission/tree/main/fission-web-server/sql

Let's get it updated with a one-liner to be able to run it in a docker container

bgins:
This PR adds some more details on migrating postgres when it's running in a docker container.

Closes #627.

### DIFF
--- a/fission-web-server/sql/README.md
+++ b/fission-web-server/sql/README.md
@@ -4,11 +4,22 @@
 This folder contains SQL files that should be run to get our database up to date. Each is prefixed with the unix time (in seconds) when it was created.
 
 ### Running the files
-You can run them using `\i` in you postgres terminal.
+
+You can run them using `\i` in your postgres terminal.
 
 ```sql
 \i 1576714308_selda_to_persistent.sql
 ```
+
+Note that `\i` assumes a relative path.
+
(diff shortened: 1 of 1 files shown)


fission-web-server/sql/README.md
```diff
@@ -4,11 +4,22 @@
...
 ```sql
 \i 1576714308_selda_to_persistent.sql
 ```
 
+Note that `\i` assumes a relative path.
+
+If you are running postgres in a Docker container, connect to container locally then run the migration file
+
+```sh
+psql -U postgres -d web_api -h runfission.test -p 5432
```
walkah, line 19: We should (could?) add `postgresql` to the nix shell environment for convenience - wanna do it in this PR?

walkah (approved):
:shipit: Looks good to me! (We can update the shell env separately)

fission-web-server/sql/README.md
bgins: Ah yeah, I see the comment above. We'll get this in and update the shell separately. 👌

PR #633, opened 2022-12-07
kubo rename and bump cli to 0.17

walkah:
Title says it all, bumping the bundled kubo version from 0.12 to 0.17 (which includes the s/go-ipfs/kubo/ rename)


walkah:
> Looks good! Do we want to update the IPFS image in the docker-compose file as well?

I've got a few docker / server tweaks coming I'll put this in with those.

PR #636, opened 2023-03-06
Endpoint user/challenger/recover

avivash:
## Summary

* [x] `user/challenge/recover` endpoint to return a `challenge` directly from the API rather than sending an email to the user that has the `challenge` appended to the dashboard link

We need an endpoint that returns a `challenge` directly so the `rs-email-service` can call `/v2/api/user/did/{Username}/{Challenge}` on behalf of user to recover their account.

## Test plan (required)

We will deploy this branch to staging to allow for easier testing before it's merged

Create a new user and call `/v2/api/user/challenge/recover/{Username}` with a valid UCAN as the bearer token, expect a `challenge` to be returned

## After Merge
* [x] Does this change require a release to be made? Is so please create and deploy the release

### DIFF
--- a/fission-web-api/library/Fission/Web/API/User/Challenge/Types.hs
+++ b/fission-web-api/library/Fission/Web/API/User/Challenge/Types.hs
@@ -1,0 +1,22 @@
+module Fission.Web.API.User.Challenge.Types (Routes (..)) where
+
+-- import           Fission.Challenge.Types
+
+import           Fission.Web.API.Prelude
+
+import           Fission.User.Username.Types
+import qualified Fission.Web.API.Auth.Types  as Auth
+
+data Routes mode = Routes
+  { recover ::
+      mode
+      :- "recover"
+      :> Summary "Return challenge for account recovery"
+      --
+      :> Capture "Username" Username
+      --
+      :> Auth.HigherOrder
+      :> GetNoContent
+      -- :> Post '[JSON] Challenge
+  }
+  deriving Generic
--- a/fission-web-api/library/Fission/Web/API/User/Challenge/Types.hs
+++ b/fission-web-api/library/Fission/Web/API/User/Challenge/Types.hs
@@ -15,7 +15,7 @@ data Routes mode = Routes
       --
       :> Capture "Username" Username
       --
-      -- :> Auth.HigherOrder
+      :> Auth.HigherOrder
       :> Post '[JSON] Challenge
   }
   deriving Generic
--- a/fission-web-api/library/Fission/Web/API/User/Challenge/Types.hs
+++ b/fission-web-api/library/Fission/Web/API/User/Challenge/Types.hs
@@ -4,6 +4,8 @@ import           Fission.Challenge.Types
 
 import           Fission.Web.API.Prelude
 
+import           Fission.User.Username.Types
+
 import qualified Fission.Web.API.Auth.Types  as Auth
 
 data Routes mode = Routes
@@ -11,6 +13,8 @@ data Routes mode = Routes
       mode
       :- "recover"
       :> Summary "Return challenge for account recovery"
+      --
+      :> Capture "Username" Username
       --
       :> Auth.HigherOrder
       :> Post '[JSON] Challenge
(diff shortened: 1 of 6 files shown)


fission-web-api/library/Fission/Web/API/User/Challenge/Types.hs
```diff
@@ -0,0 +1,21 @@
...
+
+data Routes mode = Routes
+  { recover ::
+      mode
+      :- "recover"
+      :> Summary "Return challenge for account recovery"
+      --
+      :> Capture "Username" Username
+      --
+      :> Auth.HigherOrder
```
matheus23, line 18: This line ensures that only fission-registered accounts (and UCANs that are delegated from fission-registered accounts) can make these requests :+1:

fission-web-server/library/Fission/Web/Server/Handler/User/Challenge.hs
```diff
@@ -0,0 +1,39 @@
...
+  , MonadLogger               m
+  , MonadTime                 m
+  )
+  => Challenge.Routes (AsServerT m)
+handler = Challenge.Routes {..}
+  where
+    recover username = do
+      Entity userId User { } <- Web.Err.ensureMaybe couldntFindUser =<< getByUsername username
+      now       <- currentTime
+      challenge <- RecoveryChallenge.create userId now
```
matheus23, line 34: However, here we don't check any other property about the authentication, right?

So if I read this correctly, doesn't this allow any fission user to issue & get a challenge for any other fission user?

E.g. a malicious actor could create a new fission account, and then use their UCAN to run a request `/user/challenge/recover/bob`, and then use the response to call `/user/did/bob/{challenge}` with a DID they control as the body for that request, essentially taking away my control over bob's fission account and giving it to themselves?

matheus23 (commented):
Hi :wave: I'm still subscribed to this repo, so I'm getting notifications and saw this coming up (& I worked on the email challenge bits before), so I thought I'd give it a short review, even though it wasn't requested :sweat_smile:

fission-web-server/library/Fission/Web/Server/Handler/User/Challenge.hs
avivash: Ahh thanks for catching that! I'll admit I have a fairly tenuous understanding of this codebase. This new endpoint was cobbled together from bits of various other ones I saw in the repo 😅 Looking at some other examples, I think I see how to add that validation here. I'll get this fixed today!

matheus23: What's the goal, actually? How should the auth gating work for that endpoint?
Is the plan to require the user to be authed to access this endpoint? Although in that case I wonder why they'd need to recover in the first place.
Is the idea to allow this endpoint only for the email service? I.e. the email service has a special DID that this fission server respects and it opens up this endpoint for it?
Or is the idea that the fission service's DID would be the "root owner" in this case, and we'd check that the root DID of a UCAN coming in to use the recovery endpoint is the fission service's DID, allowing us to delegate a UCAN for accessing the recovery endpoint to other services such as the email service?

avivash: The current approach we're taking is:
- On registration, the client builds a UCAN that delegates root access to the `rs-email-service`'s DID
- That UCAN is sent to the `rs-email-service` and stored in a DB that maps usernames to UCANS
- Then, when a user goes to recover their account, they enter their username and a request is sent from the client to the `rs-email-service`, which finds the associated UCAN in the DB and uses that to make a request to the this new fission server endpoint to get a challenge
- With the challenge in hand, the `rs-email-service` should be able to call `/v2/api/user/did/{Username}/{Challenge}` to recover the user's account

avivash: I should add that the `rs-email-service` is just meant to be an example of how this email recovery flow could work. We expect devs to create their own services that maybe reference our `rs-email-service` example

matheus23: That sound awesome and exactly like what I'd expect.

Let's go 🚀

